PR #11, opened 2019-08-08
People Adapter Interface

akoushke:


src/adapters/PeopleAdapter.test.js
```diff
@@ -0,0 +1,28 @@
+import {isObservable} from 'rxjs';
```
akoushke: I prefer to have `rxjs` globally available in our all testing module. but we can do that later.

src/adapters/PeopleAdapter.js
```diff
@@ -0,0 +1,13 @@
...
+
+/**
+ * This is the base class that `PeopleAdapter` will inherit from.
+ * When creating your subclass `peopleAdapter`, make sure all methods
+ * are defined and the items returned have the proper shape.
+ */
+export default class PeopleAdapter {
+  // eslint-disable-next-line no-unused-vars
+  getPerson(id) {
+    return throwError(new Error('getPerson(id) must be defined.'));
```
lalli-flores: `getPerson(id) must be defined in PeopleAdapter`

No period (.) at the end, and we should also add the context of where the error is coming from

src/adapters/PeopleAdapter.js
```diff
@@ -0,0 +1,13 @@
+import {throwError} from 'rxjs';
+
+/**
+ * This is the base class that `PeopleAdapter` will inherit from.
+ * When creating your subclass `peopleAdapter`, make sure all methods
+ * are defined and the items returned have the proper shape.
+ */
+export default class PeopleAdapter {
+  // eslint-disable-next-line no-unused-vars
+  getPerson(id) {
```
lalli-flores: Add a docstring for `getPerson`, specifying what it does, parameter descriptions and return values.
When we add the data object, we should add a link here for a sample of the expected return structure.

src/adapters/PeopleAdapter.js
```diff
@@ -0,0 +1,20 @@
+import {throwError} from 'rxjs';
+
+/**
+ * This is the base class that `PeopleAdapter` will inherit from.
```
lalli-flores: `This is a base class that defines the interface that maps people data.
Developers that want to extend `PeopleAdapater` must implement all of its methods, adhering to the exact parameters and structure of the returned objects
`

PR #13, opened 2019-08-14
Avatar

akoushke:


src/components/WebexAvatar/WebexAvatar.stories.js
```diff
@@ -1,6 +1,21 @@
...
-storiesOf('Webex Avatar', module).add('Avatar with title', () => <WebexAvatar title="Tom Smith" />);
+const stories = storiesOf('Webex Avatar', module);
+const peopleJSONAdapter = new PeopleJSONAdapter();
+
+stories.add('With appropriate PersonID', () => <WebexAvatar personID={person.id} adapter={peopleJSONAdapter} />);
+
+stories.add('With inappropriate PersonID', () => <WebexAvatar personID="Wrong personID" adapter={peopleJSONAdapter} />);
+
+stories.add('With different data', () => {
+  peopleJSONAdapter.getPerson(person.id).pipe(map((data) => ({...data, status: 'ooo'}))); // Mapped Observable
```
akoushke: we need to somehow pass this mapped observable to the component.

src/components/WebexAvatar/README.md
```diff
@@ -1 +1,7 @@
-Webex Avatar Component!
+# Webex Avatar Component
+Webex avatar component displays the avatar of a *Webex Teams* user with a given `personID`.
+
+## How to use
+```js
+  <WebexAvatar personID="PersonID" />
```
adamweeks: Should have the adapter defined here along with some details on methods the adapter needs and the properties the component expects to get from the adapter.

.storybook/config.js
```diff
@@ -1,6 +1,9 @@
-import { configure } from '@storybook/react';
+import {configure, addDecorator} from '@storybook/react';
+import {withKnobs} from '@storybook/addon-knobs';
 import '../momentum-ui.scss';
 
+addDecorator(withKnobs);
```
lalli-flores: Being nit-picky here, but can you add the decorator just before the configure call?
As we might add more add-ons, it feels more natural (to me) for them to be close the the configuration return

src/components/WebexAvatar/README.md
```diff
@@ -1 +1,7 @@
...
+```
```
lalli-flores: We should add pictures of what the component looks like

src/components/WebexAvatar/WebexAvatar.js
```diff
@@ -1,6 +1,41 @@
...
+  personID: PropTypes.string.isRequired,
+  adapter: PropTypes.object.isRequired,
+};
+
+export default class WebexAvatar extends PureComponent {
+  constructor(props) {
+    super(props);
+    this.props = props;
+    this.state = {
+      person: {},
```
lalli-flores: linter didn't complain about the `,`??

src/components/WebexAvatar/WebexAvatar.js
```diff
@@ -1,6 +1,41 @@
...
+  componentDidMount() {
+    this.props.adapter.getPerson(this.props.personID).subscribe(
+      (person) => {
+        this.setState({person});
+      },
+      (error) => {
+        // eslint-disable-next-line no-console
+        console.error(error.message);
+      },
+      () => {
```
lalli-flores: If we are not handling anything at completion, is not necessary to add this.

src/components/WebexAvatar/WebexAvatar.js
```diff
@@ -1,6 +1,41 @@
...
+        // Observable completed
+      }
+    );
+  }
+
+  render() {
+    const {avatar, displayName, status} = this.state.person;
+
+    return <Avatar src={avatar} title={displayName} type={status} />;
```
lalli-flores: Let's add `alt` as well, with the person's `displayname`

src/components/WebexAvatar/WebexAvatar.stories.js
lalli-flores: With @adamweeks's idea of passing in the JSON object to the JSON adapter class, I think that solves the issue of how to "change data".
We would have a story for each of the UI states, and we would just modify the JSON object `status` before passing it to `PeopleJSONAdapter`.

src/components/WebexAvatar/WebexAvatar.js
akoushke: no

src/components/WebexAvatar/WebexAvatar.js
```diff
@@ -1,6 +1,41 @@
...
+  constructor(props) {
+    super(props);
+    this.props = props;
+    this.state = {
+      person: {},
+    };
+  }
+
+  componentDidMount() {
+    this.props.adapter.getPerson(this.props.personID).subscribe(
```
adamweeks: Can we structure our subscribe params into a shape that is a little more readable? I don't remember the order of `next`, `error`, `complete`. The [docs from RxJS](https://rxjs-dev.firebaseapp.com/guide/operators) has a good example that I find easier to process in my head:

```js
  observable.subscribe({
      next(value) {
        const timerID = setTimeout(() => {
          observer.next(value);
          allTimerIDs.delete(timerID);
        }, delayInMillis);
        allTimerIDs.add(timerID);
      },
      error(err) {
        observer.error(err);
      },
      complete() {
        observer.complete();
      }
    });
```

src/adapters/PeopleJSONAdapter.js
```diff
@@ -3,6 +3,11 @@ import {Observable} from 'rxjs';
 import PeopleAdapter from './PeopleAdapter';
 
 export default class PeopleJSONAdapter extends PeopleAdapter {
+  constructor(datasource) {
+    super(datasource);
+    this.getPerson = this.getPerson.bind(this);
```
adamweeks: Good catch.

src/components/WebexAvatar/WebexAvatar.js
```diff
@@ -1,6 +1,41 @@
...
+      (person) => {
+        this.setState({person});
+      },
+      () => {
+        this.setState({
+          person: {
+            displayName: 'X X',
```
adamweeks: Can we come up with an agreed state that represents the avatar being in an error state? I didn't see one mentioned in the ticket @lalli-flores

akoushke: I removed it

src/components/WebexAvatar/WebexAvatar.js
akoushke: so the cool thing about `anonymous` arrow function is that `this` is not defined and mapped to the parent `this` which is what we need for `this.setState` function to work properly.

src/adapters/PeopleAdapter.js
```diff
@@ -2,6 +2,21 @@ import {throwError} from 'rxjs';
 
 import WebexAdapter from './WebexAdapter';
 
+// Person Status Enum
```
adamweeks: I really like this enum addition. Could you document it with an [enum](https://jsdoc.app/tags-enum.html) jsdoc?

adamweeks: Also, let's get rid of the `PersonStatus` jsdoc typedef and just do something like:

```js
* @property {null|string}   status      The presence status of the user. @see PersonStatus enum
```

src/components/WebexAvatar/WebexAvatar.js
```diff
@@ -1,6 +1,38 @@
...
+      person: {},
+    };
+  }
+
+  componentDidMount() {
+    this.props.adapter.getPerson(this.props.personID).subscribe(
+      (person) => {
+        this.setState({person});
+      },
+      () => {
```
adamweeks: We should pass the error from the adapter and print it to the console as well.

src/components/WebexAvatar/WebexAvatar.js
adamweeks: Could we do something like this then?:

```js
const onNext = (person) => {dostuff();};
const onError = (error) => {dostuff();};

this.props.adapter.getPerson(this.props.personID).subscribe(onNext, onError);
```

Just trying to come up with a way to make this as readable as possible for external collaborators.

src/components/WebexAvatar/WebexAvatar.stories.js
```diff
@@ -1,6 +1,28 @@
...
+import {personStatuses} from '../../adapters/PeopleAdapter';
+import data, {ID} from '../../data/person';
+
 import WebexAvatar from './WebexAvatar';
 
-storiesOf('Webex Avatar', module).add('Avatar with title', () => <WebexAvatar title="Tom Smith" />);
+// Setup for the stories
+const person = {};
+
+person[ID] = data;
```
adamweeks: Should we build a people.json data file instead of person.json? I was considering doing so with the PeopleJSONAdapter implementation so it would line up one-to-one data:adapter

akoushke: yeah, I'm going to do that. its cleaner

src/adapters/PeopleAdapter.js
```diff
@@ -10,7 +29,7 @@ import WebexAdapter from './WebexAdapter';
 export default class PeopleAdapter extends WebexAdapter {
   /**
    * The status a person can have.
-   * @typedef {null|'active'|'bot'|'call'|'dnd'|'group'|'inactive'|'meeting'|'ooo'|'presenting'|'self'|'typing'} PersonStatus
+   * @property {null|string}   status      The presence status of the user. @see personStatus enum
```
adamweeks: This should be replaced on line 46 and lines 30-33 removed completely.

adamweeks:
I think the "fix" commits should qualify for "feat" instead

src/components/WebexAvatar/WebexAvatar.js
```diff
@@ -1,6 +1,34 @@
...
+      person: {},
+    };
+  }
+
+  componentDidMount() {
+    const onNext = (person) => this.setState({person});
+    // eslint-disable-next-line no-console
+    const onError = (error) => console.error(error.message);
+
+    this.props.adapter.getPerson(this.props.personID).subscribe(onNext, onError);
```
adamweeks: 😎

src/components/WebexAvatar/WebexAvatar.js
```diff
@@ -1,6 +1,34 @@
...
+  }
+
+  render() {
+    const {avatar, displayName, status} = this.state.person;
+
+    return <Avatar src={avatar} title={displayName} type={status} alt={displayName} />;
```
adamweeks: Does momentum default the alt to title? If so, the `alt={displayName}` is redundant.

src/components/WebexAvatar/WebexAvatar.test.js
```diff
@@ -1,19 +1,41 @@
...
+    global.console.error = jest.fn();
+  });
+
+  test('matches snapshot with "default" status', () => {
+    expect(shallow(<WebexAvatar personID={personID} adapter={peopleJSONAdapter} />)).toMatchSnapshot();
   });
 
-  test('exists', () => {
-    expect(wrapper).toMatchSnapshot();
+  Object.keys(personStatus).forEach((status) => {
```
adamweeks: I'm really not a fan of these test generators in the SDK. I'd prefer there being a test case for each.

src/components/WebexAvatar/WebexAvatar.js
akoushke: no it does not: https://github.com/momentum-design/momentum-ui/blob/master/react/src/lib/Avatar/index.js#L238

src/components/WebexAvatar/WebexAvatar.test.js
lalli-flores: Agreed, while is easier to code, it's much harder to read later. It is also counter to "tests as documentation" way of thinking.

src/components/WebexAvatar/WebexAvatar.js
```diff
@@ -1,6 +1,34 @@
...
+  componentDidMount() {
+    const onNext = (person) => this.setState({person});
+    // eslint-disable-next-line no-console
+    const onError = (error) => console.error(error.message);
+
+    this.props.adapter.getPerson(this.props.personID).subscribe(onNext, onError);
+  }
+
+  render() {
+    const {avatar, displayName, status} = this.state.person;
```
lalli-flores: What happens when `person = {}`?

akoushke:
I don't use commit suggestion. since it will mess my commits. you guys don't have to use it.

src/components/WebexAvatar/WebexAvatar.js
akoushke: nothing gets rendered.

src/components/WebexAvatar/WebexAvatar.test.js
```diff
@@ -1,19 +1,92 @@
...
-  test('exists', () => {
-    expect(wrapper).toMatchSnapshot();
+  test(`matches snapshot with "${PersonStatus.MEETING}" person status`, () => {
+    peopleJSONAdapter.datasource[personID].status = PersonStatus.MEETING;
+
+    expect(shallow(<WebexAvatar personID={personID} adapter={peopleJSONAdapter} />)).toMatchSnapshot();
+  });
+
+  test(`matches snapshot with "${PersonStatus.OUT_OF_OFFICE}" person status`, () => {
+    peopleJSONAdapter.datasource[personID].status = PersonStatus.MEETING;
```
lalli-flores: This doesn't match the description

src/components/WebexAvatar/WebexAvatar.stories.js
```diff
@@ -1,6 +1,78 @@
...
-storiesOf('Webex Avatar', module).add('Avatar with title', () => <WebexAvatar title="Tom Smith" />);
+// Setup for the stories
+const [personID] = Object.keys(person);
+const peopleJSONAdapter = new PeopleJSONAdapter(person);
+const stories = storiesOf('Webex Avatar', module);
+
+// Stories
+stories.add('With no person status', () => <WebexAvatar personID={personID} adapter={peopleJSONAdapter} />);
+
+stories.add(`With '${PersonStatus.ACTIVE}' person status`, () => {
```
lalli-flores: I think the story names would be easier to read if it was just the status name.
```suggestion
stories.add(`${PersonStatus.ACTIVE}`, () => {
```

PR #14, opened 2019-08-15
feat(activity): add ActivityAdapter interface

lalli-flores:


src/adapters/ActivityAdapter.js
```diff
@@ -0,0 +1,20 @@
...
+ * This is a base class that defines the interface that maps activity data.
+ * Developers that want to extend `ActivityAdapter` must implement all of its methods,
+ * adhering to the exact parameters and structure of the returned objects.
+ */
+export default class ActivityAdapter {
+  /**
+   * Returns an observable that emits message data of the given ID.
+   *
+   * @param {String} id - ID of the message to get
+   * @returns {Observable<Activity>}
```
adamweeks: Are we going to define the shape of the `<Activity>` object in the jsdoc as well?

lalli-flores: Yes, that's a good idea!

src/adapters/ActivityAdapter.js
```diff
@@ -0,0 +1,31 @@
+import {throwError} from 'rxjs';
+
+/**
+ * This is a base class that defines the interface that maps activity data.
+ * Developers that want to extend `ActivityAdapter` must implement all of its methods,
+ * adhering to the exact parameters and structure of the returned objects.
+ */
+export default class ActivityAdapter {
```
adamweeks: Don't forget to extend from the base WebexAdapter class!

PR #15, opened 2019-08-15
docs(PeopleAdapter): add PersonObject typedef

adamweeks:


package.json
```diff
@@ -106,7 +106,17 @@
...
       "@commitlint/config-conventional"
-    ]
+    ],
+    "rules": {
+      "scope-case": [
+        2,
+        "always",
+        [
+          "lower-case",
+          "pascal-case"
```
adamweeks: Added pascal case so we can do things like `feat(PeopleAdapter):`

src/adapters/PeopleAdapter.js
```diff
@@ -6,11 +6,26 @@ import {throwError} from 'rxjs';
...
+   * @param {string} status The presence status of the user.
+   *                        One of: '', 'active', 'bot', 'call', 'dnd', 'group', 'inactive', 'meeting', 'ooo', 'presenting', 'self', 'typing'
+   */
+
   /**
    * Returns an observable that emits person data
    *
    * @param {String} id - ID of person to get
-   * @returns {Observable<PersonObject>}
+   * @returns {Observable.<PersonObject>}
```
adamweeks: Added the `.` to match the type documentation from jsdoc: https://jsdoc.app/tags-type.html

Definitely open to discussion on this.

lalli-flores: Can you also rename the class?
```suggestion
   * @returns {Observable.<Person>}
```

src/adapters/PeopleAdapter.js
```diff
@@ -6,11 +6,26 @@ import {throwError} from 'rxjs';
  * adhering to the exact parameters and structure of the returned objects
  */
 export default class PeopleAdapter {
+  /**
+   * A Person object with details about the person
+   * @typedef {object} PersonObject
```
lalli-flores: We should probably define this as `Person`, adding `Object` seems redundant.

src/adapters/PeopleAdapter.js
```diff
@@ -6,11 +6,26 @@ import {throwError} from 'rxjs';
...
+   * @param {string} id The person identifier
+   * @param {Array.<String>} emails An array of emails for the person
```
lalli-flores: This would still be `string`, no?

```suggestion
   * @param {Array.<string>} emails An array of emails for the person
```

src/adapters/PeopleAdapter.js
lalli-flores: I'm ok with adding the `.`

src/adapters/PeopleAdapter.js
```diff
@@ -6,11 +6,26 @@ import {throwError} from 'rxjs';
...
+   * @param {string} displayName The name to be displayed for the person
+   * @param {string} firstName The first name of the person
+   * @param {string} lastName The last name of the person
+   * @param {string} nickName The short name for the person
+   * @param {string} avatar The full url to the person's avatar
+   * @param {string} orgId The id of the organization the person belongs to
+   * @param {string} status The presence status of the user.
+   *                        One of: '', 'active', 'bot', 'call', 'dnd', 'group', 'inactive', 'meeting', 'ooo', 'presenting', 'self', 'typing'
```
lalli-flores: We should probably document the only the ones we will support, and update if/when we support more states. From Jira specs there are:
* `null` -> no state, (open for discussion on this one because the API returns `'unknown'` for no state)
* `self`
* `active`
* `bot`
* `dnd`
* `ooo`

package.json
lalli-flores: Should we be specifying the scope as a filename? Since we didn't have this before, I was thinking it was actually a little bit more useful to specify the "area" where the change was made, for instance, for this PR it would be `feat(people)`.
For my PR I did `feat(activity)`

src/adapters/PeopleAdapter.js
```diff
@@ -6,11 +6,26 @@ import {throwError} from 'rxjs';
  * adhering to the exact parameters and structure of the returned objects
  */
 export default class PeopleAdapter {
+  /**
+   * A Person object with details about the person
```
lalli-flores: Would it be too specific to say:
`A Person object with details about a Webex Teams user`.

src/adapters/PeopleAdapter.js
```diff
@@ -6,11 +6,26 @@ import {throwError} from 'rxjs';
...
+   * @param {string} lastName The last name of the person
+   * @param {string} nickName The short name for the person
+   * @param {string} avatar The full url to the person's avatar
+   * @param {string} orgId The id of the organization the person belongs to
+   * @param {string} status The presence status of the user.
+   *                        One of: '', 'active', 'bot', 'call', 'dnd', 'group', 'inactive', 'meeting', 'ooo', 'presenting', 'self', 'typing'
+   */
+
   /**
    * Returns an observable that emits person data
```
lalli-flores: Can you add `of the given ID.` at the end? I'm doing the same for `ActivityAdapter`

src/adapters/PeopleAdapter.js
lalli-flores: Shouldn't this also be `Object`?
```suggestion
   * @typedef {Object} PersonObject
```

src/adapters/PeopleAdapter.js
```diff
@@ -6,11 +6,26 @@ import {throwError} from 'rxjs';
  * adhering to the exact parameters and structure of the returned objects
  */
 export default class PeopleAdapter {
+  /**
+   * A Person object with details about the person
+   * @typedef {object} PersonObject
+   * @param {string} id The person identifier
```
lalli-flores: Shouldn't this be `@property` instead? https://jsdoc.app/tags-property.html

package.json
adamweeks: I like that the scope translates to a specific thing that users will import. It makes the change log align with what users are familiar with, like:

### Features

* **WebexAvatar:** add support for animated profiles like Harry Potter

adamweeks:
Removed commitlint changes and pulled into a separate [PR](https://github.com/webex/components/pull/17)

src/adapters/PeopleAdapter.js
adamweeks: Will do. I just grabbed these from momentum

adamweeks: Actually, was there a reason for not supporting all the types that momentum supports?

src/adapters/PeopleAdapter.js
adamweeks: I think it might be too specific. Users of the library could populate the data from an external source. I know it isn't very likely, but might as well keep it generic enough for those use cases.

src/adapters/PeopleAdapter.js
adamweeks: Ugh, I keep getting `Object` and `string` capitalization backwards! Good find

src/adapters/PeopleAdapter.js
adamweeks: Yup. Good 👀

adamweeks:
@lalli-flores addressed your comments and added some questions for discussion.

src/adapters/PeopleAdapter.js
lalli-flores: I guess I was thinking about it in terms of what we will implement and how we get the data. There are some status that we won't be supporting with the MVP of the SDK adapter, like `typing`, but it really doesn't affect the avatar's ability to do it since Momentum supports it. You can ignore my comment

src/adapters/PeopleAdapter.js
```diff
@@ -7,10 +7,25 @@ import {throwError} from 'rxjs';
...
+   * @property {string} avatar The full url to the person's avatar
+   * @property {string} orgId The id of the organization the person belongs to
+   * @property {string} status The presence status of the user.
+   *                        One of: null, 'active', 'bot', 'call', 'dnd', 'group', 'inactive', 'meeting', 'ooo', 'presenting', 'self', 'typing'
+   */
+
+  /**
+   * Returns an observable that emits person data of the given id
    *
    * @param {String} id - ID of person to get
```
lalli-flores:
```suggestion
   * @param {string} id - ID of person to get
```

src/adapters/PeopleAdapter.js
lalli-flores: Sure

src/adapters/PeopleAdapter.js
```diff
@@ -7,10 +7,25 @@ import {throwError} from 'rxjs';
...
+   * @property {string} id The person identifier
+   * @property {Array.<string>} emails An array of emails for the person
+   * @property {string} displayName The name to be displayed for the person
+   * @property {string} firstName The first name of the person
+   * @property {string} lastName The last name of the person
+   * @property {string} nickName The short name for the person
+   * @property {string} avatar The full url to the person's avatar
+   * @property {string} orgId The id of the organization the person belongs to
+   * @property {string} status The presence status of the user.
+   *                        One of: null, 'active', 'bot', 'call', 'dnd', 'group', 'inactive', 'meeting', 'ooo', 'presenting', 'self', 'typing'
```
lalli-flores: Do we have a column limit? I really dislike horizontal scrolling. At least on the Github interface this line causes the horizontal scroll

src/adapters/PeopleAdapter.js
```diff
@@ -7,10 +7,25 @@ import {throwError} from 'rxjs';
...
+   * @typedef {Object} Person
+   * @property {string} id The person identifier
+   * @property {Array.<string>} emails An array of emails for the person
+   * @property {string} displayName The name to be displayed for the person
+   * @property {string} firstName The first name of the person
+   * @property {string} lastName The last name of the person
+   * @property {string} nickName The short name for the person
+   * @property {string} avatar The full url to the person's avatar
+   * @property {string} orgId The id of the organization the person belongs to
+   * @property {string} status The presence status of the user.
```
lalli-flores: Since you added `null` for not status, should this be `{null|string}`?

src/adapters/PeopleAdapter.js
adamweeks: img

src/adapters/PeopleAdapter.js
adamweeks: Resolved via teams.

src/adapters/PeopleAdapter.js
```diff
@@ -7,14 +7,33 @@ import {throwError} from 'rxjs';
  */
 export default class PeopleAdapter {
   /**
-   * Returns an observable that emits person data
+   * The status a person can have.
+   * @typedef {null|'active'|'bot'|'call'|'dnd'|'group'|'inactive'|'meeting'|'ooo'|'presenting'|'self'|'typing'} PersonStatus                      a              a      a    a
```
lalli-flores: There are a bunch of random characters at the end of this line.

```suggestion
   * @typedef {null|'active'|'bot'|'call'|'dnd'|'group'|'inactive'|'meeting'|'ooo'|'presenting'|'self'|'typing'} PersonStatus
```

src/adapters/PeopleAdapter.js
```diff
@@ -7,14 +7,33 @@ import {throwError} from 'rxjs';
...
+   * @property {string} firstName The first name of the person.
+   * @property {string} lastName The last name of the person.
+   * @property {string} nickName The short name for the person.
+   * @property {string} avatar The full url to the person's avatar.
+   * @property {string} orgID The ID of the organization the person belongs to.
+   * @property {PersonStatus} status The presence status of the user.
+   */
+
+  /**
+   * Returns an observable that emits person data of the given ID
```
lalli-flores:
```suggestion
   * Returns an observable that emits person data of the given ID.
```

Sorry 😬

src/adapters/PeopleAdapter.js
```diff
@@ -7,14 +7,33 @@ import {throwError} from 'rxjs';
...
   /**
-   * Returns an observable that emits person data
+   * The status a person can have.
+   * @typedef {null|'active'|'bot'|'call'|'dnd'|'group'|'inactive'|'meeting'|'ooo'|'presenting'|'self'|'typing'} PersonStatus                      a              a      a    a
+   */
+
+  /**
+   * A Person object with details about the person.
+   * @typedef {Object} Person
+   * @property {string} ID The person identifier.
```
lalli-flores: Not sure how this would affect doc generation, but isn't it a bit more readable to have the descriptions aligned? That's what I did for [activity on my PR](https://github.com/webex/components/blob/3d72bc1965f192ac60b2e6bc3ca1ba9fc2ebf372/src/adapters/ActivityAdapter.js#L13), and how they have it in JSDocs example, https://jsdoc.app/tags-property.html

src/adapters/PeopleAdapter.js
adamweeks: Wow. I didn't think I committed that, was testing the width limits 🤦‍♂

src/adapters/PeopleAdapter.js
adamweeks: Sure!

PR #17, opened 2019-08-16
build(commitlint): allow pascal case scopes for classes

adamweeks:
Moving this out of the other PR so discussion can focus on one topic at a time 😃

Previous discussion between myself and @lalli-flores
image

### DIFF
--- a/package.json
+++ b/package.json
@@ -106,7 +106,17 @@
   "commitlint": {
     "extends": [
       "@commitlint/config-conventional"
-    ]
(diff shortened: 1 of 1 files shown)


lalli-flores:
I think there are 2 things for discussion, 1) should we have Pascal case, 2) How we identify scopes.
For the first one, I don't have any problems with allow it, so technically we can merge this PR, for the other, that's something to agree on

PR #19, opened 2019-08-22
Activityjsonadapter

akoushke:


src/adapters/ActivitiesAdapter.js
```diff
@@ -4,14 +4,14 @@ import WebexAdapter from './WebexAdapter';
...
+ * Developers that want to extend `ActivitiesAdapter` must implement all of its methods,
  * adhering to the exact parameters and structure of the returned objects.
  */
-export default class ActivityAdapter extends WebexAdapter {
+export default class ActivitiesAdapter extends WebexAdapter {
   /**
    * An activity a person performs in Webex.
    *
-   * @typedef {Object} Activity
+   * @typedef  {Object} Activities
```
adamweeks: This is just a singular activity so while the adapter's name `ActivitiesAdapter` works, this should just be an `Activity`.

src/adapters/ActivitiesAdapter.js
```diff
@@ -4,14 +4,14 @@ import WebexAdapter from './WebexAdapter';
...
    * @property {string} ID        The activity identifier.
    * @property {string} roomID    ID of the room where the activity happens.
    * @property {string} text      Any text the activity may contain.
```
adamweeks: Can you add `displayAuthor` which is in the JSON as well?

src/adapters/PeopleJSONAdapter.test.js
```diff
@@ -1,40 +1,39 @@
-import person from './../data/people';
+import people from './../data/people';
```
adamweeks: Thanks for fixing these!

src/adapters/ActivitiesAdapter.js
```diff
@@ -0,0 +1,34 @@
...
+   * @property {string}  personID       ID of the person performing the activity.
+   * @property {Date}    created        Timestamp of the time when the activity happened.
+   * @property {Boolean} displayAuthor  Whether to display author information or not.
+   */
+
+  /**
+   * Returns an observable that emits activity data of the given ID.
+   *
+   * @param {string} ID  ID of the activity to get.
+   * @returns {Observable.<Activities>}
```
adamweeks:
```suggestion
   * @returns {Observable.<Activity>}
```

src/adapters/ActivitiesJSONAdapter.test.js
```diff
@@ -0,0 +1,50 @@
...
+  beforeEach(() => {
+    [activityID] = Object.keys(activities);
+    activitiesJSONAdapter = new ActivitiesJSONAdapter(activities);
+  });
+
+  test('getActivity() returns an observable', () => {
+    expect(rxjs.isObservable(activitiesJSONAdapter.getActivity())).toBeTruthy();
+  });
+
+  test('getActivity() returns a activity data', (done) => {
```
adamweeks:
```suggestion
  test('getActivity() returns an activity', (done) => {
```

src/adapters/ActivitiesJSONAdapter.test.js
```diff
@@ -0,0 +1,50 @@
...
+    activitiesJSONAdapter.getActivity(activityID).subscribe((data) => {
+      expect(data).toEqual(activities[activityID]);
+      done();
+    });
+  });
+
+  test('getActivity() throws a proper error message', (done) => {
```
adamweeks:
```suggestion
  test('getActivity() throws a proper error message when activity doesn\'t exist', (done) => {
```

src/adapters/ActivitiesJSONAdapter.test.js
```diff
@@ -0,0 +1,50 @@
...
+
+describe('Activities Adapter Interface', () => {
+  let activitiesJSONAdapter, activityID;
+
+  beforeEach(() => {
+    [activityID] = Object.keys(activities);
+    activitiesJSONAdapter = new ActivitiesJSONAdapter(activities);
+  });
+
+  test('getActivity() returns an observable', () => {
```
adamweeks: Since `getActivity` is the only method in the adapter, we may not need this right now, but it would be good to group all of the `getActivity()` tests in their own describe block similar to the way the SDK does:

```js
describe('getActivity() functionality', () => {
  test('getActivity() returns an observable', () => {
```

akoushke: yes, will do that after we come up with additional method.

src/adapters/ActivitiesJSONAdapter.js
```diff
@@ -0,0 +1,45 @@
+import {Observable} from 'rxjs';
+
+import ActivitiesAdapter from './ActivitiesAdapter';
+
+export default class ActivitiesJSONAdapter extends ActivitiesAdapter {
```
lalli-flores: We should add a docstring to this class. Something like:
```
/*
 * Implements the ActivitiesAdapter interface with a JSON object as its datasource. See @ActivitiesJSON
 */
```

We should probably do the same on `PeopleJSONAdapter`.

src/adapters/ActivitiesJSONAdapter.js
```diff
@@ -0,0 +1,45 @@
...
+  constructor(datasource) {
+    super(datasource);
+    this.getActivity = this.getActivity.bind(this);
+  }
+
+  /**
```
lalli-flores: Can we pull this out of the class and before the class docstring?

src/adapters/ActivitiesJSONAdapter.js
```diff
@@ -0,0 +1,45 @@
...
+   * @typedef ActivitiesJSON
+   * @param {object} datasource An object that contains a set of activity keyed by ID.
```
lalli-flores: `.. a set of activities`

src/adapters/ActivitiesJSONAdapter.test.js
```diff
@@ -0,0 +1,50 @@
+import activities from './../data/activities';
+import ActivitiesJSONAdapter from './ActivitiesJSONAdapter';
+
+describe('Activities Adapter Interface', () => {
```
lalli-flores: `Activities JSON Adapter`

src/adapters/ActivitiesJSONAdapter.test.js
```diff
@@ -0,0 +1,50 @@
...
+  });
+
+  test('getActivity() returns an activity', (done) => {
+    activitiesJSONAdapter.getActivity(activityID).subscribe((data) => {
+      expect(data).toEqual(activities[activityID]);
+      done();
+    });
+  });
+
+  test("getActivity() throws a proper error message when activity doesn't exist", (done) => {
```
lalli-flores: Can you use single-quote like the other strings?

PR #20, opened 2019-08-22
Static proptypes

lalli-flores:


akoushke:
what is the point of making this `static`?

package.json
```diff
@@ -37,6 +37,7 @@
   },
   "devDependencies": {
     "@babel/core": "^7.5.5",
+    "@babel/plugin-proposal-class-properties": "^7.5.5",
```
adamweeks: This is required to do static props?

src/components/WebexAvatar/WebexAvatar.js
```diff
@@ -30,5 +30,3 @@ export default class WebexAvatar extends PureComponent {
     return <Avatar src={avatar} title={displayName} type={status} alt={displayName} />;
   }
 }
-
-WebexAvatar.propTypes = propTypes;
```
adamweeks: I like that it removes this strange looking code.

package.json
lalli-flores: Yes, `static` is considered experimental since it's still a proposal

lalli-flores:
> what is the point of making this `static`?

It's a cleaner way of attaching a property to an object

PR #25, opened 2019-08-30
CircleCI Cache Fix

akoushke:
So I installed our components library and noticed that our `dist` directory is not downloaded into the directory. in our previous PR we had to separate our `build` job from `release` job in case of blocking the release if the build job had failed. after that change, I noticed that the `dist` directory is not included to the `release` job, which makes sense. We did not add the `dist` directory anywhere to be included to the next job which is a release job. So I decided to use `workspace` to persist the data (node_modules from the `install` job + dist directory from the `build` job) between the jobs. This actually turned out to be a better performant process, since we are not unarchiving the whole repo for each job anymore. and only apply the `node_modules` - 63 MB layer to all jobs and `dist` - 1.3 MB layer to `release` job. this will definitely pay off in the future as our repo keeps getting larger. Let me know if you have anymore questions.

## How to test

comment out the workflows rules and you should be able to view all the jobs behavioral details. `npm run release` does not publish on a non-master branch. so you could leave the code as is

## Example
I thought [this](https://circleci.com/workflow-run/6f0a749a-3b7b-4940-aa39-f50d2b0e4a17) was a good example of our workflow. don't mind the workflow order. our flow right now is as follows:

install -> test -> build -> release

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -8,31 +8,36 @@ executors:
 
 # reusable command for all jobs
 commands:
-  checkout_from_cache:
-    description: 'To checkout and restore the dependencies cache'
+  checkout_from_workspace:
+    description: 'To checkout and attach the workspace'
     steps:
       - checkout
-      - restore_cache:
-          keys:
-            - dependency-cache-{{ checksum "package.json" }}
-            - dependency-cache-
+      - attach_workspace:
+          at: ~/webex-components
 
 # define jobs
 jobs:
   install:
     executor: main-executor
     steps:
-      - checkout_from_cache
+      - checkout
+      - restore_cache:
+          keys:
+            - dependency-cache-{{ checksum "package-lock.json" }}
       - run:
           name: Install Dependencies
-          command: npm ci
+          command: npm i
       - save_cache:
-          key: dependency-cache-{{ checksum "package.json" }}
+          key: dependency-cache-{{ checksum "package-lock.json" }}
           paths: node_modules
+      - persist_to_workspace:
+          root: .
+          paths: node_modules
+
   test:
     executor: main-executor
     steps:
-      - checkout_from_cache
+      - checkout_from_workspace
       - run:
           name: Linting
           command: npm run test:eslint
@@ -41,24 +46,28 @@ jobs:
           command: npm run test:coverage
       - store_test_results:
           path: test_results
+
   build:
     executor: main-executor
     steps:
-      - checkout_from_cache
+      - checkout_from_workspace
       - run:
-          name: build
+          name: Build
           command: npm run build
+      - persist_to_workspace:
+          root: .
+          paths: dist
+
   release:
     executor: main-executor
     steps:
-      - checkout_from_cache
+      - checkout_from_workspace
       - run:
-          name: release
+          name: Release
           command: npm run release
 
 # execute the jobs in a orderly manner
 workflows:
-  version: 2
   setup_test_release:
     jobs:
       - install


.circleci/config.yml
```diff
@@ -8,31 +8,37 @@ executors:
...
+      - restore_cache:
+          keys:
+            - dependency-cache-{{ checksum "package-lock.json" }}
+            - dependency-cache-
       - run:
           name: Install Dependencies
           command: npm ci
       - save_cache:
-          key: dependency-cache-{{ checksum "package.json" }}
+          key: dependency-cache-{{ checksum "package-lock.json" }}
```
akoushke: decided to pick `package-lock` instead of `package`, since we are using `npm ci`

.circleci/config.yml
```diff
@@ -8,31 +8,37 @@ executors:
...
+          paths: node_modules
+      - persist_to_workspace:
+          root: .
           paths: node_modules
```
akoushke: add the node_modules to the workspace

.circleci/config.yml
```diff
@@ -41,17 +47,22 @@ jobs:
...
     executor: main-executor
     steps:
-      - checkout_from_cache
+      - checkout_from_workspace
       - run:
           name: build
           command: npm run build
+      - persist_to_workspace:
+          root: .
+          paths: dist
```
akoushke: add the dist directory to the workspace

adamweeks:
The changes look good to me. I've added @taymoork2 as a reviewer since he has some more recent experience with circle too.

.circleci/config.yml
```diff
@@ -67,6 +78,7 @@ workflows:
             - install
       - build:
           requires:
+            - install
```
lalli-flores: What does requiring `install` for `build` change?

.circleci/config.yml
```diff
@@ -8,31 +8,37 @@ executors:
...
+          at: ~/webex-components
 
 # define jobs
 jobs:
   install:
     executor: main-executor
     steps:
-      - checkout_from_cache
+      - checkout
+      - restore_cache:
```
taymoork2: Restoring before `npm ci` is meaningless, since `npm ci` wipes `node_modules` before installing.
I removed the restore before installing dependencies in my PR and just have it `save_cache` after

.circleci/config.yml
taymoork2: I changed root to `root: ~/webex-js-sdk` (or in your case `root: ~/webex-components`) since I didn't know what `./` was pointing too

.circleci/config.yml
taymoork2: Nothing, I guess it's a _"just-in-case"_. But `install` should be blocked by the `test` job anyway which requires `install` there

taymoork2:
Remove the `version: 2` from the workflows section as that sometimes causes circleci to get confused which version to use when it tries to parse the config.
Also, it's unnecessary to specify since you're specifying 2.1 above

.circleci/config.yml
akoushke: It looks at workspace definition. . Is fine

.circleci/config.yml
akoushke: then what's the point of having a cache. I think I figured what to do now.

taymoork2: `save_cache` caches between triggers. So everytime circleci runs it will use the same node_modules cache. This "could" be used for the deploy job since there isn't an expectation of new modules being installed at that time and the checksum on `package-lock` should be the same

akoushke: sure, your example save the cache but you are not using/restoring the cache if you remove `restore_cache`. And that does NOT serve the purpose of caching. and thats why saving the cache and not restoring it is as good as not having the cache in the first place. Anyways, as I mentioned earlier I figured what to do. But I do suggest you read more about CircleCI configuration here: https://circleci.com/docs/2.0/configuration-reference/#save_cache
Screen Shot 2019-08-31 at 11 22 23 AM

.circleci/config.yml
```diff
@@ -41,24 +46,28 @@ jobs:
...
       - store_test_results:
           path: test_results
+
   build:
     executor: main-executor
     steps:
-      - checkout_from_cache
+      - checkout_from_workspace
       - run:
           name: build
```
lalli-flores: Capitalize for consistency, e.g. `Build`

.circleci/config.yml
```diff
@@ -41,24 +46,28 @@ jobs:
...
+          root: .
+          paths: dist
+
   release:
     executor: main-executor
     steps:
-      - checkout_from_cache
+      - checkout_from_workspace
       - run:
           name: release
```
lalli-flores: Capitalize for consistency, e.g. `Release`

PR #27, opened 2019-09-09
feat(rooms): implemented rooms JSON adapter interface

akoushke:


src/adapters/RoomsJSONAdapter.js
```diff
@@ -0,0 +1,83 @@
...
+ * }
+ */
+
+/*
+ * Implements the RoomsAdapter interface with a JSON object as its datasource. See @RoomsJSON
+ */
+export default class RoomsJSONAdapter extends RoomsAdapter {
+  constructor(datasource) {
+    super(datasource);
+    this.getRoom = this.getRoom.bind(this);
```
lalli-flores: Is there a reason for not binding `getPreviousRoomActivities` and `getRoomActivities`?

src/adapters/RoomsJSONAdapter.js
```diff
@@ -0,0 +1,83 @@
...
+   * @param {string} ID  ID of the room to get.
+   * @returns {Observable.<Array.<string>>}
+   * @memberof RoomsAdapter
+   */
+  getPreviousRoomActivities(ID) {
+    return Observable.create((observer) => {
+      if (this.datasource[ID]) {
+        observer.next(this.datasource[ID]);
+      } else {
+        observer.error(new Error(`Could not find previous activities associated to the room with ID "${ID}"`));
```
lalli-flores: In case of error, we should probably send an empty array. We shouldn't crash the app because there was no data. Instead, the component will render nothing by passing it an empty array..
In addition, we can `console.error` a message for developers if we want.

src/adapters/RoomsJSONAdapter.js
```diff
@@ -0,0 +1,83 @@
...
+   * @param {string} ID  ID of the room to get.
+   * @returns {Observable.<Array.<string>>}
+   * @memberof RoomsAdapter
+   */
+  getRoomActivities(ID) {
+    return Observable.create((observer) => {
+      if (this.datasource[ID]) {
+        observer.next(this.datasource[ID]);
+      } else {
+        observer.error(new Error(`Could not find current activities associated to the room with ID "${ID}"`));
```
lalli-flores: As with `getPreviousActivities`, we should send an empty array instead of an error, and potentially log the error to console `Could not retrieve new activity for room "${ID}"`.

src/adapters/RoomsJSONAdapter.js
akoushke: I kinda addressed your comments. take a look

src/adapters/RoomsJSONAdapter.test.js
```diff
@@ -0,0 +1,111 @@
...
+  beforeEach(() => {
+    [roomID, roomActivitiesID] = Object.keys(rooms);
+    roomsJSONAdapter = new RoomsJSONAdapter(rooms);
+  });
+
+  test('getRoom() returns an observable', () => {
+    expect(rxjs.isObservable(roomsJSONAdapter.getRoom())).toBeTruthy();
+  });
+
+  test('getRoom() returns a person data', (done) => {
```
lalli-flores:
```suggestion
  test('getRoom() returns room data', (done) => {
```

adamweeks: Looks like this request was overlooked

akoushke: I fixed it on my next branch already

src/adapters/RoomsJSONAdapter.js
```diff
@@ -0,0 +1,82 @@
...
+  /**
+   * Returns an observable that emits an array of current activity IDs of the given roomID.
+   *
+   * @param {string} ID  ID of the room to get.
+   * @returns {Observable.<Array.<string>>}
+   * @memberof RoomsAdapter
+   */
+  getRoomActivities(ID) {
+    return Observable.create((observer) => {
+      const data = !this.datasource[ID] ? [] : this.datasource[ID];
```
adamweeks: If a user passes in the room ID as defined in the jsdoc, this will not return the activities array, this returns the room data. 

It should do: 
```js
this.datasource[`${ID}-activities`]
```

This way we can still just use the room ID and don't have to pass the knowledge of adding a `-activities`, especially since the RoomsAdapter defines the prop as the room ID and not room ID + activities string.

akoushke: yeah I was going to do that at the hook level, but adapter should be fine too.

PR #30, opened 2019-09-12
docs(readme): add adapter explanation and usage

lalli-flores:


README.md
```diff
@@ -34,36 +35,68 @@ npm install --save @webex/components
...
+### Adapters
+Webex Components are self-updating, meaning, they know how to fetch the data they need.
+Data is passed to components via adapter classes (see [adapters](./src/adapters)).
+Adapters are an uniform interface for the Webex Components to consume.
+They also know how to map the data from their data source to the data the components need.
 
+To use a Webex Component, start by creating a Webex Adapter:
 ```js
-import {WebexAvatar} from '@webex/components';
+import '@webex/components/dist/webexComponents.css';
```
akoushke: I dont think this is necessary. you have it imported in your components file.

lalli-flores: What components file?
The idea is that the sections build on each other: style, style + adapter, style + adapter + component

akoushke: this looks like you are making this file an entry file, which does not have to be. it looks a bit confusing to me. You are introducing the adapter but adding extra code about style import declaration which is not relevant to your description.

akoushke: we are addressing the styling import declaration in the earlier step with two ways. and we should leave it at that.

akoushke: this means that you are basically assuming that all these declaration will occur in `index.js` which potentially will not be the case.

PR #33, opened 2019-09-17
ActivityStream hook UseState --> useReducer

akoushke:


src/components/hooks/useActivityStream.js
```diff
@@ -1,5 +1,21 @@
-import {useState, useEffect} from 'react';
-import {merge} from 'rxjs';
+import {useEffect, useReducer} from 'react';
+
+const APPEND_ACTIVITIES = 'append_activities';
+
+function reducer(activityIDs, action) {
```
lalli-flores: Can you add a docstring here?

```
...
Returns a new array of activityIDs based on the given action. If no action is passed, it will return the same activityIDs array without any changes.
...
```
or something like that

akoushke:
> For your commit, I think is more of a refactor than a new feature, and make sure to spell `useReducer` react hook to clarify on the commit message.

camel case is not allowed. should I enable it?

src/components/hooks/useActivityStream.js
```diff
@@ -1,5 +1,30 @@
...
+
+const APPEND_ACTIVITIES = 'append_activities';
+
+/**
+ * Returns a new array of activityIDs based on the given action.
+ * If no action is passed, it will return the same activityIDs
+ * array without any changes.
+ *
+ * @param {Array.<string>} activityIDs activityIDs associated to the room
+ * @param {obj} action action to apply to given activityIDs
```
lalli-flores: This should not be `obj` (also in line 33). We should do a quick [typedef|https://jsdoc.app/tags-typedef.html] for the Action object expected.

Line 33 should be a RoomAdapter object type.

akoushke: `typedef` to me is not necessary right now. I did change line 33 to RoomAdapter though

src/components/hooks/useActivityStream.js
```diff
@@ -1,22 +1,45 @@
...
+
+  return newActivityIDs;
+}
 
 /**
  * Custom hook that returns activity data associated to the room of the given ID.
  *
  * @param {string} roomID  ID of the room for which to return data.
- * @param {obj} roomsAdapter  Component data adapter from which to retrieve data.
+ * @param {RoomAdapter} roomsAdapter  Component data adapter from which to retrieve data.
```
adamweeks:
```suggestion
 * @param {RoomsAdapter} roomsAdapter  Component data adapter from which to retrieve data.
```

src/components/hooks/useActivityStream.js
```diff
@@ -1,22 +1,45 @@
...
+ * @param {Array.<string>} activityIDs activityIDs associated to the room
+ * @param {object} action action to apply to given activityIDs
+ * @returns {Array.<string>} New activityIDs associated to the room
+ */
+function reducer(activityIDs, action) {
+  let newActivityIDs = [];
+
+  switch (action.type) {
+    case APPEND_ACTIVITIES:
+      newActivityIDs = [...activityIDs].concat(action.payload);
```
adamweeks: Is the array deconstruction necessary here since `activityIDs` is already an array?

src/components/hooks/useActivityStream.js
```diff
@@ -1,22 +1,45 @@
...
-  const [activityIDs, setActivityIDs] = useState([]);
+  const [activityIDs, dispatch] = useReducer(reducer, []);
 
   useEffect(() => {
-    const activityStream = merge(
-      roomsAdapter.getPreviousRoomActivities(roomID),
-      roomsAdapter.getRoomActivities(roomID)
-    );
-    const subscription = activityStream.subscribe(setActivityIDs);
+    const subscription = roomsAdapter.getRoomActivities(roomID).subscribe((activities) => {
```
adamweeks: Is the `getPreviousRoomActivities` case going to be handled in another ticket? Is that why it was removed?

akoushke: yes

src/components/hooks/useActivityStream.js
akoushke: no it's not.

PR #34, opened 2019-09-17
Time ruler support

adamweeks:
image

### DIFF
--- a/src/adapters/RoomsAdapter.js
+++ b/src/adapters/RoomsAdapter.js
@@ -29,6 +29,12 @@ export default class RoomsAdapter extends WebexAdapter {
    */
 
   /**
+   * @typedef {Object} ActivityData
+   * @param {string} date The date to render a time ruler for. Must be a valid date-time string
+   *     https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/parse#Date_Time_String_Format
+   */
+
+  /**
    * Returns an observable that emits room data of the given ID.
    *
    * @param {string} ID  ID of the room to get.
@@ -53,10 +59,10 @@ export default class RoomsAdapter extends WebexAdapter {
   }
 
   /**
-   * Returns an observable that emits an array of current activity IDs of the given roomID.
+   * Returns an observable that emits an array of current activities of the given roomID.
    *
    * @param {string} ID  ID of the room to get.
-   * @returns {Observable.<Array.<string>>}
+   * @returns {Observable.<Array.<string|ActivityData>>}
    * @memberof RoomsAdapter
    */
   // eslint-disable-next-line no-unused-vars
--- a/src/adapters/RoomsJSONAdapter.test.js
+++ b/src/adapters/RoomsJSONAdapter.test.js
@@ -94,6 +94,15 @@ describe('Rooms JSON Adapter Interface', () => {
     });
   });
 
+  test('getRoomActivities() returns objects with a date parameter for time rulers', (done) => {
+    const timeRulerRoomID = 'time-rulers';
+
+    roomsJSONAdapter.getRoomActivities(timeRulerRoomID).subscribe((data) => {
+      expect(data[0].date).toBeDefined();
+      done();
+    });
+  });
+
   test('getRoomActivities() completes the observable', (done) => {
     roomsJSONAdapter.getRoomActivities(roomID).subscribe(
       () => {},
--- a/src/components/WebexActivityStream/WebexActivityStream.js
+++ b/src/components/WebexActivityStream/WebexActivityStream.js
@@ -1,11 +1,43 @@
 import React, {Fragment} from 'react';
 import PropTypes from 'prop-types';
+import {ListSeparator} from '@momentum-ui/react';
+import {format, isToday, isSameWeek, isYesterday} from 'date-fns';
 
 import {RoomType} from '../../adapters/RoomsAdapter';
 import {useActivityStream, useRoom} from '../hooks';
 import WebexActivity from '../WebexActivity/WebexActivity';
 
 import './WebexActivityStream.scss';
+
+/**
+ * Returns a formatted timestamp based on the given date's offset from the current time.
+ *
+ * Divisor for messages from today display today
+ * Divisor for messages from dates from a previous day display as Yesterday
+ * Divisor for messages of dates from a previous day of the week (but not yesterday) display as <day of the week>
+ * Divisor for messages of dates older than a week from today display as M/D/YY
+ *
+ * @param {Date} timestamp Date instance to format
+ * @returns {string} formattedDate
+ */
+export function formatTimeRulerText(timestamp) {
+  let formattedDate;
+
+  if (isToday(timestamp)) {
+    formattedDate = 'today';
+  } else if (isYesterday(timestamp)) {
+    // Yesterday
+    formattedDate = 'Yesterday';
+  } else if (isSameWeek(timestamp, new Date())) {
+    // Monday
+    formattedDate = format(timestamp, 'iiii');
+  } else {
+    // 1/1/2020
+    formattedDate = format(timestamp, 'P');
+  }
+
+  return formattedDate;
+}
 
 export function GreetingDirectSVG() {
   return (
@@ -138,19 +170,34 @@ Greeting.propTypes = {
   personName: PropTypes.string.isRequired,
 };
 
+export function TimeRuler({text}) {
+  return <ListSeparator className="time-ruler" role="listitem" text={text} />;
+}
+
+TimeRuler.propTypes = {
+  text: PropTypes.string.isRequired,
+};
+
 export default function WebexActivityStream(props) {
   const {roomID, adapters} = props;
   const {roomsAdapter, activitiesAdapter, peopleAdapter} = adapters;
   const {title, roomType} = useRoom(roomID, roomsAdapter);
-  const activityIDs = useActivityStream(roomID, roomsAdapter);
+  const activitiesData = useActivityStream(roomID, roomsAdapter);
   const personName = roomType === RoomType.DIRECT ? title : '';
-  const activities = activityIDs.map((activityID) => (
-    <WebexActivity key={activityID} activityID={activityID} adapters={{activitiesAdapter, peopleAdapter}} />
-  ));
+  const activities = activitiesData.map((activity) => {
+    // If the activity is an object with a date property, it is a time ruler
+    const activityComponent = activity.date ? (
+      <TimeRuler key={activity.date.toString()} text={formatTimeRulerText(new Date(activity.date))} />
+    ) : (
+      <WebexActivity key={activity} activityID={activity} adapters={{activitiesAdapter, peopleAdapter}} />
+    );
+
+    return activityComponent;
+  });
 
   return (
     <div className="activity-stream">
-      {activityIDs.length ? <Fragment>{activities}</Fragment> : <Greeting personName={personName} />}
+      {activities.length ? <Fragment>{activities}</Fragment> : <Greeting personName={personName} />}
     </div>
   );
 }
--- a/src/components/WebexActivityStream/WebexActivityStream.scss
+++ b/src/components/WebexActivityStream/WebexActivityStream.scss
@@ -16,3 +16,7 @@
     padding: 0 3.125rem;
   }
 }
+
+.time-ruler {
+  padding: 0 3rem;
+}
--- a/src/components/WebexActivityStream/__snapshots__/WebexActivityStream.test.js.snap
+++ b/src/components/WebexActivityStream/__snapshots__/WebexActivityStream.test.js.snap
@@ -1108,3 +1108,930 @@ exports[`Webex Activity Stream component Webex Activity Stream snapshots matches
   />
 </div>
 `;
+
+exports[`Webex Activity Stream component Webex Activity Stream snapshots matches with time rulers stream 1`] = `
+<div
+  className="activity-stream"
+>
+  <TimeRuler
+    key="2019-08-20T21:00:07.047Z"
+    text="2019-08-20T21:00:07.047Z P"
+  />
+  <WebexActivity
+    activityID="default"
+    adapters={
+      Object {
+        "activitiesAdapter": ActivitiesJSONAdapter {
+          "datasource": Object {
+            "default": Object {
+              "ID": "default",
+              "created": "2019-08-20T21:00:07.047Z",
+              "displayHeader": true,
+              "personID": "default",
+              "roomID": "roomID",
+              "text": "Lorem ipsum dolor sit amet, consectetur adipiscing elit.",
+            },
+            "long": Object {
+              "ID": "long",
+              "created": "2019-08-20T21:00:07.047Z",
+              "displayHeader": true,
+              "personID": "default",
+              "roomID": "roomID",
+              "text": "Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor incididunt ut labore et dolore magna aliqua. Ut enim ad minim veniam, quis nostrud exercitation ullamco laboris nisi ut aliquip ex ea commodo consequat. Duis aute irure dolor in reprehenderit in voluptate velit esse cillum dolore eu fugiat nulla pariatur. Excepteur sint occaecat cupidatat non proident, sunt in culpa qui officia deserunt mollit anim id est laborum.",
+            },
+            "multi-line": Object {
+              "ID": "multi-line",
+              "created": "2019-08-20T21:00:07.047Z",
+              "displayHeader": true,
+              "personID": "default",
+              "roomID": "roomID",
+              "text": "Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor incididunt ut labore et dolore magna aliqua.
+Ut enim ad minim veniam, quis nostrud exercitation ullamco laboris nisi ut aliquip ex ea commodo consequat.
+Duis aute irure dolor in reprehenderit in voluptate velit esse cillum dolore eu fugiat nulla pariatur.
+Excepteur sint occaecat cupidatat non proident, sunt in culpa qui officia deserunt mollit anim id est laborum.",
+            },
+            "no-header": Object {
+              "ID": "no-header",
+              "created": "2019-08-20T21:00:07.047Z",
+              "displayHeader": false,
+              "personID": "default",
+              "roomID": "roomID",
+              "text": "no header text",
+            },
+            "old": Object {
+              "ID": "old",
+              "created": "old",
+              "displayHeader": true,
+              "personID": "default",
+              "roomID": "roomID",
+              "text": "Lorem ipsum dolor sit amet, consectetur adipiscing elit.",
+            },
+            "sameWeek": Object {
+              "ID": "sameWeek",
+              "created": "sameWeek",
+              "displayHeader": true,
+              "personID": "default",
+              "roomID": "roomID",
+              "text": "Lorem ipsum dolor sit amet, consectetur adipiscing elit.",
+            },
+            "today": Object {
+              "ID": "today",
+              "created": "today",
+              "displayHeader": true,
+              "personID": "default",
+              "roomID": "roomID",
+              "text": "Lorem ipsum dolor sit amet, consectetur adipiscing elit.",
+            },
+            "yesterday": Object {
+              "ID": "yesterday",
+              "created": "yesterday",
+              "displayHeader": true,
+              "personID": "default",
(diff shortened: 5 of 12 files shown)


src/adapters/RoomsJSONAdapter.test.js
```diff
@@ -94,6 +94,17 @@ describe('Rooms JSON Adapter Interface', () => {
     });
   });
 
+  test('getRoomActivities() converts date strings into date objects for time rulers', (done) => {
+    const timeRulerRoomID = 'time-rulers';
+
+    roomsJSONAdapter.getRoomActivities(timeRulerRoomID).subscribe((data) => {
```
akoushke: I would prefer three test cases

src/components/WebexActivityStream/WebexActivityStream.js
```diff
@@ -1,12 +1,44 @@
...
+ *
+ * Divisor for messages from today display today
+ * Divisor for messages from dates from a previous day display as Yesterday
+ * Divisor for messages of dates from a previous day of the week (but not yesterday) display as <day of the week>
+ * Divisor for messages of dates older than a week from today display as M/D/YY
+ *
+ * @param {Date} timestamp Date instance to format
+ * @returns {string} formattedDate
+ */
+export function formatTimeRulerText(timestamp) {
```
akoushke: we do have another function that does the same in `Activity` component. maybe it's time to extract it out and use it in here and there.

src/components/WebexActivityStream/WebexActivityStream.js
```diff
@@ -138,19 +170,33 @@ Greeting.propTypes = {
...
   const personName = roomType === RoomType.DIRECT ? title : '';
-  const activities = activityIDs.map((activityID) => (
-    <WebexActivity key={activityID} activityID={activityID} adapters={{activitiesAdapter, peopleAdapter}} />
-  ));
+  const activities = activitiesData.map((activity) => {
+    if (activity.date) {
+      return <TimeRuler key={activity.date.toString()} text={formatTimeRulerText(activity.date)} />;
+    }
+
+    const activityID = activity;
```
akoushke: can't quite follow why we are declaring `activityID` again.

src/components/WebexActivityStream/WebexActivityStream.js
```diff
@@ -138,19 +170,33 @@ Greeting.propTypes = {
...
   const {roomID, adapters} = props;
   const {roomsAdapter, activitiesAdapter, peopleAdapter} = adapters;
   const {title, roomType} = useRoom(roomID, roomsAdapter);
-  const activityIDs = useActivityStream(roomID, roomsAdapter);
+  const activitiesData = useActivityStream(roomID, roomsAdapter);
   const personName = roomType === RoomType.DIRECT ? title : '';
-  const activities = activityIDs.map((activityID) => (
-    <WebexActivity key={activityID} activityID={activityID} adapters={{activitiesAdapter, peopleAdapter}} />
-  ));
+  const activities = activitiesData.map((activity) => {
```
akoushke: I would prefer a ternary operator

src/components/WebexActivityStream/WebexActivityStream.scss
```diff
@@ -16,3 +16,8 @@
     padding: 0 3.125rem;
   }
 }
+
+.time-ruler {
+  padding-left: 4.5rem;
```
akoushke: maybe 
```scss
padding: 0 1.5rem 0 4.5rem
```
to be consistent?

src/components/WebexActivityStream/__snapshots__/WebexActivityStream.test.js.snap
```diff
@@ -1108,3 +1108,13 @@ exports[`Webex Activity Stream component Webex Activity Stream snapshots matches
   />
 </div>
 `;
+
+exports[`Webex Activity Stream component Webex Activity Stream snapshots matches with time rulers stream 1`] = `
```
akoushke: snapshot is not correct.

src/adapters/RoomsJSONAdapter.js
```diff
@@ -74,9 +76,32 @@ export default class RoomsJSONAdapter extends RoomsAdapter {
...
 }
+
+/**
+ * Activities could be objects with metadata (like date separators).
+ * Convert the data into proper javascript objects
+ *
+ * @param {Array} activityData json formatted activity data
+ * @returns {Array}
+ */
+RoomsJSONAdapter.formatActivity = (activityData) =>
```
akoushke: can we implement the function the way we did it in all other adapters?

src/adapters/RoomsJSONAdapter.js
```diff
@@ -74,9 +76,32 @@ export default class RoomsJSONAdapter extends RoomsAdapter {
...
+  activityData.map((activity) => {
+    let formattedActivity = activity;
+
+    if (typeof activity === 'object') {
```
akoushke: this should be `activityID` I would assume? It was a bit hard for me to realize that this is not actual activity object.

src/adapters/RoomsJSONAdapter.js
```diff
@@ -74,9 +76,32 @@ export default class RoomsJSONAdapter extends RoomsAdapter {
...
+      // Clone the object since we will be modifying it
+      formattedActivity = Object.assign({}, activity);
```
akoushke: spread operator is what we have been utilizing for the rest of the implementation.

akoushke: that way a simple ternary operator would do it here, I think?

src/adapters/RoomsJSONAdapter.js
```diff
@@ -74,9 +76,32 @@ export default class RoomsJSONAdapter extends RoomsAdapter {
     return Observable.create((observer) => {
       const data = !this.datasource[`${ID}-activities`] ? [] : this.datasource[`${ID}-activities`];
 
-      observer.next(data);
+      const formattedData = RoomsJSONAdapter.formatActivity(data);
```
akoushke: any reason we have the extra new line here?

src/components/WebexActivityStream/WebexActivityStream.js
```diff
@@ -138,19 +170,33 @@ Greeting.propTypes = {
   personName: PropTypes.string.isRequired,
 };
 
+export function TimeRuler({text}) {
```
akoushke: any reason we have the extra `{}` in the params here?

adamweeks: Destructures the props object so we don't have to do `props.text`.

src/adapters/RoomsJSONAdapter.js
adamweeks: I'm not sure I understand what you are asking

src/adapters/RoomsJSONAdapter.js
adamweeks: This is to clone the `activity` object so we are not modified the object directly within the `.map` loop.

src/adapters/RoomsJSONAdapter.js
adamweeks: updated to be more legible.

src/adapters/RoomsJSONAdapter.test.js
adamweeks: The test verifies the date is being correctly converted. We don't need a test case for every `expect` statement.

src/components/WebexActivityStream/WebexActivityStream.js
adamweeks: They are similar but the business logic is different enough to warrant separate functions.

src/components/WebexActivityStream/WebexActivityStream.js
adamweeks: It was for readability since `activity` could be an object or a string. I've removed it.

src/components/WebexActivityStream/__snapshots__/WebexActivityStream.test.js.snap
adamweeks: img

adamweeks:
@akoushke ready for another round, but still having snapshot issues as discussed in the teams space.

akoushke:
@adamweeks, 

change `src/components/hooks/__mock__/useActivityStream.js` line 5: 
```js
  return roomIDs.includes(ID) ? rooms[`${ID}-activities`] : [];
```

src/components/WebexActivityStream/WebexActivityStream.js
akoushke: I wish we could have done that for all other components. now there is inconsistency on how we extract the props.

src/adapters/RoomsJSONAdapter.js
akoushke: yeah thats what spread operator would do too. 
```js
formattedActivity = {...activity};
```

src/adapters/RoomsJSONAdapter.js
akoushke: Lalli explained it better [here](https://github.com/webex/components/pull/34#discussion_r325929812)

src/adapters/RoomsJSONAdapter.test.js
akoushke: IMO the last expect statement covers the other two statements: 
```js
expect(data[0].date instanceof Date).toBeTruthy();
```
in order for `data[0].data` to be a `Date` instance, it must:
 1) be an object 
 2) be defined

Therefore, to me the first two expect statements are unnecessary to check.

src/components/hooks/__mocks__/useActivityStream.js
```diff
@@ -1,6 +1,6 @@
 export default function useActivityStream(ID, adapter) {
-  const [roomID] = Object.keys(adapter.datasource);
+  const roomIDs = Object.keys(adapter.datasource);
   const rooms = adapter.datasource;
 
-  return ID === roomID ? rooms[`${ID}-activities`] : [];
+  return roomIDs.includes(ID) ? rooms[`${ID}-activities`] : [];
```
akoushke: sorry this should be: 
```js
roomIDs.includes(`${ID}-activities`) ? rooms[`${ID}-activities`] : [];
```

src/components/WebexActivityStream/WebexActivityStream.js
```diff
@@ -138,19 +170,34 @@ Greeting.propTypes = {
...
+TimeRuler.propTypes = {
+  text: PropTypes.string.isRequired,
+};
+
 export default function WebexActivityStream(props) {
   const {roomID, adapters} = props;
   const {roomsAdapter, activitiesAdapter, peopleAdapter} = adapters;
   const {title, roomType} = useRoom(roomID, roomsAdapter);
-  const activityIDs = useActivityStream(roomID, roomsAdapter);
+  const activitiesData = useActivityStream(roomID, roomsAdapter);
```
akoushke: could this be `activityStream`? too picky I know. Definitely not a blocker though.

src/components/WebexActivityStream/WebexActivityStream.js
akoushke: gotcha

src/components/WebexActivityStream/WebexActivityStream.js
adamweeks: I can change to `props` if needed for consistency.

src/adapters/RoomsJSONAdapter.js
```diff
@@ -73,10 +74,34 @@ export default class RoomsJSONAdapter extends RoomsAdapter {
...
+ * Convert the data into proper javascript objects
+ *
+ * @param {Array} activitiesData json formatted activity data
+ * @returns {Array}
+ */
+RoomsJSONAdapter.formatActivitiesData = (activitiesData) =>
+  activitiesData.map((activity) => {
+    if (typeof activity === 'object') {
+      // Clone the object since we will be modifying it
+      const formattedActivity = Object.assign({}, activity);
```
lalli-flores: 2 part comment.
1) Wouldn't it be easier to just return the date object instead of an object that contains a date object? That way we avoid creating a lot of new objects.

2) If cloning the object is necessary, can you update this to use the spread operator? That's how it's being done in other places when objects are being copied.
```suggestion
      const formattedActivity = {...activity};
```

Alternatively, I actually think we can avoid this method altogether if we just return `{date: "string date"}` and have the `WebexActivityStream` make the object conversion just-in-time, since by the time we have to call `TimeRuler`, we know it has to be a date.
This would make our code a little more efficient by not looping through the whole array every time there's a render and potentially creating 2 new objects for non-activities. See my comment on `WebexActivityStream`

src/adapters/RoomsJSONAdapter.js
adamweeks: If this is the required way to do it, we need to add it to our eslint rules:
https://github.com/eslint/eslint/pull/9955

src/components/WebexActivityStream/WebexActivityStream.js
```diff
@@ -138,19 +170,34 @@ Greeting.propTypes = {
...
   const personName = roomType === RoomType.DIRECT ? title : '';
-  const activities = activityIDs.map((activityID) => (
-    <WebexActivity key={activityID} activityID={activityID} adapters={{activitiesAdapter, peopleAdapter}} />
-  ));
+  const activities = activitiesData.map((activity) => {
+    // If the activity is an object with a date property, it is a time ruler
+    const activityComponent = activity.date ? (
+      <TimeRuler key={activity.date.toString()} text={formatTimeRulerText(activity.date)} />
```
lalli-flores: You can avoid pre-formatting by just creating the Date object here.
```suggestion
      <TimeRuler key={activity.date.toString()} text={formatTimeRulerText(new Date(activity.date))} />
```

src/components/WebexActivityStream/WebexActivityStream.js
adamweeks: > Right, the time rulers format does not include time.

Ironically. 🤣

src/adapters/RoomsJSONAdapter.test.js
adamweeks: It helped when developing to know what part failed, but I will remove it.

src/components/hooks/__mocks__/useActivityStream.js
lalli-flores: There's actually an [`in`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/in) operator.
```suggestion
  return `${ID}-activities` in roomIDs ? rooms[`${ID}-activities`] : [];
```

src/components/WebexActivityStream/WebexActivityStream.js
akoushke: no, I actually like it better. Lalli suggested to fix it for other components too. it will save us some lines total.

src/adapters/RoomsJSONAdapter.js
adamweeks: Sorry, didn't mean for that to come off short. If we have preferred methods of doing something, we should convey that via eslint instead of having to rely on code reviews 👍

src/components/hooks/__mocks__/useActivityStream.js
akoushke: then you need to change it everywhere else in our hooks. but is it necessary?

src/components/WebexActivityStream/WebexActivityStream.js
lalli-flores: I'm already using it, and will fix it on my upcoming PR

src/adapters/RoomsJSONAdapter.js
adamweeks: 2 part answer: 😄 

1. I'm following the requirements in the JIRA (https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-92114) that has the data shape form the adapter being `{date: 09/12/2019}`. Please update the requirements if you'd like it that way.

2. @akoushke commented the same thing. If there are consistent ways we want to do things, we need to add them as eslint rules (https://github.com/eslint/eslint/pull/9955 for this case). We can't expect contributors to be familiar with the entire codebase and to know how we expect certain patterns.

src/components/WebexActivityStream/WebexActivityStream.js
adamweeks: I don't understand what you mean by this. `activity.date` is already a date object.

src/components/hooks/__mocks__/useActivityStream.js
adamweeks: If we use `in` it should be: 
```js
`${ID}-activities` in adapter.datasource ?
```

Otherwise we can use the `roomIDs` keys array. What would you prefer?

lalli-flores: yes, `adapter.datasource` should be fine. I was just looking at the one line without context 😛

src/adapters/RoomsJSONAdapter.js
lalli-flores: @adamweeks, you're right, the requirements state that the adapter should return a `Date` object. The idea being, having some way of identifying when something is not an activity.

I realized that having that requirements makes the code unnecessarily and artificially inefficient. I will update the requirement for the adapter to have an object with a value string that is compliant with the [date-time string format](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/parse#Date_Time_String_Format).
E.g. something like `{date: "2019-12-03"}`

Alternatively, we could flatten the array and only have strings and just validate whenever a string is a valid date with [datefns's `isValid`](https://date-fns.org/v1.9.0/docs/isValid)
e.g. `["2019-09-12". "activity1", "activity2" ...]`

Open for feedback

src/components/WebexActivityStream/WebexActivityStream.js
lalli-flores: This is based on my previous comment on the activity formatting, assuming that the adapter won't pass a `Date` object anymore

src/components/hooks/__mocks__/useActivityStream.js
```diff
@@ -1,6 +1,3 @@
 export default function useActivityStream(ID, adapter) {
-  const [roomID] = Object.keys(adapter.datasource);
-  const rooms = adapter.datasource;
-
-  return ID === roomID ? rooms[`${ID}-activities`] : [];
+  return `${ID}-activities` in adapter.datasource ? adapter.datasource[`${ID}-activities`] : [];
```
akoushke: can you make changes for all other mocks then? if you are planning to use `in`? `usePerson`, `useRoom` are all using `includes` right now.

src/adapters/RoomsJSONAdapter.js
adamweeks: I'm good with the data being a date string format as your example shows with: `{date: "2019-12-03"}`.

I prefer the data being passed to the components have a well defined shape. I think performing the datefn's validation on every item in the array on every render would be too much processing.

I'll update the PR to push the string to date conversion later in the process.

src/adapters/RoomsAdapter.js
```diff
@@ -28,6 +28,11 @@ export default class RoomsAdapter extends WebexAdapter {
    * @property {RoomType}  type    The type of the room. @see RoomType enum
    */
 
+  /**
+   * @typedef {Object} ActivityData
+   * @param {string} date The date to render a time ruler for
```
lalli-flores: I would like to specify the format for the string
```suggestion
   * @param {string} date The date to render a time ruler for. Must be a valid date-time string
   *     https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/parse#Date_Time_String_Format
```

src/adapters/RoomsAdapter.js
```diff
@@ -56,7 +61,7 @@ export default class RoomsAdapter extends WebexAdapter {
    * Returns an observable that emits an array of current activity IDs of the given roomID.
```
lalli-flores: Since we're returning more than IDs now:
```suggestion
   * Returns an observable that emits an array of current activities of the given roomID.
```

src/components/WebexActivityStream/WebexActivityStream.js
```diff
@@ -138,19 +170,34 @@ Greeting.propTypes = {
   personName: PropTypes.string.isRequired,
 };
 
+export function TimeRuler({text}) {
+  return <ListSeparator className="time-ruler" role="listitem" text={text} />;
+}
+
+TimeRuler.propTypes = {
+  text: PropTypes.string.isRequired,
```
lalli-flores: Remove trailing comma, please
```suggestion
  text: PropTypes.string.isRequired
```

PR #37, opened 2019-09-23
add destructuring rule

akoushke:


.eslintrc
```diff
@@ -76,6 +66,16 @@
     ],
     "react-hooks/rules-of-hooks": "error", // Checks rules of Hooks
     "react-hooks/exhaustive-deps": "warn", // Checks effect dependencies
-    "require-jsdoc": "off"
+    "require-jsdoc": "off",
+    "prefer-destructuring": [
```
adamweeks: Could you move this to line 77 so the rules stay consistently in alphabetical order?

PR #39, opened 2019-09-23
Avatar OnError

akoushke:


src/components/WebexAvatar/__snapshots__/WebexAvatar.test.js.snap
```diff
@@ -241,3 +241,5 @@ exports[`Webex Avatar component matches snapshot with "typing" person status 1`]
   type="typing"
 />
 `;
+
+exports[`Webex Avatar component matches snapshot with wrong personID 1`] = `[Function]`;
```
adamweeks: Snapshot looks wrong here.

src/components/hooks/usePerson.js
```diff
@@ -14,7 +14,12 @@ export default function usePerson(personID) {
 
   useEffect(() => {
     const onError = (error) => {
-      throw error;
+      setPerson({
+        displayName: ' ',
```
lalli-flores: What happens if you set displayName to `null` or `undefined`?

akoushke: nothing gets rendered

akoushke:
> Could you squash the commits?

yes i did

src/components/hooks/usePerson.js
adamweeks: Could we pass alt text to say something like "Error: this is the error message"? I know it wasn't defined in the JIRA, but might be beneficial. Not a blocker though.

akoushke: I think the log should be good enough.

lalli-flores: You should have set this to empty string `''`, not space `' '`

lalli-flores: I like the alt idea, something like `Unknown user`

PR #41, opened 2019-09-24
docs(README): add WebexDataProvider to sample

adamweeks:


README.md
```diff
@@ -90,7 +90,7 @@ import '@webex/components/dist/webexComponents.css';
 
 import React from 'react';
 import ReactDOM from 'react-dom';
-import {WebexJSONAdapter, WebexAvatar} from '@webex/components';
+import {WebexJSONAdapter, WebexAvatar, WebexDataProvider} from '@webex/components';
```
akoushke: could it be 
```js
import {WebexAvatar, WebexDataProvider, WebexJSONAdapter} from '@webex/components';
```

not a blocker

adamweeks: There wasn't an order previously, otherwise I would have. Fixing.

PR #42, opened 2019-09-25
Activity stream loads previous activities of a room

lalli-flores:
Tested with "Hellen Keller" data:
https://github.com/lalli-flores/components/tree/hellen-keller

I would prefer to have this tested with different data set.

### DIFF
--- a/src/adapters/RoomsAdapter.js
+++ b/src/adapters/RoomsAdapter.js
@@ -1,6 +1,23 @@
+/* eslint-disable valid-jsdoc */
+/* eslint-disable no-unused-vars */
 import {throwError} from 'rxjs';
 
 import WebexAdapter from './WebexAdapter';
+
+/**
+ * A Room object with details about the room.
+ *
+ * @typedef {Object}  Room
+ * @property {string}    ID     The room identifier.
+ * @property {string}    title  The room title.
+ * @property {RoomType}  type   The type of the room. @see RoomType enum
+ */
+
+/**
+ * @typedef {Object}  ActivityDate
+ * @param {string}  date  Date that will render on a time ruler. Must be a valid date-time string.
+ *                        See https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/parse#Date_Time_String_Format
+ */
 
 /**
  * Enum for room type values.
@@ -17,23 +34,12 @@ export const RoomType = {
  * This is a base class that defines the interface that maps room data.
  * Developers that want to extend `RoomsAdapter` must implement all of its methods,
  * adhering to the exact parameters and structure of the returned objects.
+ *
+ * `RoomsAdapter` handles data of a room such as details about that room but also
+ * activities that happened/will happen within that room. Activities are expected
+ * to be lazy loaded, and therefore, chunked.
  */
 export default class RoomsAdapter extends WebexAdapter {
-  /**
-   * A Room object with details about the room.
-   *
-   * @typedef {Object} Room
-   * @property {string}    ID      The room identifier.
-   * @property {string}    title   The room title.
-   * @property {RoomType}  type    The type of the room. @see RoomType enum
-   */
-
-  /**
-   * @typedef {Object} ActivityData
-   * @param {string} date The date to render a time ruler for. Must be a valid date-time string
-   *     https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/parse#Date_Time_String_Format
-   */
-
   /**
    * Returns an observable that emits room data of the given ID.
    *
@@ -41,32 +47,44 @@ export default class RoomsAdapter extends WebexAdapter {
    * @returns {Observable.<Room>}
    * @memberof RoomsAdapter
    */
-  // eslint-disable-next-line no-unused-vars
   getRoom(ID) {
     return throwError(new Error('getRoom(ID) must be defined in RoomsAdapter'));
   }
 
   /**
-   * Returns an observable that emits an array of previous activity IDs of the given roomID.
+   * Returns an observable that emits an array of current and future activities of the given roomID.
    *
-   * @param {string} ID  ID of the room to get.
-   * @returns {Observable.<Array.<string>>}
+   * @param {string} ID  ID of the room for which to get activities.
+   * @returns {Observable.<Array.<string|ActivityDate>>}
    * @memberof RoomsAdapter
    */
-  // eslint-disable-next-line no-unused-vars
+  getRoomActivities(ID) {
+    return throwError(new Error('getRoomActivities(ID) must be defined in RoomsAdapter'));
+  }
+
+  /**
+   * Returns an observable that emits an array of the next chunk of previous
+   * activity data of the given roomID. If `hasMoreActivities` returns false,
+   * the observable will complete.
+   * **Previous activity data must be sorted oldest-to-latest.**
+   *
+   * @param {string} ID  ID of the room for which to get activities.
+   * @returns {Observable.<Array.<string|ActivityDate>>}
+   * @memberof RoomsAdapter
+   */
   getPreviousRoomActivities(ID) {
     return throwError(new Error('getPreviousRoomActivities(ID) must be defined in RoomsAdapter'));
   }
 
   /**
-   * Returns an observable that emits an array of current activities of the given roomID.
+   * Returns `true` if there are more activities to load from the room of the given ID.
+   * Otherwise, it returns `false`.
    *
-   * @param {string} ID  ID of the room to get.
-   * @returns {Observable.<Array.<string|ActivityData>>}
+   * @param {string} ID ID of the room for which to verify activities.
+   * @returns {boolean}
    * @memberof RoomsAdapter
    */
-  // eslint-disable-next-line no-unused-vars
-  getRoomActivities(ID) {
-    return throwError(new Error('getRoomActivities(ID) must be defined in RoomsAdapter'));
+  hasMoreActivities(ID) {
+    throw new Error('hasMoreActivities(ID) must be defined in RoomsAdapter');
   }
 }
--- a/src/adapters/RoomsAdapter.test.js
+++ b/src/adapters/RoomsAdapter.test.js
@@ -11,25 +11,11 @@ describe('Rooms Adapter Interface', () => {
     expect(rxjs.isObservable(roomsAdapter.getRoom())).toBeTruthy();
   });
 
-  test('getRoom() throws a proper error message', (done) => {
-    roomsAdapter.getRoom('id').subscribe(
+  test('getRoom() errors because it needs to be defined', (done) => {
+    roomsAdapter.getRoom('ID').subscribe(
       () => {},
       (error) => {
         expect(error.message).toBe('getRoom(ID) must be defined in RoomsAdapter');
-        done();
-      }
-    );
-  });
-
-  test('getPreviousRoomActivities() returns an observable', () => {
-    expect(rxjs.isObservable(roomsAdapter.getPreviousRoomActivities())).toBeTruthy();
-  });
-
-  test('getPreviousRoomActivities() throws a proper error message', (done) => {
-    roomsAdapter.getPreviousRoomActivities('id').subscribe(
-      () => {},
-      (error) => {
-        expect(error.message).toBe('getPreviousRoomActivities(ID) must be defined in RoomsAdapter');
         done();
       }
     );
@@ -39,7 +25,7 @@ describe('Rooms Adapter Interface', () => {
     expect(rxjs.isObservable(roomsAdapter.getRoomActivities())).toBeTruthy();
   });
 
-  test('getRoomActivities() throws a proper error message', (done) => {
+  test('getRoomActivities() errors because it needs to be defined', (done) => {
     roomsAdapter.getRoomActivities('id').subscribe(
       () => {},
       (error) => {
@@ -47,6 +33,27 @@ describe('Rooms Adapter Interface', () => {
         done();
       }
     );
+  });
+
+  test('getPreviousRoomActivities() returns an observable', () => {
+    expect(rxjs.isObservable(roomsAdapter.getRoomActivities())).toBeTruthy();
+  });
+
+  test('getPreviousRoomActivities() errors because it needs to be defined', (done) => {
+    roomsAdapter.getPreviousRoomActivities('id').subscribe(
+      () => {},
+      (error) => {
+        expect(error.message).toBe('getPreviousRoomActivities(ID) must be defined in RoomsAdapter');
+        done();
+      }
+    );
+  });
+
+  test('hasMoreActivities() errors because it needs to be defined', () => {
+    try {
+      expect(roomsAdapter.hasMoreActivities('ID')).toThrowError();
+      // eslint-disable-next-line no-empty
+    } catch (error) {}
   });
 
   afterEach(() => {
--- a/src/adapters/RoomsJSONAdapter.js
+++ b/src/adapters/RoomsJSONAdapter.js
@@ -22,9 +22,14 @@ import RoomsAdapter from './RoomsAdapter';
 export default class RoomsJSONAdapter extends RoomsAdapter {
   constructor(datasource) {
     super(datasource);
+
+    this.dataChunkSize = 4; // Arbitrary chunk size
+    this.lastDataIndex = {}; // Keeps track of last index returned for each room
+
     this.getRoom = this.getRoom.bind(this);
-    this.getPreviousRoomActivities = this.getPreviousRoomActivities.bind(this);
     this.getRoomActivities = this.getRoomActivities.bind(this);
+    this.getPreviousRoomActivities = this.getPreviousRoomActivities.bind(this);
+    this.hasMoreActivities = this.hasMoreActivities.bind(this);
   }
 
   /**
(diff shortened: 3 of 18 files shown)


.storybook/setupTests.js
```diff
@@ -2,6 +2,8 @@ import React from 'react';
 import Enzyme, {shallow, render, mount} from 'enzyme';
 import Adapter from 'enzyme-adapter-react-16';
 import * as rxjs from 'rxjs';
+import 'core-js/stable';
```
akoushke: what are these for?

src/components/hooks/useActivityStream.js
```diff
@@ -1,53 +1,129 @@
...
+
+  const handleScroll = (event) => {
+    // Only load more data when user has reached the top
+    const isViewportTop = event.target.scrollTop === 0;
+
+    if (!showLoader && isViewportTop && roomsAdapter.hasMoreActivities(roomID)) {
+      setShowLoader(true);
+
+      // Throttle scroll
+      setTimeout(async () => {
```
akoushke: is the `setTimeout` still valid here? I don't think we would need it anymore, since we are awaiting.

src/components/hooks/useActivityStream.js
```diff
@@ -1,53 +1,129 @@
...
+        loadPreviousActivities();
+        // if there are no more activities to load, scroll to last activity
+      } else {
+        lastElementRef.current.scrollIntoView();
+        setScrollToLastActivity(false);
+      }
+    } else {
+      // Scroll to bottom once after loading enough content
+      // eslint-disable-next-line no-lonely-if
+      if (scrollToLastActivity) {
```
akoushke: why not `else if` ?

src/data/activities.json
```diff
@@ -62,5 +62,181 @@
     "personID": "default",
     "created": "old",
     "displayHeader": true
+  },
+  "hellen-keller-1": {
```
akoushke: great, more data (y)

src/components/hooks/useActivityStream.js
```diff
@@ -1,53 +1,129 @@
...
+        lastElementRef.current.scrollIntoView();
+        setScrollToLastActivity(false);
+      }
+    }
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [activities]);
+
+  // Attach scroll event listener and subscribe to future updates on load
+  useEffect(() => {
```
akoushke: so two `useEffect`? elaborate a bit

src/components/WebexActivityStream/WebexActivityStream.scss
```diff
@@ -1,3 +1,71 @@
+// For Storybook, so that activity stream loads content properly
+#root {
```
akoushke: why don't you put this at the top level? this might get lost here. maybe `scss` file to be imported into setupFile

.storybook/setupTests.js
akoushke: maybe a short comment.

src/adapters/RoomsAdapter.js
```diff
@@ -47,26 +63,42 @@ export default class RoomsAdapter extends WebexAdapter {
...
+   * Returns [] if `hasMoreActivities` returns false, otherwise it returns an
+   * array of the next chunk of previous activity data of the given roomID.
+   *
+   * The next chunk is based on the adapter's `chunkSize`.
+   *
+   * @param {string} ID  ID of the room for which to get activities.
+   * @returns {Array.<string|ActivityDate>}
    * @memberof RoomsAdapter
    */
   // eslint-disable-next-line no-unused-vars
```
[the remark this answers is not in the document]
lalli-flores: I like that

src/adapters/RoomsAdapter.test.js
```diff
@@ -11,8 +11,8 @@ describe('Rooms Adapter Interface', () => {
     expect(rxjs.isObservable(roomsAdapter.getRoom())).toBeTruthy();
   });
 
-  test('getRoom() throws a proper error message', (done) => {
-    roomsAdapter.getRoom('id').subscribe(
+  test('getRoom() throws error since it is not defined', (done) => {
```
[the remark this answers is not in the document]
lalli-flores: Updated to `because it needs to be defined`. I think it's more explicit why.

src/adapters/RoomsAdapter.test.js
```diff
@@ -21,25 +21,11 @@ describe('Rooms Adapter Interface', () => {
...
-      }
-    );
-  });
-
   test('getRoomActivities() returns an observable', () => {
     expect(rxjs.isObservable(roomsAdapter.getRoomActivities())).toBeTruthy();
   });
 
-  test('getRoomActivities() throws a proper error message', (done) => {
+  test('getRoomActivities() throws error since it is not defined', (done) => {
```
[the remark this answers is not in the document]
lalli-flores: Updated to `because it needs to be defined`. I think it's more explicit why.

.storybook/setupTests.js
lalli-flores: They are [polyfills](https://babeljs.io/docs/en/next/babel-polyfill.html) so we can use `async`/`await`.
I changed implementation to observable, so this will not be necessary anymore.

src/adapters/RoomsJSONAdapter.js
```diff
@@ -22,9 +22,8 @@ import RoomsAdapter from './RoomsAdapter';
 export default class RoomsJSONAdapter extends RoomsAdapter {
   constructor(datasource) {
     super(datasource);
-    this.getRoom = this.getRoom.bind(this);
-    this.getPreviousRoomActivities = this.getPreviousRoomActivities.bind(this);
-    this.getRoomActivities = this.getRoomActivities.bind(this);
+
```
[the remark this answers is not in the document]
lalli-flores: Added bindings again

src/adapters/RoomsAdapter.test.js
```diff
@@ -49,6 +35,21 @@ describe('Rooms Adapter Interface', () => {
...
+  test('getPreviousRoomActivities() returns a promise rejection since it is not defined', async () => {
+    try {
+      await roomsAdapter.getPreviousRoomActivities('ID');
+    } catch (error) {
+      expect(error.message).toEqual('getPreviousRoomActivities(ID) must be defined in RoomsAdapter');
+    }
+  });
+
+  test('hasMoreActivities() throws error since it is not defined', () => {
+    try {
```
adamweeks: Does this need to be in a try/catch?

src/components/WebexActivityStream/WebexActivityStream.scss
lalli-flores: I tried that, doesn't seem to work.

src/components/WebexActivityStream/WebexActivityStream.js
```diff
@@ -194,8 +197,10 @@ export default function WebexActivityStream({roomID}) {
   });
 
   return (
-    <div className="activity-stream">
+    <div className="activity-stream" ref={activityStreamRef}>
+      {showLoader ? <div className="activity-stream-loader" /> : null}
```
adamweeks: I usually see it as:
```js
{ showLoader && <div }
```
but this works as well, not a blocker.

src/components/hooks/useActivityStream.js
lalli-flores: This is how we "fake wait". Data loads too fast and you can't see the loader. 
In general, I think it makes the UX a little better anyway, since we indicate to the user that we actually doing something, even when data comes through really fast. It's more for acknowledgment

src/components/hooks/useActivityStream.js
lalli-flores: Logic has changed, no longer relevant.

src/components/hooks/useActivityStream.js
lalli-flores: Because there are two different sets of side-effects:
1) Side effects to execute only once, on load
2) Side effects when there is a change to the activities

src/data/activities.json
lalli-flores: Totally added it by accident, and will remove it

src/components/hooks/useActivityStream.js
```diff
@@ -1,53 +1,129 @@
...
+
+  // Keep requesting previous activities until activity stream is filled
   useEffect(() => {
-    const subscription = roomsAdapter.getRoomActivities(roomID).subscribe((activities) => {
-      dispatch({type: APPEND_ACTIVITIES, payload: activities});
+    const activityStreamBoundaries = elementRef.current.getBoundingClientRect();
+    const lastActivityBoundaries = lastElementRef.current.getBoundingClientRect();
+    const hasMoreActivities = roomsAdapter.hasMoreActivities(roomID);
+
+    if (lastActivityBoundaries.bottom < activityStreamBoundaries.bottom) {
```
adamweeks: Is this when it is at the top? I need more comments to understand what is happening here please

src/adapters/RoomsAdapter.test.js
lalli-flores: Yes, otherwise it literally errors out the process.

src/adapters/RoomsAdapter.test.js
```diff
@@ -49,6 +35,20 @@ describe('Rooms Adapter Interface', () => {
     );
   });
 
+  test('getPreviousRoomActivities() throws error since it is not defined', () => {
```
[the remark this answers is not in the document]
lalli-flores: No longer applicable with observables.

src/components/hooks/useActivityStream.js
lalli-flores: Added comment for clarity, let me know if that helps explain

src/components/hooks/usePreviousActivities.js
```diff
@@ -0,0 +1,67 @@
...
+  const [lastActivityBottom, setLastActivityBottom] = useState(0);
+
+  useEffect(() => {
+    const activityStreamBoundaries = elementRef.current.getBoundingClientRect();
+    const lastActivityBoundaries = lastElementRef.current.getBoundingClientRect();
+
+    // If the bottom position of the last activity is less than
+    // the bottom position of the container that means that the last
+    // activity is visible and there is not enough content to overflow
+    // the container
```
adamweeks: Does this mean that if the window is resized to be bigger it will automatically load more activities?

src/components/WebexActivityStream/WebexActivityStream.js
```diff
@@ -179,8 +180,16 @@ TimeRuler.propTypes = {
 };
 
 export default function WebexActivityStream({roomID}) {
+  const [activitiesData, dispatch] = useActivityStream(roomID);
+  const loadPreviousActivities = (previousActivities) => {
```
akoushke: why is this here instead of being inside the hook?

lalli-flores: Because two different hooks need the same function

src/components/hooks/usePreviousActivities.js
lalli-flores: It should adjust how much content it loads based on the size of the window, but once it has scrolled to the last activity - meaning it ran once - it will not load content again.

src/adapters/RoomsAdapter.js
```diff
@@ -17,56 +33,55 @@ export const RoomType = {
...
   }
 
   /**
-   * Returns an observable that emits an array of current activities of the given roomID.
+   * Returns `true` if there are more activities to load from the room of the given ID.
+   * Otherwise, it returns `false`.
    *
-   * @param {string} ID  ID of the room to get.
-   * @returns {Observable.<Array.<string|ActivityData>>}
+   * @param {string} ID ID of the room for which to verify activities.
```
adamweeks: Missing the `@returns` statement

src/components/WebexActivityStream/WebexActivityStream.js
```diff
@@ -179,8 +180,16 @@ TimeRuler.propTypes = {
...
+    dispatch({type: PREPEND_ACTIVITIES, payload: previousActivities});
+  };
+
   const {title, roomType} = useRoom(roomID);
-  const activitiesData = useActivityStream(roomID);
+  const activityStreamRef = useRef(null);
```
adamweeks: 👍

src/components/hooks/useActivityScroll.js
```diff
@@ -0,0 +1,66 @@
...
+  const {roomsAdapter} = useContext(AdapterContext);
+  const [showLoader, setShowLoader] = useState(false);
+
+  useEffect(() => {
+    let scrollListener = empty();
+
+    if (elementRef.current) {
+      // Listen to scroll event, and on reaching the top,
+      // And load previous activities if there are any
+      scrollListener = fromEvent(elementRef.current, 'scroll')
```
adamweeks: This is an awesome use of rxjs!

src/components/hooks/useActivityStream.js
```diff
@@ -2,52 +2,57 @@ import {useContext, useEffect, useReducer} from 'react';
...
+ * @returns {Array.<string|ActivityDate>}
  */
-function reducer(activityIDs, action) {
-  let newActivityIDs = [];
+function reducer(activities, action) {
+  let newActivities = [];
 
   switch (action.type) {
+    case PREPEND_ACTIVITIES:
+      newActivities = action.payload.reverse().concat(activities);
```
adamweeks: Why do we need to reverse the order of the previous activities? Should the adapter return them in proper order?

src/components/hooks/usePreviousActivities.js
```diff
@@ -0,0 +1,67 @@
...
+ * the given callback with that data.
+ * It completes once there is enough data to fill the viewport of the
+ * parent container, finishing by scrolling into view the last activity.
+ *
+ * @param {string} roomID  ID of the room for which to load data.
+ * @param {object} elementRef  DOM reference of the parent component.
+ * @param {previousActivitiesCallback} callback  Callback to execute once data has been fetched.
+ * @returns {object}  DOM reference to the the last activity.
+ */
+export default function usePreviousActivities(roomID, elementRef, callback) {
```
adamweeks: If this hook wasn't defined, wouldn't the top scroll hook catch this case because the scroll top would be the viewport top?

src/components/hooks/useActivityScroll.js
```diff
@@ -0,0 +1,66 @@
...
+        .pipe(
+          filter((event) => {
+            const isViewportTop = event.target.scrollTop === 0;
+            const loadMoreItems = isViewportTop && roomsAdapter.hasMoreActivities(roomID);
```
adamweeks: When implementing the `hasMoreActivities` in the SDK adapter, we need to be sure to consider that there is no real way to know the room has or doesn't have more activities. The workaround would be that if the amount returned is less than the limit requested.

src/adapters/RoomsAdapter.js
lalli-flores: There was an eslint error because the method doesn't actually return anything. I just disabled the warning for the file and added the return statement.

src/components/hooks/usePreviousActivities.js
lalli-flores: It would not, there are 2 different but related use cases.
`useActivityScroll` listens to scroll events, and while there is more data to load, it will load more of the previous activities as you're scrolling up.
`usePreviousActivities` kinda works as an initializer, which loads as much data as necessary to try to overflow the container - based on its size - and then leaves you on the most recent activity. Without `usePreviousActivities`, `useActivityScroll` wouldn't work, because if there is no content loaded or not enough content loaded, the activity stream container will not overflow, and therefore won't have a scroll bar to trigger the scroll event (load more -even older - data). Now, if there is not enough previous activities to overflow the container to begin with, there's no need to listen for scroll events because, well, there's nothing more to load on scroll up.
Before i had all this in `useActivityStream` (combined with the subscription to future events), but i thought separating them out would be more maintainable.

src/components/hooks/useActivityScroll.js
lalli-flores: really? From the SDK, there's no way to know that there is no more activities to load?

src/components/hooks/useActivityStream.js
lalli-flores: My assumption is that most data retrieved is sorted most-recent to older. Also, displaying the data in reverse order is a function of the UX of the activity stream, and not so much of the data itself, and so, part of the UI logic.
But maybe that's not the case on how data gets sent from the backend?

src/components/hooks/usePreviousActivities.js
adamweeks: I like `useOverflowActivities` a lot. It is less confusing to me since `usePreviousActivities` sounds like it is a hook specifically for `getPreviousRoomActivities`.

src/components/hooks/useActivityStream.js
adamweeks: Just investigated and saw that the `/messages` endpoint returns data in reverse date sort (latest to oldest). Conversation endpoint does the opposite (oldest to latest). If we aren't going to do any kind of "sorting" in this hook, we need to make sure to document the expected order. 

We should expect the correct order from the adapter and reduce the code here to just prepend the array to the existing array.

src/components/hooks/useActivityScroll.js
adamweeks: Correct, take a look at the messages endpoint (https://developer.webex.com/docs/api/v1/messages/list-messages), all we get is an array and no meta-data about remaining messages 😢

src/components/hooks/usePreviousActivities.js
lalli-flores: Renamed to `useOverflowActivities`.

src/components/hooks/useActivityStream.js
lalli-flores: My initial instinct is also to have activities sorted newest-to-oldest, as we on the `/messages` API; second reason being consistent with public APIs. However, I've changed and documented the expected order as you requested.

src/components/hooks/useActivityScroll.js
lalli-flores: Oh that's fine, I think we should be able to know by not getting any more data (like receiving an empty array?). I guess we'll find out on the SDK adapter implementation.

PR #43, opened 2019-10-08
fix(ActivityStream): fix display order of previous activities

lalli-flores:


lalli-flores:
So Im assuming the data exists in some form like this: [10,9,8,7,6,5,4,3,2,1] (10 is newest activity, 1 is oldest/first activity)
Now if the data is chucked, it would be sent as: [10,9,8], [7,6,5] [4,3,2], [1]
If we wanted for the adapter to do the reversing, then it would have to send data as [8,9,10], [5,6,7], [2,3,4], [1] which I think as a developer that would most likely only work with the adapter, is not intuitive why the data has to be sent like that

PR #44, opened 2019-10-22
feat(MeetingsAdapter): add initial interface

adamweeks:
https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-99431

### DIFF
--- a/src/adapters/MeetingsAdapter.js
+++ b/src/adapters/MeetingsAdapter.js
@@ -1,0 +1,90 @@
+/* eslint-disable no-unused-vars */
+import {throwError} from 'rxjs';
+
+import WebexAdapter from './WebexAdapter';
+
+/**
+ * This is a base class that defines the interface that maps meetings data.
+ * Developers that want to extend `MeetingsAdapter` must implement all of its methods,
+ * adhering to the exact parameters and structure of the returned objects.
+ */
+
+/**
+ * A meeting object that allows users to have a WebRTC meeting.
+ *
+ * @typedef {Object} Meeting
+ * @property {string}  ID The meeting identifier.
+ * @property {string}  meetingTitle The title of the meeting.
+ * @property {Date}    meetingTimeStart The time and date of the start of the meeting.
+ * @property {Date}    meetingTimeEnd The time and date of the end of the meeting.
+ * @property {Boolean} localAudioMuted If the local user has muted their audio.
+ * @property {Boolean} localVideoMuted If the local user has muted their video.
+ * @property {Boolean} meetingJoined If the local user has joined the meeting.
+ */
+
+export default class MeetingsAdapter extends WebexAdapter {
+  /**
+   * Returns an observable that emits meeting data of the given ID.
+   *
+   * @param {string} ID  ID of the meeting to get.
+   * @returns {Observable.<Meeting>}
+   * @memberof MeetingsAdapter
+   */
+  getMeeting(ID) {
+    return throwError(new Error('getMeeting(ID) must be defined in MeetingsAdapter'));
+  }
+
+  /**
+   * Returns a promise that resolved with the meeting ID that matches
+   * the requested parameters.
+   *
+   * @param {string} destinationID
+   * @param {string} destinationType
+   * @returns {Observable.<string>} Returns an observable of the meeting ID.
+   * @memberof MeetingsAdapter
+   */
+  getMeetingID(destinationID, destinationType) {
+    return throwError(new Error('getMeetingID(destinationID, destinationType) must be defined in MeetingsAdapter'));
+  }
+
+  /**
+   * Sets the local audio muted state of a particular meetingID.
+   * Returns a promise that is resolved when the set completes.
+   * Also emits an update to the getMeeting observable with the state update to `localAudioMuted`.
+   *
+   * @param {Boolean} audioMuted
+   * @param {string} meetingID
+   * @returns {Observable}
+   * @memberof MeetingsAdapter
+   */
+  setLocalAudioMuted(audioMuted, meetingID) {
+    return throwError(new Error('setLocalAudioMuted(audioMuted, meetingID) must be defined in MeetingsAdapter'));
+  }
+
+  /**
+   * Sets the local video muted state of a particular meetingID.
+   * Returns a promise that is resolved when the set completes.
+   * Also emits an update to the getMeeting observable with the state update to `localVideoMuted`.
+   *
+   * @param {Boolean} videoMuted
+   * @param {string} meetingID
+   * @returns {Observable}
+   * @memberof MeetingsAdapter
+   */
+  setLocalVideoMuted(videoMuted, meetingID) {
+    return throwError(new Error('setLocalAudioMuted(videoMuted, meetingID) must be defined in MeetingsAdapter'));
+  }
+
+  /**
+   * Joins a meeting with the local user.
+   * Returns a promise that is resolved when the set completes.
+   * Also emits an update to the getMeeting observable with the state update to `meetingJoined`.
+   *
+   * @param {string} meetingID
+   * @returns {Observable}
+   * @memberof MeetingsAdapter
+   */
+  joinMeeting(meetingID) {
+    return throwError(new Error('joinMeeting(meetingID) must be defined in MeetingsAdapter'));
+  }
+}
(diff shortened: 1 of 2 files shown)


src/adapters/MeetingsAdapter.js
```diff
@@ -0,0 +1,90 @@
...
+   * Sets the local audio muted state of a particular meetingID.
+   * Returns a promise that is resolved when the set completes.
+   * Also emits an update to the getMeeting observable with the state update to `localAudioMuted`.
+   *
+   * @param {Boolean} audioMuted
+   * @param {string} meetingID
+   * @memberof MeetingsAdapter
+   */
+  // eslint-disable-next-line no-unused-vars
+  setLocalAudioMuted(audioMuted, meetingID) {
```
lalli-flores: Shouldn't this be more like `toggleLocalAudioMute`? Otherwise, how do you unmute?
Also, I would rather this return an observable that completes, for consistency with other asynchronous-type methods.

src/adapters/MeetingsAdapter.js
```diff
@@ -0,0 +1,90 @@
...
+   * Sets the local video muted state of a particular meetingID.
+   * Returns a promise that is resolved when the set completes.
+   * Also emits an update to the getMeeting observable with the state update to `localVideoMuted`.
+   *
+   * @param {Boolean} videoMuted
+   * @param {string} meetingID
+   * @memberof MeetingsAdapter
+   */
+  // eslint-disable-next-line no-unused-vars
+  setLocalVideoMuted(videoMuted, meetingID) {
```
lalli-flores: Shouldn't this be more like `toggleLocalVideoMute`? Otherwise, how do you unmute video?
Also, I would rather this return an observable that completes, for consistency with other asynchronous-type methods.

src/adapters/MeetingsAdapter.js
```diff
@@ -0,0 +1,90 @@
...
+  /**
+   * Joins a meeting with the local user.
+   * Returns a promise that is resolved when the set completes.
+   * Also emits an update to the getMeeting observable with the state update to `meetingJoined`.
+   *
+   * @param {string} meetingID
+   * @memberof MeetingsAdapter
+   */
+  // eslint-disable-next-line no-unused-vars
+  joinMeeting(meetingID) {
```
lalli-flores: I think it would make sense for this to turn the same observables as `getMeeting`, this way, when a component joins the meeting, they also have information about the meeting they joined, and receive updates as thing change.

src/adapters/MeetingsAdapter.js
```diff
@@ -0,0 +1,90 @@
...
+   * Returns a promise that resolved with the meeting ID that matches
+   * the requested parameters.
+   *
+   * @param {string} destinationID
+   * @param {string} destinationType
+   * @returns {Promise.<string>}
+   * @memberof MeetingsAdapter
+   */
+  // eslint-disable-next-line no-unused-vars
+  getMeetingID(destinationID, destinationType) {
```
lalli-flores: I would rather this return an observable that completes after sending the meeting ID. For consistency with other asynchronous-type methods, as well as composability with other observable streams.

src/adapters/MeetingsAdapter.js
```diff
@@ -0,0 +1,90 @@
...
+
+import WebexAdapter from './WebexAdapter';
+
+/**
+ * This is a base class that defines the interface that maps meetings data.
+ * Developers that want to extend `MeetingsAdapter` must implement all of its methods,
+ * adhering to the exact parameters and structure of the returned objects.
+ */
+export default class MeetingsAdapter extends WebexAdapter {
+  /**
```
lalli-flores: Can you move the object definition outside the class, like in [other adapters](https://github.com/webex/components/blob/master/src/adapters/RoomsAdapter.js#L7)?

src/adapters/MeetingsAdapter.js
```diff
@@ -0,0 +1,90 @@
...
+   * A meeting object that allows users to have a WebRTC meeting.
+   *
+   * @typedef {Object} Meeting
+   * @property {string}  ID The meeting identifier.
+   * @property {string}  meetingTitle The title of the meeting.
+   * @property {Date}    meetingTimeStart The time and date of the start of the meeting.
+   * @property {Date}    meetingTimeEnd The time and date of the end of the meeting.
```
lalli-flores: Is there a concept of meeting "duration"? i would rather have that information and calculate the end time.

src/adapters/MeetingsAdapter.js
```diff
@@ -0,0 +1,90 @@
...
+   * @property {Boolean} localAudioMuted If the local user has muted their audio.
```
lalli-flores: I kinda don't like having `localAudioMuted`, `localVideoMuted` and `meetingJoined` mixed with the information of the meeting because they are states. Would it make sense to have `Meeting` object with the info and another `LocalMeetingState` with these states?
`LocalMeetingState` would be returned by `joinMeeting` method

src/adapters/MeetingsAdapter.js
adamweeks: Sure, I copied from https://github.com/webex/components/blob/master/src/adapters/ActivitiesAdapter.js#L14 which has it defined within.

src/adapters/MeetingsAdapter.js
adamweeks: Sure

src/adapters/MeetingsAdapter.js
adamweeks: We've discovered that giving users a `toggle` function gets meetings into a state of conflict when using it multiple times in a row before waiting on the result. That is why the meetings plugin only has `muteAudio` and `unmuteAudio` functions requiring the developer to explicitly define what they want the end state to be after the result of the command.

I will change the promises to observables.

adamweeks: >  Otherwise, how do you unmute?

`setLocalAudioMuted(false)`. I'm not against having a `setLocalAudioUnmuted` function instead.

src/adapters/MeetingsAdapter.js
adamweeks: Addressed here: https://github.com/webex/components/pull/44/files#r337701539

src/adapters/MeetingsAdapter.js
adamweeks: My thoughts were that the meeting state would be reported along with all of the meeting info in the meeting observer.

src/adapters/MeetingsAdapter.js
```diff
@@ -0,0 +1,90 @@
...
+   */
+
+  /**
+   * Returns an observable that emits meeting data of the given ID.
+   *
+   * @param {string} ID  ID of the meeting to get.
+   * @returns {Observable.<Meeting>}
+   * @memberof MeetingsAdapter
+   */
+  // eslint-disable-next-line no-unused-vars
```
akoushke: being nitpicky here, but can you define this at the top of the module?

akoushke:
I think Lalli mentioned this already, but I was wondering if we can still return an observable instead of a promise for all the methods? any specific reason we can't use observable?

src/adapters/MeetingsAdapter.js
adamweeks: I guess this gets back to how are we going to handle actions in the adapters. Since we're not really expecting a return object here, wouldn't it be more direct to emit the change in the `getMeeting` observable?

adamweeks:
> I think Lalli mentioned this already, but I was wondering if we can still return an observable instead of a promise for all the methods? any specific reason we can't use observable?

@akoushke  Done. There's no real reason to subscribe to an action function other than consistency, but I'm not fighting it.

src/adapters/MeetingsAdapter.js
adamweeks: I was trying to reduce the amount of work required on the front end since this will just be converted to local timezone and rendered.

adamweeks:
> I would like to have the return values of methods modified.

Done

lalli-flores:
The [component-adapter-repository](https://github.com/webex/component-adapter-interfaces) seems to be ready to add interfaces to. Shall we close this PR, finalize the interface and open a PR against that repository?

arun3528:
I personally like to split the mute and unmute to different function when exposing to developers .

adamweeks:
Will be migrating this to component adapter repo.

PR #45, opened 2019-10-28
Migrate interfaces

lalli-flores:
Removing interfaces from adapter repository and importing from `@webex/component-adapter-interface`


lalli-flores:
> Looks good. I would say that `fix(package): install component adapter interfaces` should be `feat(package)` instead but not a blocker.

Used fix based on your comment in the space, I think we should standardized to that.
Screen Shot 2019-10-29 at 11 15 27 AM

adamweeks:
> > Looks good. I would say that `fix(package): install component adapter interfaces` should be `feat(package)` instead but not a blocker.
> 
> Used fix based on your comment in the space, I think we should standardized to that.
> Screen Shot 2019-10-29 at 11 15 27 AM

That was just for bumping versions of packages, not adding new ones. Since there are other "feat" commits in this PR, the version will bump as expected anyway.

PR #47, opened 2019-10-30
Webex Activity component refactor

lalli-flores:
Moved `ActivityHeader` into its own component files, outside of `WebexActivity`.

Additionally, fixed date mocking in unit tests.

### DIFF
--- a/.storybook/setupTests.js
+++ b/.storybook/setupTests.js
@@ -1,9 +1,14 @@
 import React from 'react';
 import Enzyme, {shallow, render, mount} from 'enzyme';
 import Adapter from 'enzyme-adapter-react-16';
+import MockDate from 'mockdate';
 import * as rxjs from 'rxjs';
 
 import jsonData from '../src/data';
+
+// Fix date-time for tests
+process.env.TZ = 'GMT';
+MockDate.set('2019-08-01T22:00:00.000Z');
 
 // React 16 Enzyme adapter
 Enzyme.configure({adapter: new Adapter()});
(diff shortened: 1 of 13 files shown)


.storybook/setupTests.js
```diff
@@ -1,10 +1,15 @@
 import React from 'react';
 import Enzyme, {shallow, render, mount} from 'enzyme';
 import Adapter from 'enzyme-adapter-react-16';
+import MockDate from 'mockdate';
```
adamweeks: This is what we are using in the widgets repo. Works really well.

src/components/WebexActivity/ActivityHeader.test.js
```diff
@@ -0,0 +1,57 @@
...
+
+import ActivityHeader, {formatMessageDate} from './ActivityHeader';
+
+jest.mock('../hooks/usePerson');
+
+describe('Activity Header component', () => {
+  let today, sunday;
+
+  beforeEach(() => {
+    today = startOfToday(); // Mock date is set to Thu, Aug 1 2019
```
adamweeks: Can you put that it starts at 00:00 as well?

.storybook/setupTests.js
lalli-flores: I tried using it before when I first opened my PR, and I couldn't seem to make it work. I realized recently it was because `date-fns` was being mocked, and also the `Date` class in this file 🤦🏻‍♀️

src/components/WebexActivity/ActivityHeader.test.js
lalli-flores: updated

adamweeks:
I like pulling this out into its own testable component. Keeps our components basic and less complex. 

img

PR #49, opened 2019-11-01
Activity Stream component refactor

lalli-flores:
Refactor of Webex Activity Stream of helper components.
Added missing unit tests, upping the test coverage 🎆


src/components/WebexActivityStream/TimeRuler.js
```diff
@@ -0,0 +1,53 @@
...
+  }
+
+  return formattedDate;
+}
+
+/**
+ * Time Ruler component is used by the activity stream
+ * to separate activities from different days.
+ *
+ * @param {object} props
```
adamweeks: Could you add a param for the `props.date` item?

lalli-flores: Is this only for this component or all of them? I left it as generic "props" for all components, hoping the `propTypes` are enough

PR #53, opened 2019-11-19
feat(MeetingsJSONAdapter): add mute-audio meeting control

lalli-flores:


src/adapters/MeetingsJSONAdapter.js
```diff
@@ -44,5 +69,66 @@ export default class MeetingsJSONAdapter extends MeetingsAdapter {
...
+
+  /**
+   * Toggles muting the local audio media stream track.
+   * Used by "mute-audio" meeting control.
+   *
+   * @param {string} ID  ID of the meeting for which to mute local audio.
+   * @memberof MeetingsJSONAdapter
+   * @private
+   */
+  toggleMuteAudio(ID) {
```
adamweeks: As discussed before, we've seen that "toggling" async media operations causes user confusion because multiple clicks causing toggles to fire repeatedly. I know with the json adapter, we don't have async operations, but I'd prefer to not have this as a precedent. 

A more direct approach with the action accepting the desired state would be better:
`meetingsJSONAdapter.muteAudio(meetingID, true)`

src/adapters/MeetingsJSONAdapter.js
```diff
@@ -35,7 +60,7 @@ export default class MeetingsJSONAdapter extends MeetingsAdapter {
    * @memberof MeetingsJSONAdapter
    */
   getMeeting(ID) {
-    return Observable.create((observer) => {
+    const json$ = Observable.create((observer) => {
```
akoushke: can we change this to `getMeeting$` to have a consistent naming pattern with the SDK for now?

src/adapters/MeetingsJSONAdapter.js
lalli-flores: We discussed this, but putting here for the record.
It is not possible for the action to take a second parameter, as there is no way for the component to know what the "right" state in a generic manner (e.g. any control with different and potentially different number of states).
The async operations are an implementation detail that shouldn't bubble on design, so we came up with solutions for adapters like the SDK:
* Set the control to `DISABLE` while we get acknowledgement from the server and then set it to the appropriate state
* Throttle or ignore subsequent request until we get server acknowledgement

src/adapters/MeetingsJSONAdapter.js
```diff
@@ -66,14 +70,109 @@ export default class MeetingsJSONAdapter extends MeetingsAdapter {
       observer.complete();
     });
 
-    return getMeeting$.pipe(
+    // Attach local media stream if meeting `localVideo` property is not `null`.
```
akoushke: 😶

src/adapters/MeetingsJSONAdapter.test.js
```diff
@@ -1,53 +1,119 @@
...
   beforeEach(() => {
     [meetingID] = Object.keys(meetings);
     meetingsJSONAdapter = new MeetingsJSONAdapter(meetings);
     meetingsJSONAdapter.getLocalVideo = jest.fn(() => Promise.resolve('mock-stream'));
   });
 
-  test('getMeeting() returns an observable', () => {
-    expect(rxjs.isObservable(meetingsJSONAdapter.getMeeting())).toBeTruthy();
+  afterEach(() => {
+    if (subscription) {
```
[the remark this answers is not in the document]
lalli-flores: Yes, the tests that don't do subscriptions, it still runs for them

src/adapters/MeetingsJSONAdapter.test.js
```diff
@@ -1,53 +1,119 @@
...
+            icon: 'microphone-muted',
+            tooltip: 'Unmute',
+            state: 'active',
+          });
+          done();
+        });
+    });
+
+    test('inactive control display values', (done) => {
+      rxjs.fromEvent = jest.fn(() => {
```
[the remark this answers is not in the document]
lalli-flores: Because I'm mocking it, not using the real implementation

src/adapters/MeetingsJSONAdapter.js
```diff
@@ -22,28 +26,28 @@ import {MeetingsAdapter} from '@webex/component-adapter-interfaces';
...
-    try {
-      stream = await navigator.mediaDevices.getUserMedia(constraints);
-    } catch (reason) {
-      // eslint-disable-next-line no-console
-      console.error('Meetings JSON adapter can not display the user local stream', reason);
-    }
+    this.meetingControls = {};
 
-    return stream;
+    this.meetingControls[MUTE_AUDIO_CONTROL] = {
```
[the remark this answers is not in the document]
lalli-flores: I don't understand your suggestion?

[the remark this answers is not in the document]
akoushke: Screen Shot 2019-11-25 at 10 27 46 AM

ID: `mute-audio`  is missing from this object.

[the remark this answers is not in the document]
lalli-flores: Good catch 👍🏻

src/adapters/MeetingsJSONAdapter.test.js
```diff
@@ -1,53 +1,119 @@
...
 
   beforeEach(() => {
     [meetingID] = Object.keys(meetings);
     meetingsJSONAdapter = new MeetingsJSONAdapter(meetings);
     meetingsJSONAdapter.getLocalVideo = jest.fn(() => Promise.resolve('mock-stream'));
   });
 
-  test('getMeeting() returns an observable', () => {
-    expect(rxjs.isObservable(meetingsJSONAdapter.getMeeting())).toBeTruthy();
+  afterEach(() => {
```
akoushke: we have 2 `afterEaches` now.

src/adapters/MeetingsJSONAdapter.js
```diff
@@ -66,14 +71,108 @@ export default class MeetingsJSONAdapter extends MeetingsAdapter {
...
+      ID: MUTE_AUDIO_CONTROL,
+      icon: 'microphone-muted',
+      tooltip: 'Unmute',
+      state: MeetingControlState.ACTIVE,
+    };
+
+    const default$ = Observable.create((observer) => {
+      const meeting = this.datasource[ID];
+
+      if (meeting && meeting.localAudio) {
```
akoushke: I think if the meeting is not defined, we should throw an error and to mix it up with `meeting.localAudio` check.

src/adapters/MeetingsJSONAdapter.js
```diff
@@ -66,14 +71,108 @@ export default class MeetingsJSONAdapter extends MeetingsAdapter {
...
+
+        observer.next(isMuted ? muted : unmuted);
+      }
+
+      observer.complete();
+    });
+
+    const muteEvent$ = fromEvent(document, MUTE_AUDIO_CONTROL).pipe(
+      filter((event) => event.detail.ID === ID),
+      map((event) => {
```
akoushke:
```js
map((event) => event.detail.localAudio.muted ? muted :unmuted;
```

src/adapters/MeetingsJSONAdapter.test.js
akoushke: and honestly, I prefer to have this at the end for *THIS* repo #consistency.

src/adapters/MeetingsJSONAdapter.js
```diff
@@ -22,28 +26,29 @@ import {MeetingsAdapter} from '@webex/component-adapter-interfaces';
...
-    try {
-      stream = await navigator.mediaDevices.getUserMedia(constraints);
-    } catch (reason) {
-      // eslint-disable-next-line no-console
-      console.error('Meetings JSON adapter can not display the user local stream', reason);
-    }
+    this.meetingControls = {};
 
-    return stream;
+    this.meetingControls[MUTE_AUDIO_CONTROL] = {
```
akoushke: can you just combine the line 45 with 47 to define the `meetingControls` in one place?

.storybook/setupTests.js
```diff
@@ -27,3 +27,5 @@ jest.spyOn(React, 'useContext').mockImplementation((context) => {
     roomsAdapter: jsonData.rooms
   };
 });
+
+jest.spyOn(document, 'dispatchEvent').mockImplementation(() => {});
```
adamweeks: Could you add a comment on what this is used for?

src/adapters/MeetingsJSONAdapter.test.js
```diff
@@ -1,53 +1,119 @@
...
+    });
   });
 
-  test('getMeeting() renders the local media if localVideo property is defined', (done) => {
-    meetingsJSONAdapter.getMeeting('localVideo').subscribe((data) => {
-      expect(data.localVideo).toEqual('mock-stream');
-      done();
+  describe('toggleMuteAudio()', () => {
+    test('dispatches a "mute-audio" event', () => {
+      meetingsJSONAdapter.toggleMuteAudio();
```
adamweeks: Doesn't this need an ID?

src/adapters/MeetingsJSONAdapter.js
lalli-flores: In order to use the constant for the attribute name, I need to use the `[]` notation

adamweeks: I like the separation of having each control defined in its own code block.

src/adapters/MeetingsJSONAdapter.test.js
lalli-flores: Totally didn't even realize the 2 `afterEach`. I combined them at the end of the `describe` block

src/adapters/MeetingsJSONAdapter.test.js
```diff
@@ -1,57 +1,133 @@
...
+    test('dispatches a "mute-audio" event', () => {
+      meetingID = 'localAudio';
+
+      meetingsJSONAdapter.toggleMuteAudio(meetingID);
+      expect(global.document.dispatchEvent).toHaveBeenCalled();
     });
   });
 
   afterEach(() => {
+    if (subscription) {
```
akoushke: you are basically doing the check for two test cases out of however many we have total. subscription should be available throughout the whole suite, since it is utilized for most of the test cases. this is how we set variables globally in most our test suites. if there is an object being utilized across multiple test cases, mind as well create them at the top level for clarity and accessibility. 

not all these objects have been utilized in all test cases but they have beed defined at the top level:

```js
[meetingID] = Object.keys(meetings);
    meetingsJSONAdapter = new MeetingsJSONAdapter(meetings);	    meetingsJSONAdapter = new MeetingsJSONAdapter(meetings);
    meetingsJSONAdapter.getLocalVideo = jest.fn(() => Promise.resolve('mock-stream'));	    meetingsJSONAdapter.getLocalVideo = jest.fn(() => Promise.resolve('mock-stream'));
```
that we could remove this check statement too. 
also, extra subscription here means not much since this is all mock data.

src/adapters/MeetingsJSONAdapter.test.js
```diff
@@ -1,57 +1,133 @@
...
+    });
   });
 
-  test('getMeeting() renders the local media if localVideo property is defined', (done) => {
-    meetingsJSONAdapter.getMeeting('localVideo').subscribe((data) => {
-      expect(data.localVideo).toEqual('mock-stream');
-      done();
+  describe('toggleMuteAudio()', () => {
+    test('dispatches a "mute-audio" event', () => {
+      meetingID = 'localAudio';
```
akoushke: we are exactly doing something that you are claiming the opposite. we are setting `meetingID` to something first and then set it to something else for this test case.

src/adapters/MeetingsJSONAdapter.test.js
akoushke: RESOLVED!!!!

PR #54, opened 2019-11-19
Webex Local Media Component

akoushke:
Add Local Media Component.

### DIFF
--- a/package.json
+++ b/package.json
@@ -40,6 +40,8 @@
   "devDependencies": {
     "@babel/core": "^7.5.5",
     "@babel/plugin-proposal-export-default-from": "^7.5.2",
+    "@babel/plugin-transform-regenerator": "^7.7.0",
+    "@babel/plugin-transform-runtime": "^7.6.2",
     "@babel/preset-env": "^7.5.5",
     "@babel/preset-react": "^7.0.0",
     "@commitlint/cli": "^8.1.0",
@@ -91,7 +93,9 @@
     ],
     "plugins": [
       "@babel/plugin-proposal-object-rest-spread",
-      "@babel/plugin-proposal-export-default-from"
+      "@babel/plugin-proposal-export-default-from",
+      "@babel/plugin-transform-regenerator",
+      "@babel/plugin-transform-runtime"
     ]
   },
   "jest": {
--- a/src/adapters/MeetingsJSONAdapter.js
+++ b/src/adapters/MeetingsJSONAdapter.js
@@ -1,6 +1,6 @@
+import {from, Observable} from 'rxjs';
+import {flatMap, map} from 'rxjs/operators';
 import {MeetingsAdapter} from '@webex/component-adapter-interfaces';
-import {Observable} from 'rxjs';
-
 /**
  * @typedef MeetingsJSON
  * @param {object} datasource An object that contains a set of meetings keyed by ID.
@@ -21,10 +21,31 @@ import {Observable} from 'rxjs';
  * }
  */
 
-/*
+/**
  * Implements the MeetingsAdapter interface with a JSON object as its datasource. See @MeetingsJSON
  */
 export default class MeetingsJSONAdapter extends MeetingsAdapter {
+  /**
+   * Returns a MediaStream object obtained from local user media.
+   * @returns {MediaStream}
+   */
+  async getLocalVideo() {
+    const constraints = {
+      video: true,
+      audio: false,
+    };
+    let stream;
+
+    try {
+      stream = await navigator.mediaDevices.getUserMedia(constraints);
+    } catch (reason) {
+      // eslint-disable-next-line no-console
+      console.error('Meetings JSON adapter can not display the user local stream', reason);
+    }
+
+    return stream;
+  }
+
   /**
    * Returns an observable that emits a Meeting object.
    * Whenever there is an update to the meeting, the observable
@@ -35,7 +56,7 @@ export default class MeetingsJSONAdapter extends MeetingsAdapter {
    * @memberof MeetingsJSONAdapter
    */
   getMeeting(ID) {
-    return Observable.create((observer) => {
+    const getMeeting$ = Observable.create((observer) => {
       if (this.datasource[ID]) {
         observer.next(this.datasource[ID]);
       } else {
@@ -44,5 +65,15 @@ export default class MeetingsJSONAdapter extends MeetingsAdapter {
 
       observer.complete();
     });
+
+    return getMeeting$.pipe(
+      flatMap((meeting) =>
+        // Attach the localMedia stream if meeting localVideo
+        // property is not `null`. Since we can not attach the
+        // MediaStream object in out JSON module, the work needs
+        // to be done here.
+        from(this.getLocalVideo()).pipe(map((localVideo) => (meeting.localVideo ? {...meeting, localVideo} : meeting)))
+      )
+    );
   }
 }
--- a/src/adapters/MeetingsJSONAdapter.test.js
+++ b/src/adapters/MeetingsJSONAdapter.test.js
@@ -7,6 +7,7 @@ describe('Meetings JSON Adapter', () => {
   beforeEach(() => {
     [meetingID] = Object.keys(meetings);
     meetingsJSONAdapter = new MeetingsJSONAdapter(meetings);
+    meetingsJSONAdapter.getLocalVideo = jest.fn(() => Promise.resolve('mock-stream'));
   });
 
   test('getMeeting() returns an observable', () => {
@@ -41,6 +42,13 @@ describe('Meetings JSON Adapter', () => {
         done();
       }
     );
+  });
+
+  test('getMeeting() renders the local media if localVideo property is defined', (done) => {
+    meetingsJSONAdapter.getMeeting('localVideo').subscribe((data) => {
+      expect(data.localVideo).toEqual('mock-stream');
+      done();
+    });
   });
 
   afterEach(() => {
--- a/src/components/WebexLocalMedia/WebexLocalMedia.js
+++ b/src/components/WebexLocalMedia/WebexLocalMedia.js
@@ -1,0 +1,29 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+import {WebexAvatar} from '../';
+import {useMeeting, useVideo} from '../hooks';
+
+import './WebexLocalMedia.scss';
+
+/**
+ * Webex Local Media component displays the user's local video
+ *
+ * @param {object} props
+ * @returns {object} JSX of the component
+ */
+export default function WebexLocalMedia({meetingID, personID}) {
+  const {localVideo} = useMeeting(meetingID);
+  const videoRef = useVideo(localVideo);
+
+  return (
+    <div className="local-media">
+      {localVideo ? <video ref={videoRef} playsInline /> : <WebexAvatar personID={personID} />}
+    </div>
+  );
+}
+
+WebexLocalMedia.propTypes = {
+  meetingID: PropTypes.string.isRequired,
+  personID: PropTypes.string.isRequired,
+};
--- a/src/components/WebexLocalMedia/WebexLocalMedia.scss
+++ b/src/components/WebexLocalMedia/WebexLocalMedia.scss
@@ -1,0 +1,17 @@
+.local-media {
+  display: flex;
+  justify-content: center;
+  align-items: center;
+  height: 100%;
+  width: 100%;
+}
+
+.local-media video {
+  width: 100%;
+  height: 100%;
+}
+
+.local-media .md-avatar {
+  width: 6.25rem !important;
+  height: 6.25rem !important;
+}
--- a/src/components/WebexLocalMedia/WebexLocalMedia.stories.js
+++ b/src/components/WebexLocalMedia/WebexLocalMedia.stories.js
@@ -1,0 +1,27 @@
+import React from 'react';
+import {storiesOf} from '@storybook/react';
+
+import jsonData from '../../data';
+import {WebexJSONAdapter} from '../../adapters';
+import {WebexLocalMedia, WebexDataProvider} from '../';
+
+// Setup for the stories
+const stories = storiesOf('Webex Local Media', module);
+const webexAdapter = new WebexJSONAdapter(jsonData);
+const wrapperStyle = {height: '500px', width: '800px', border: '1px solid black'};
+
+stories.add('enabled', () => (
+  <div style={wrapperStyle}>
+    <WebexDataProvider adapter={webexAdapter}>
+      <WebexLocalMedia meetingID="localVideo" personID="default" />
+    </WebexDataProvider>
+  </div>
+));
+
+stories.add('disabled', () => (
+  <div style={wrapperStyle}>
+    <WebexDataProvider adapter={webexAdapter}>
+      <WebexLocalMedia meetingID="scheduledMeeting" personID="default" />
+    </WebexDataProvider>
+  </div>
+));
--- a/src/components/WebexLocalMedia/WebexLocalMedia.test.js
+++ b/src/components/WebexLocalMedia/WebexLocalMedia.test.js
@@ -1,0 +1,18 @@
+import React from 'react';
+
+import WebexLocalMedia from './WebexLocalMedia';
+
+jest.mock('../hooks/useMeeting');
+jest.mock('../hooks/useVideo');
+
+describe('Webex Local Media component', () => {
+  describe('snapshot', () => {
+    test('matches snapshot of local video', () => {
+      expect(shallow(<WebexLocalMedia meetingID="localVideo" personID="default" />)).toMatchSnapshot();
+    });
+
+    test('matches snapshot of disabled local video', () => {
+      expect(shallow(<WebexLocalMedia meetingID="scheduledMeeting" personID="default" />)).toMatchSnapshot();
+    });
+  });
+});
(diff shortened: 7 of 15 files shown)


src/data/meetings.json
```diff
@@ -49,5 +49,31 @@
...
     "localShare": {},
     "remoteShare": {}
+  },
+  "localVideo": {
+    "ID": "localVideo",
+    "title": "",
+    "startTime": "",
+    "endTime": "",
+    "localVideo": {
+      "src": "./sample.mp4"
```
lalli-flores: Can we name this something more specific? like `localMedia.mp4`?

src/data/meetings.json
```diff
@@ -49,5 +49,31 @@
...
+    },
+    "remoteVideo": {},
+    "localAudio": {},
+    "remoteAudio": {},
+    "localShare": {},
+    "remoteShare": {}
+  },
+  "noLocalVideo": {
```
lalli-flores: Is this necessary? Can you just use one of the existing meeting objects?

src/data/people.json
```diff
@@ -109,6 +109,16 @@
     "orgID": "Y2lzY29zcGFyazovL3VzL09SR0FOSVpBVElPTi9jb25zdW1lcg",
     "status": "self"
   },
+  "noStatus": {
```
lalli-flores: `default` object is supposed to represent the "no status" state. I don't think this object is necessary

src/data/index.js
```diff
@@ -3,5 +3,7 @@ import meetings from './meetings';
 import people from './people';
 import rooms from './rooms';
 
+export {default as SampleMP4Url} from './sample.mp4';
```
lalli-flores: Can you rename this to `localMedia`?

src/components/WebexLocalMedia/WebexLocalMedia.js
```diff
@@ -0,0 +1,33 @@
...
+ * Webex Local Media component displays the user's local video
+ * previous to joining a meeting from a given meetingID.
+ *
+ * @param {object} props
+ * @returns {object} JSX of the component
+ */
+export default function WebexLocalMedia({meetingID, personID}) {
+  const {localVideo} = useMeeting(meetingID);
+  const media =
+    localVideo && Object.entries(localVideo).length > 0 ? (
```
lalli-flores: Is there a reason to check for `localvideo` properties? If there is no local video, wouldn't it be `null`?

src/components/WebexLocalMedia/WebexLocalMedia.js
```diff
@@ -0,0 +1,33 @@
...
+      <video controls>
+        <source src={localVideo.src} />
```
lalli-flores: In the [WebRTC samples](https://github.com/webrtc/samples/blob/gh-pages/src/content/getusermedia/gum/js/main.js#L22), they set [`srcObject`](https://developer.mozilla.org/en-US/docs/Web/API/HTMLMediaElement/srcObject) instead of `src`.

`srcObject` from the browser support perspective, mostly only works with `MediaStream` objects (like the ones the SDK will be sending), and `src` is meant to be used for URLS, files or blobs. We should follow their `Supporting fallback to the src property` sample code.

Screen Shot 2019-11-20 at 11 45 33 AM

src/components/WebexLocalMedia/WebexLocalMedia.js
```diff
@@ -0,0 +1,33 @@
...
+ * previous to joining a meeting from a given meetingID.
+ *
+ * @param {object} props
+ * @returns {object} JSX of the component
+ */
+export default function WebexLocalMedia({meetingID, personID}) {
+  const {localVideo} = useMeeting(meetingID);
+  const media =
+    localVideo && Object.entries(localVideo).length > 0 ? (
+      <video controls>
```
lalli-flores: I don't think we should be including `controls`. Video should be controlled via meeting controls

src/components/WebexLocalMedia/WebexLocalMedia.scss
```diff
@@ -0,0 +1,16 @@
+.container {
+  display: flex;
+  justify-content: center;
+  align-items: center;
+  height: 100%;
```
lalli-flores: Any reason not to include `width`?

src/components/WebexLocalMedia/WebexLocalMedia.test.js
```diff
@@ -0,0 +1,17 @@
...
+
+jest.mock('../hooks/useMeeting');
+
+describe('Webex Local Media component', () => {
+  describe('snapshot', () => {
+    test('matches snapshot of local video', () => {
+      expect(shallow(<WebexLocalMedia meetingID="localVideo" personID="noStatus" />)).toMatchSnapshot();
+    });
+
+    test('matches snapshot of one on one meeting', () => {
```
lalli-flores:
```suggestion
    test('matches snapshot of disabled local video', () => {
```

src/adapters/MeetingsJSONAdapter.js
```diff
@@ -37,6 +38,11 @@ export default class MeetingsJSONAdapter extends MeetingsAdapter {
   getMeeting(ID) {
     return Observable.create((observer) => {
       if (this.datasource[ID]) {
+        // Add the correct local file url to the localVideo object
+        if (this.datasource[ID].localVideo && this.datasource[ID].localVideo.src) {
```
lalli-flores: You can simplify this and follow the "check" pattern we have
```suggestion
        if (localVideo in this.datasource[ID) {
```
There's no need to check for `src` since you will be assigning it (if it doesn't exist, it gets added)

src/components/WebexLocalMedia/WebexLocalMedia.stories.js
```diff
@@ -0,0 +1,27 @@
...
+import jsonData from '../../data';
+import {WebexJSONAdapter} from '../../adapters';
+import {WebexLocalMedia, WebexDataProvider} from '../';
+
+// Setup for the stories
+const stories = storiesOf('Webex Local Media', module);
+const webexAdapter = new WebexJSONAdapter(jsonData);
+const wrapperStyle = {height: '500px', width: '800px', border: '1px solid black'};
+
+stories.add('local video', () => (
```
lalli-flores: We can call this "enabled" since the component name is already Local Media
```suggestion
stories.add('enabled', () => (
```

src/components/WebexLocalMedia/WebexLocalMedia.stories.js
```diff
@@ -0,0 +1,27 @@
...
+  <div style={wrapperStyle}>
+    <WebexDataProvider adapter={webexAdapter}>
+      <WebexLocalMedia meetingID="localVideo" personID="noStatus" />
+    </WebexDataProvider>
+  </div>
+));
+
+stories.add('no Local video', () => (
```
lalli-flores:
```suggestion
stories.add('disabled', () => (
```

To match the previous name suggestion

src/components/WebexLocalMedia/WebexLocalMedia.js
akoushke: this is for the initial render when the object returned from `useMeeting(meetingID)` is `undefined` and `localVideo` is a nested property. Therefore we need to check if `localVideo` or the parent object is defined.

src/components/WebexLocalMedia/WebexLocalMedia.js
akoushke: I added that to be able to control the video. I'm not sure how else to play the video, unless we don't need to.

src/components/WebexLocalMedia/WebexLocalMedia.js
akoushke: you want to embed the object this way instead of using `JSX` and attributes? the object can NOT be passed as an attribute.

src/adapters/MeetingsJSONAdapter.js
akoushke: we don't need `in`. I'm doing something simpler

src/components/WebexLocalMedia/WebexLocalMedia.js
lalli-flores: You're not checking for parent here though? You already "grabbed" localVideo in the previous line, so it's either `null` or `undefined` at this point.
Wouldn't `localVideo ? ...` be enough?

src/components/WebexLocalMedia/WebexLocalMedia.js
lalli-flores: Right, that is how is handled in all examples of media I've seen

src/components/WebexLocalMedia/WebexLocalMedia.js
akoushke: no. we are. that's how destructing works. we cant destruct from undefined.

src/components/WebexLocalMedia/WebexLocalMedia.js
akoushke: we need to discuss about this then.

akoushke: this is why we are using `source` DOM

src/components/WebexLocalMedia/WebexLocalMedia.js
lalli-flores: You can add "autoplay" but not sure how the controls would work with a `MediaStream`. We certainly don't want "browser controls" in a meeting experience

src/adapters/MeetingsJSONAdapter.js
```diff
@@ -25,6 +25,28 @@ import {Observable} from 'rxjs';
...
+      video: true,
+      audio: false,
+    };
+    let stream;
+
+    try {
+      stream = await navigator.mediaDevices.getUserMedia(constraints);
+    } catch (reason) {
+      // eslint-disable-next-line no-console
+      console.error('Meeting JSON adapter is not able to display user local stream', reason);
```
adamweeks:
```suggestion
      console.error('Meetings JSON adapter is not able to display user local stream', reason);
```

src/adapters/MeetingsJSONAdapter.js
```diff
@@ -44,5 +66,15 @@ export default class MeetingsJSONAdapter extends MeetingsAdapter {
 
       observer.complete();
     });
+
+    const getMeetingWithLocalVideo$ = getMeeting$.pipe(
+      // Attach the localMedia stream if meeting.localVideo in not `null`.
```
adamweeks:
```suggestion
      // Attach the localMedia stream if meeting.localVideo is not `null`.
```

src/adapters/MeetingsJSONAdapter.test.js
```diff
@@ -43,6 +43,17 @@ describe('Meetings JSON Adapter', () => {
     );
   });
 
+  test('getMeeting() renders the local media if localVideo property is defined', (done) => {
+    window.navigator.mediaDevices = {
+      getUserMedia: () => jest.fn(Promise.resolve()),
```
adamweeks: Would like to see this resolve something, and then checked in the expectation:

```suggestion
      getUserMedia: () => jest.fn(Promise.resolve('mock-stream')),
```

src/adapters/MeetingsJSONAdapter.test.js
```diff
@@ -43,6 +43,17 @@ describe('Meetings JSON Adapter', () => {
...
+    };
+    meetingID = 'localVideo';
+    meetingsJSONAdapter.getMeeting(meetingID).subscribe((data) => {
+      expect(data).toEqual(meetings[meetingID]);
```
adamweeks: If we resolve a mock string as mentioned above, we could test it here:

```suggestion
      expect(data.localVideo).toEqual('mock-stream');
```

akoushke: yeah. that's cool

src/components/WebexLocalMedia/WebexLocalMedia.js
```diff
@@ -0,0 +1,32 @@
...
+ *
+ * @param {object} props
+ * @returns {object} JSX of the component
+ */
+export default function WebexLocalMedia({meetingID, personID}) {
+  const {localVideo} = useMeeting(meetingID);
+  const videoRef = useVideo(localVideo);
+  const localMedia = (
+    <Fragment>
+      <audio playsInline />
```
adamweeks: Local media shouldn't have audio. This will only be needed on the remote stream.

akoushke: ahhh, I was a bit confused about this too.

src/components/hooks/useVideo.js
```diff
@@ -0,0 +1,26 @@
+import {useEffect, useRef} from 'react';
+
+/**
+ * Custom hook that returns activity data of the given ID.
```
adamweeks: Copy paste fail

src/components/hooks/useVideo.js
```diff
@@ -0,0 +1,26 @@
...
+ *
+ * @param {MediaStream} stream  ID of the activity for which to return data.
```
adamweeks: another copypasta fail 😃

adamweeks:
Really elegant solution! Just a few minor changes requested.

img

src/adapters/MeetingsJSONAdapter.js
```diff
@@ -25,6 +25,28 @@ import {Observable} from 'rxjs';
  * Implements the MeetingsAdapter interface with a JSON object as its datasource. See @MeetingsJSON
  */
 export default class MeetingsJSONAdapter extends MeetingsAdapter {
+  /**
+   * Returns a MediaStream object that contains local stream of a user.
+   *
+   * @returns {MediaStream}
```
lalli-flores:
```suggestion
   * @returns {MediaStream}
   * @private
```

src/adapters/MeetingsJSONAdapter.js
```diff
@@ -25,6 +25,28 @@ import {Observable} from 'rxjs';
  * Implements the MeetingsAdapter interface with a JSON object as its datasource. See @MeetingsJSON
  */
 export default class MeetingsJSONAdapter extends MeetingsAdapter {
+  /**
+   * Returns a MediaStream object that contains local stream of a user.
```
lalli-flores:
```suggestion
   * Returns a MediaStream object obtained from local user media.
```

src/components/hooks/useVideo.js
```diff
@@ -0,0 +1,26 @@
...
+ * @returns {Element} Video reference element
```
lalli-flores:
```suggestion
 * @returns {Element} Video element reference
```

src/adapters/MeetingsJSONAdapter.js
akoushke: there will be a `techDebt` ticket.

package.json
```diff
@@ -40,6 +40,8 @@
   "devDependencies": {
     "@babel/core": "^7.5.5",
     "@babel/plugin-proposal-export-default-from": "^7.5.2",
+    "@babel/plugin-transform-regenerator": "^7.7.0",
```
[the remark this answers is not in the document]
akoushke: we didn't have any async calls before.

src/components/WebexLocalMedia/WebexLocalMedia.js
```diff
@@ -0,0 +1,31 @@
...
+ * previous to joining a meeting from a given meetingID.
+ *
+ * @param {object} props
+ * @returns {object} JSX of the component
+ */
+export default function WebexLocalMedia({meetingID, personID}) {
+  const {localVideo} = useMeeting(meetingID);
+  const videoRef = useVideo(localVideo);
+  const localMedia = (
+    <Fragment>
```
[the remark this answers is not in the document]
akoushke: not anymore.

src/adapters/MeetingsJSONAdapter.test.js
```diff
@@ -43,6 +44,15 @@ describe('Meetings JSON Adapter', () => {
     );
   });
 
+  test('getMeeting() renders the local media if localVideo property is defined', (done) => {
+    meetingID = 'localVideo';
+
+    meetingsJSONAdapter.getMeeting(meetingID).subscribe(async (data) => {
+      expect(await data.localVideo).toEqual('mock-stream');
```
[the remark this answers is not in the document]
akoushke: forgot to remove.

src/adapters/MeetingsJSONAdapter.js
```diff
@@ -21,10 +21,31 @@ import {Observable} from 'rxjs';
...
+      video: true,
+      audio: false,
+    };
+    let stream;
+
+    try {
+      stream = await navigator.mediaDevices.getUserMedia(constraints);
+    } catch (reason) {
+      // eslint-disable-next-line no-console
+      console.error('Meetings JSON adapter can not displaying the user local stream', reason);
```
lalli-flores:
```suggestion
      console.error('Meetings JSON adapter can not display the user local stream', reason);
```

src/components/WebexLocalMedia/WebexLocalMedia.js
```diff
@@ -0,0 +1,30 @@
...
+import PropTypes from 'prop-types';
+
+import {WebexAvatar} from '../';
+import {useMeeting, useVideo} from '../hooks';
+
+import './WebexLocalMedia.scss';
+
+/**
+ * Webex Local Media component displays the user's local video
+ * previous to joining a meeting from a given meetingID.
```
lalli-flores: I think you should remove this part of the comment. Local media will be used in-meeting as well, for meeting "self-view"

PR #55, opened 2019-11-25
Jest eslint hooks

adamweeks:


adamweeks:
Updated with @lalli-flores PR changes.

PR #60, opened 2019-12-13
Local Media component uses "getMe" observable instead of prop

lalli-flores:


src/adapters/PeopleJSONAdapter.js
```diff
@@ -22,14 +22,27 @@ import {Observable} from 'rxjs';
...
+   * Returns an observable that emits person data of the default user.
+   *
+   * @returns {Observable.<Person>}
+   * @memberof PeopleJSONAdapter
+   */
+  getMe() {
+    const defaultUserID = 'default';
 
-    this.getPerson = this.getPerson.bind(this);
+    return Observable.create((observer) => {
```
akoushke: no `observer.error`?

lalli-flores: No, the current user should always exist

PR #62, opened 2019-12-17
fix(WebexDataProvider): use proptype of shape

adamweeks:


adamweeks:
> the commit message could be more descriptive: `replace exact with shape` or something like that. not a blocker though

Updated the commit message

PR #63, opened 2019-12-20
Interstitial Component

akoushke:


src/components/hooks/useMeeting.js
```diff
@@ -1,28 +1,50 @@
...
 import {AdapterContext} from '../../components/';
 
 /**
  * Custom hook that returns meeting data of the given ID.
+ * If no ID is given, a `meetingDestination` must be provided to
+ * create a new meeting.
  *
- * @param {string} meetingID  ID of the meeting for which to return data.
+ * @param {string} meetingID           ID of the meeting for which to return data.
+ * @param {string} meetingDestination  Virtual location where the meeting should take place
```
adamweeks:
```suggestion
 * @param {string} meetingDestination  Virtual location where the meeting should take place.
```

src/components/hooks/useMeeting.js
```diff
@@ -1,28 +1,50 @@
...
-    const subscription = meetingsAdapter.getMeeting(meetingID).subscribe(setMeeting, onError);
+
+    if (meetingID) {
+      subscription = meetingsAdapter.getMeeting(meetingID).subscribe(setMeeting, onError);
+    } else if (meetingDestination) {
+      // Create a meeting, start event listeners by subscribing to getMeeting,
+      // wait for local media to get added and complete
+      subscription = meetingsAdapter
+        .createMeeting(meetingDestination)
+        .pipe(
```
adamweeks: It would be nice if the `createMeeting` subscription could handle all this logic for us instead of needing it in the hook.

src/adapters/MeetingsJSONAdapter.js
```diff
@@ -226,7 +226,7 @@ export default class MeetingsJSONAdapter extends MeetingsAdapter {
       const meeting = this.datasource[ID];
 
       if (meeting) {
-        observer.next(meeting.localAudio.muted ? muted : unmuted);
+        observer.next(meeting.localAudio ? muted : unmuted);
```
adamweeks: I like this concept of checking the object instead of diving into the stream properties. 👍

src/components/hooks/useMeeting.js
```diff
@@ -1,28 +1,50 @@
...
+          flatMap(({ID}) => meetingsAdapter.getMeeting(ID)),
+          take(1),
+          delayWhen(({ID}) => from(meetingsAdapter.addLocalMedia(ID)))
```
adamweeks: Where does `addLocalMedia` get called if the `meetingID` exists?

src/components/hooks/useMeeting.js
akoushke: that could be another action item for enhancement maybe?

src/components/hooks/useMeeting.js
akoushke: `addLocalMedia` should not get invoked at that point at least for our MVP logic. the only that we have the `meetingID` and we need to invoke `addLocalMedia` is when there is a scheduled meeting.

src/components/WebexLocalMedia/WebexLocalMedia.js
```diff
@@ -16,12 +17,9 @@ export default function WebexLocalMedia({meetingID}) {
   const {localVideo} = useMeeting(meetingID);
   const {ID} = useMe();
   const videoRef = useStream(localVideo);
+  const disabledVideo = ID ? <WebexAvatar personID={ID} /> : <Spinner />;
```
lalli-flores: Can you add a snapshot test for the loading state?

src/components/hooks/useMeeting.js
lalli-flores: Agreed, let's do it in a separate PR. However, not sure that I like the idea of `createMeeting` code calling `getMeeting` behind the scenes

akoushke: we could either wrap all methods into one or figure out if there are any other cases when we need to invoke `createMeeting` by itself? if not, then we can move all the logic inside `createMeeting`

src/components/WebexLocalMedia/WebexLocalMedia.js
akoushke: not sure how to mock `useMe()` to return undefined.

akoushke:
Addressed comments.

src/adapters/MeetingsJSONAdapter.test.js
```diff
@@ -81,6 +104,29 @@ describe('Meetings JSON Adapter', () => {
...
+    });
+
+    test('attaches local audio stream if targeted meeting object requires one', async () => {
+      meetingsJSONAdapter.getStream = jest.fn();
+      await meetingsJSONAdapter.addLocalMedia('localAudio');
+
+      expect(meetingsJSONAdapter.getStream).toHaveBeenCalledWith({audio: true, video: false});
+    });
+
+    test('does nothing if neither are local audio or video are required', async () => {
```
adamweeks:
```suggestion
    test('does nothing if neither local audio or video are required', async () => {
```

PR #71, opened 2020-01-24
feat(UsePerson): convert PersonStatus keys to values

akoushke:


akoushke:
> Looks good, but got me thinking what should we do in cases we don't want the status to show? Like in `LocalMedia` component?

that should be handled inside localMedia logic

PR #72, opened 2020-01-24
feat(UseStream): play video is an async call

akoushke:


src/components/hooks/useStream.js
```diff
@@ -10,12 +10,21 @@ export default function useStream(stream) {
...
-      mediaElement.srcObject = stream;
-      mediaElement.play();
-    }
+        if (stream instanceof MediaStream) {
+          mediaElement.srcObject = stream;
+          await mediaElement.play();
+        }
+      } catch (error) {
+        // eslint-disable-next-line no-console
+        console.error('new stream is being added. not sure if we should error this out');
```
adamweeks: Perhaps a more informative error message

```suggestion
        console.error('Unable to play stream: ', error.message);
```

akoushke: method is removed

PR #73, opened 2020-01-25
RXJS peer dependency

akoushke:
Moving RxJS as a peer dependency instead of bundling it with the component code. An issue of incompatible RxJS libraries was found when trying to use the component bundle with `create-react-app` + RxJS.

### DIFF
--- a/package.json
+++ b/package.json
@@ -28,14 +28,14 @@
   "license": "MIT",
   "peerDependencies": {
     "react": "^16.9.0",
-    "react-dom": "^16.9.0"
+    "react-dom": "^16.9.0",
+    "rxjs": "^6.5.4"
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -74,7 +74,7 @@
     "jest": "^24.8.0",
     "jest-junit": "^7.0.0",
     "mockdate": "^2.0.5",
-    "node-sass": "^4.12.0",
+    "node-sass": "^4.13.1",
```
lalli-flores: Don't update `node-sass`. If you look at [their support chart](https://github.com/sass/node-sass#node-sass), the minor version (13), means the node version `node-sass` is compatible with.

PR #78, opened 2020-02-03
 json adapters should not be part of the bundle

akoushke:


akoushke:
> I don't necessarily agree with this changes. I think it's ok for the JSON adapters to be part of the bundle. This way developers have a way of testing different adapters, and quickly get started when trying to familiarize themselves with the components.

but there is no data

adamweeks:
Since the developers that want to test out json adapters will have to pull data in from the src folder, it doesn't really help having them exported. I'm good with removing them.

PR #79, opened 2020-02-15
Momentum Icon compatibility fix

lalli-flores:
Commit type for Momentum UI is of type `fix` because this is a fixing change that needs to be released


src/components/WebexMeetingControl/__snapshots__/WebexMeetingControl.test.js.snap
```diff
@@ -1,37 +1,22 @@
 // Jest Snapshot v1, https://goo.gl/fbAQLP
 
 exports[`Webex Meeting Control component snapshot matches with active icon button 1`] = `
-<Button
-  active={false}
+<ContextTransform(Button)
```
adamweeks: Is this a momentum change that affected our snapshot?

lalli-flores: Yes, I was trying to figure out where this changes happened but couldn't really get a good idea. Seems like there have been changes within the codebase to use React contexts

PR #81, opened 2020-02-19
Local media Style Fixes

lalli-flores:
Includes mobile-first approach.


akoushke:
so we are doing sass now?

src/styles/index.scss
```diff
@@ -1,2 +1,3 @@
+@import '~include-media/dist/_include-media';
```
adamweeks: Could you add a comment as to what this package does?

akoushke:
for some reasons, I'm not seeing the the size changes on an avatar, after joining a meeting with `webex meeting` I was under assumption that the size of the avatar should change if the size of the container had changed.

lalli-flores:
> for some reasons, I'm not seeing the the size changes on an avatar, after joining a meeting with `webex meeting` I was under assumption that the size of the avatar should change if the size of the container had changed.

The reason is because the change is based on media queries. When using local media within in-meeting component, the window is still large (only the component is small in size), so the media query styles doesn't apply.
I changed implementation to use ResizeObserver API (https://github.com/webex/components/pull/83), which should fix that issue

PR #83, opened 2020-02-25
Webex Local Media styles based on container width

lalli-flores:
This a version that relies on the size on the container based on the [Resize Observable API](https://developer.mozilla.org/en-US/docs/Web/API/ResizeObserver).

Replaces https://github.com/webex/components/pull/81, which is based on media queries (which look at the window, not the container).

### DIFF
--- a/package.json
+++ b/package.json
@@ -83,7 +83,7 @@
     "rollup-plugin-node-resolve": "^5.2.0",
     "rollup-plugin-scss": "^1.0.2",
     "rxjs": "^6.5.4",
-    "sass-loader": "^7.1.0",
+    "sass-loader": "^8.0.2",
     "sass-resources-loader": "^2.0.1",
     "semantic-release": "^15.13.19",
     "style-loader": "^0.23.1"
--- a/package.json
+++ b/package.json
@@ -32,6 +32,7 @@
     "rxjs": "^6.5.4"
   },
   "dependencies": {
+    "@juggle/resize-observer": "^3.0.2",
     "@momentum-ui/react": "^23.2.1",
     "@webex/component-adapter-interfaces": "^1.9.0",
     "date-fns": "^2.6.0",
--- a/package.json
+++ b/package.json
@@ -55,6 +55,7 @@
     "babel-jest": "^24.8.0",
     "babel-loader": "^8.0.6",
     "babel-plugin-transform-object-rest-spread": "^6.26.0",
+    "classnames": "^2.2.6",
     "conventional-changelog-angular": "^5.0.3",
     "css-loader": "^3.1.0",
     "enzyme": "^3.10.0",
--- a/src/components/WebexLocalMedia/WebexLocalMedia.js
+++ b/src/components/WebexLocalMedia/WebexLocalMedia.js
@@ -4,7 +4,7 @@ import classNames from 'classnames';
 import {Spinner} from '@momentum-ui/react';
 
 import {WebexAvatar} from '../';
-import {useMe, useMeeting, useStream} from '../hooks';
+import {useElementDimensions, useMe, useMeeting, useStream} from '../hooks';
 
 import './WebexLocalMedia.scss';
 
@@ -15,18 +15,24 @@ import './WebexLocalMedia.scss';
  * @returns {Object} JSX of the component
  */
 export default function WebexLocalMedia({meetingID}) {
+  const [mediaRef, {width}] = useElementDimensions();
   const {localVideo} = useMeeting(meetingID);
   const {ID} = useMe();
   const videoRef = useStream(localVideo);
 
   const cssClasses = classNames({
     'local-media': true,
+    'local-media-desktop': width >= 425, // Standard large phone width
     'no-media': localVideo === null,
   });
 
   const disabledVideo = ID ? <WebexAvatar personID={ID} displayStatus={false} /> : <Spinner />;
 
-  return <div className={cssClasses}>{localVideo ? <video ref={videoRef} playsInline autoPlay /> : disabledVideo}</div>;
+  return (
+    <div ref={mediaRef} className={cssClasses}>
+      {localVideo ? <video ref={videoRef} playsInline autoPlay /> : disabledVideo}
+    </div>
+  );
 }
 
 WebexLocalMedia.propTypes = {
--- a/src/components/WebexLocalMedia/WebexLocalMedia.js
+++ b/src/components/WebexLocalMedia/WebexLocalMedia.js
@@ -1,5 +1,6 @@
 import React from 'react';
 import PropTypes from 'prop-types';
+import classNames from 'classnames';
 import {Spinner} from '@momentum-ui/react';
 
 import {WebexAvatar} from '../';
@@ -17,11 +18,15 @@ export default function WebexLocalMedia({meetingID}) {
   const {localVideo} = useMeeting(meetingID);
   const {ID} = useMe();
   const videoRef = useStream(localVideo);
+
+  const cssClasses = classNames({
+    'local-media': true,
+    'no-media': localVideo === null,
+  });
+
   const disabledVideo = ID ? <WebexAvatar personID={ID} displayStatus={false} /> : <Spinner />;
 
-  return (
-    <div className="local-media">{localVideo ? <video ref={videoRef} playsInline autoPlay /> : disabledVideo}</div>
-  );
+  return <div className={cssClasses}>{localVideo ? <video ref={videoRef} playsInline autoPlay /> : disabledVideo}</div>;
 }
 
 WebexLocalMedia.propTypes = {
--- a/src/components/WebexLocalMedia/WebexLocalMedia.js
+++ b/src/components/WebexLocalMedia/WebexLocalMedia.js
@@ -8,10 +8,10 @@ import {useMe, useMeeting, useStream} from '../hooks';
 import './WebexLocalMedia.scss';
 
 /**
- * Webex Local Media component displays the user's local video
+ * Webex Local Media component displays the user's local video.
  *
- * @param {object} props
- * @returns {object} JSX of the component
+ * @param {string} props.meetingID  ID of the meeting from which to obtain local media
+ * @returns {Object} JSX of the component
  */
 export default function WebexLocalMedia({meetingID}) {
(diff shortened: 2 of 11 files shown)


package.json
```diff
@@ -83,7 +84,7 @@
     "rollup-plugin-node-resolve": "^5.2.0",
     "rollup-plugin-scss": "^1.0.2",
     "rxjs": "^6.5.4",
-    "sass-loader": "^7.1.0",
+    "sass-loader": "^8.0.2",
```
akoushke: upgrade that sass!

src/components/WebexLocalMedia/WebexLocalMedia.stories.js
```diff
@@ -5,23 +5,17 @@ import jsonData from '../../data';
...
 const webexAdapter = new WebexJSONAdapter(jsonData);
-const wrapperStyle = {height: '500px', width: '800px', border: '1px solid black'};
 
 stories.add('enabled', () => (
-  <div style={wrapperStyle}>
-    <WebexDataProvider adapter={webexAdapter}>
-      <WebexLocalMedia meetingID="localVideo" personID="default" />
-    </WebexDataProvider>
-  </div>
+  <WebexDataProvider adapter={webexAdapter}>
```
akoushke: good catch!

src/components/hooks/useElementDimensions.js
```diff
@@ -0,0 +1,42 @@
...
+ *
+ * @returns {Array} Element reference and dimensions in pixels
+ */
+export default function useElementDimensions() {
+  const elementRef = useRef({});
+  const [height, setHeight] = useState(0);
+  const [width, setWidth] = useState(0);
+
+  useEffect(() => {
+    const ResizeObserver = window.ResizeObserver || Polyfill;
```
akoushke: why this or that? can you explain a bit what this means? brief one line comment perhaps?

src/components/hooks/useElementDimensions.js
```diff
@@ -0,0 +1,42 @@
...
+    const observer = new ResizeObserver((entries) => {
+      // Get the first DOM element (our ref) and look at the border box dimensions
+      const [entry] = entries;
+      const {contentRect} = entry;
+
+      // Update height only if there has been height change
+      if (height !== contentRect.height) {
```
akoushke: this is cool!

src/components/WebexLocalMedia/WebexLocalMedia.js
```diff
@@ -3,24 +3,29 @@ import PropTypes from 'prop-types';
...
- * @returns {object} JSX of the component
+ * @param {string} props.meetingID  ID of the meeting from which to obtain local media
+ * @returns {Object} JSX of the component
  */
 export default function WebexLocalMedia({meetingID}) {
+  const [mediaRef, {width}] = useElementDimensions();
   const {localVideo} = useMeeting(meetingID);
   const {ID} = useMe();
   const videoRef = useStream(localVideo);
+  const noMediaClass = localVideo ? '' : ' no-media';
```
akoushke: I think you need to trim that ` no-media`

lalli-flores: it's on purpose, so there is a space between class names if it gets added

src/components/WebexLocalMedia/WebexLocalMedia.scss
```diff
@@ -2,16 +2,30 @@
...
+
+  &-large {
+    .md-avatar {
+      transform: scale(3); // Avatar is 40x40px, this makes it 120x120px
+    }
+  }
 }
+
+.no-media {
+  background: white;
```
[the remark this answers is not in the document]
lalli-flores: The default background is `black`, for when there is video. When video is turned off, we want the background to be `white`

src/components/hooks/useElementDimensions.js
lalli-flores: Sure, I can add a comment. This checks to see if the browser has a native implementation of `ResizeObserver`, and uses that instead of the polyfill

akoushke:
two things:

 1 - I noticed that the avatar size does not change dynamically as we change the size of the browser. I know this applies to the parent container, but shouldn't that localMedia classname change if the parent container size had changed?
Screen Shot 2020-02-25 at 10 40 15 AM
2 - also Firefox shows the border :/ 
Screen Shot 2020-02-25 at 10 33 33 AM

src/components/WebexLocalMedia/WebexLocalMedia.js
```diff
@@ -3,24 +3,29 @@ import PropTypes from 'prop-types';
...
+  const sizeClass = width >= 425 ? ' local-media-large' : ''; // Standard large phone size
   const disabledVideo = ID ? <WebexAvatar personID={ID} displayStatus={false} /> : <Spinner />;
 
   return (
-    <div className="local-media">{localVideo ? <video ref={videoRef} playsInline autoPlay /> : disabledVideo}</div>
+    <div ref={mediaRef} className={`local-media${sizeClass}${noMediaClass}`}>
```
adamweeks: Not needed, but if you want to reduce the complexity of this, you can use the tool we use in the widgets called "classnames" and allows you to throw classnames into an array like:

```js
const localMediaClasses = ['local-media'];
if (width >= 425) localMediaClasses.push(' local-media-large');
if (!localVideo) localMediaClasses.push('no-media');
<div className={classNames(localMediaClasses)}>
```

https://www.npmjs.com/package/classnames

src/components/WebexLocalMedia/WebexLocalMedia.scss
```diff
@@ -2,16 +2,30 @@
   display: flex;
   justify-content: center;
   align-items: center;
+
   height: 100%;
+  min-height: 7.5rem; // Matches avatar's tallest height since transforms are just visual
```
adamweeks: The min height matches the "tallest"? height of the avatar? Shouldn't it match the smallest and then transform?

src/components/hooks/useElementDimensions.js
```diff
@@ -0,0 +1,43 @@
...
+        setHeight(contentRect.height);
+      }
+
+      // Update weight only if there has been weight change
```
adamweeks:
```suggestion
      // Update width only if there has been width change
```

src/components/WebexLocalMedia/WebexLocalMedia.scss
lalli-flores: So what this means is that the smallest you can shrink local media (height-wise) is as tall as the biggest size an avatar inside _can_ get.
do you have suggestions on how to rephrase that comment?

src/components/WebexLocalMedia/WebexLocalMedia.js
lalli-flores: Yeah, I'm aware of it. I remember when trying to dive into the widget code the first few times, it actually made it harder for me to figure out what the code was doing. I had to spend extra time figuring out and googling what this `classNames` function-looking thing was supposed to be doing (was that React syntax? why isn't that in the docs? is that a third-party?), and it made cognitive harder for to to me to keep track of the CSS class names.
I would personally avoid it

lalli-flores:
> two things:
> 
> 1 - I noticed that the avatar size does not change dynamically as we change the size of the browser. I know this applies to the parent container, but shouldn't that localMedia classname change if the parent container size had changed?
> Screen Shot 2020-02-25 at 10 40 15 AM
> 2 - also Firefox shows the border :/
> Screen Shot 2020-02-25 at 10 33 33 AM

1. That's right, the avatar size should change according to the changes of *_its_* parent container (`WebexLocalMedia`), not the window or any other of its ancestors (e.g. `WebexInMeeting`)

2. Where is that border?

src/components/WebexLocalMedia/WebexLocalMedia.js
adamweeks: Not to be argumentative, but knowing that all the class names in an array are going to be spread properly seems like MUCH less cognitive overhead than having to know which variables need spaces before them or nulled out when not needed. Like I said above though, not needed and not blocking the PR.

src/components/WebexLocalMedia/WebexLocalMedia.scss
adamweeks: > So what this means is that the smallest you can shrink local media (height-wise) is as tall as the biggest size an avatar inside _can_ get.
> do you have suggestions on how to rephrase that comment?

`the smallest you can shrink local media (height-wise) is as tall as the biggest size an avatar inside can get.` 😆

adamweeks: Do we need a `min-width` as well?

lalli-flores: > > So what this means is that the smallest you can shrink local media (height-wise) is as tall as the biggest size an avatar inside _can_ get.
> > do you have suggestions on how to rephrase that comment?
> 
> `the smallest you can shrink local media (height-wise) is as tall as the biggest size an avatar inside can get.` 😆

so wordy! lol

lalli-flores: > Do we need a `min-width` as well?

I didn't set one because the avatar sizing is based on the width, but I could set it to the same thing as the height

src/components/WebexLocalMedia/WebexLocalMedia.js
lalli-flores: I used the utility after all. It was already part of our dev dependencies 😛 and I found their object syntax much more readable

akoushke:
> > two things:
> > 1 - I noticed that the avatar size does not change dynamically as we change the size of the browser. I know this applies to the parent container, but shouldn't that localMedia classname change if the parent container size had changed?
> > Screen Shot 2020-02-25 at 10 40 15 AM
> > 2 - also Firefox shows the border :/
> > Screen Shot 2020-02-25 at 10 33 33 AM
> 
> 1. That's right, the avatar size should change according to the changes of __its__ parent container (`WebexLocalMedia`), not the window or any other of its ancestors (e.g. `WebexInMeeting`). In this case, the size of `WebexLocalMedia` is hardcoded, so it would never change.
> 2. Where is that border?

avatar border

PR #85, opened 2020-02-26
Mobile-first styling for Webex In-Meeting

lalli-flores:


src/components/WebexInMeeting/WebexInMeeting.js
```diff
@@ -13,8 +16,15 @@ import './WebexInMeeting.scss';
  * @returns {Object} JSX of the component
  */
 export default function WebexInMeeting({meetingID}) {
+  const [meetingRef, {width}] = useElementDimensions();
+  const cssClasses = classNames({
```
adamweeks: img

src/components/WebexInMeeting/WebexInMeeting.js
```diff
@@ -1,7 +1,10 @@
 import React from 'react';
 import PropTypes from 'prop-types';
+import classNames from 'classnames';
 
-import {WebexLocalMedia, WebexRemoteMedia} from '..';
+import {WebexLocalMedia, WebexRemoteMedia} from '../';
```
adamweeks: I've never seen this module resolution style before, usually either `from '..'` or `from '../index'`. What was the reason for this change?

lalli-flores: That's how we're importing it in [Webex Local Media](https://github.com/webex/components/blob/master/src/components/WebexLocalMedia/WebexLocalMedia.js#L6).
Also, when the import is `from '..'` the linter thinks of it as its own import group. Importing as `from '../` allows it to be in the same importing group (like it should because they're on the same level).
I don't mind using `from '../index'` as long as we agree to do that everywhere and going forward

src/components/WebexInMeeting/WebexInMeeting.js
lalli-flores: I don't mind this "object syntax", this is readable lol

PR #86, opened 2020-02-27
feat(rollup): remove cjs and umd bundles

akoushke:


akoushke:
> Can you update your commit types so that they don't trigger a release?
> I'd suggest using `build`

sure

akoushke:
> Please update https://github.com/webex/components/blob/master/release.config.js#L35 as part of this PR

done

PR #90, opened 2020-03-03
docs(storybook): generate initial storybook

InteractiveTimmy:
The scope of the changes in this pull request is around generating the initial `./docs/storybook` folder for GitHub pages publishing.


akoushke:
you don't want to release this job?

InteractiveTimmy:
> @InteractiveTimmy, I'm assuming all files where generated from `build-storybook` and committed unmodified?
> 
> The only thing I would suggest is to change `favicon.ico` to the Webex logo instead of the storybook logo. You can get a file from [the widgets](https://github.com/webex/widgets/blob/master/demo/teams.png)

This will be over-written as soon as a change occurs. We will need to write a small script in the future to replace the favicon file. Should i still aim to implement this inside the scope of this JIRA/Pull request?

PR #93, opened 2020-03-17
Class namespace usage

adamweeks:


adamweeks:
> Are there new scss processing requirements for the component? Or will all that be handled at build time before it's published?

@andrewholsted no. We will still be publishing css for consumption.

src/components/WebexActivityStream/TimeRuler.scss
```diff
@@ -1,85 +1,5 @@
-// For Storybook, so that activity stream loads content properly
```
adamweeks: None of these styles had to do with the time ruler, so I moved them to `src/components/WebexActivityStream/WebexActivityStream.scss`

src/components/WebexMeetingInfo/WebexMeetingInfo.scss
```diff
@@ -1,19 +1,21 @@
...
   flex-direction: column;
   text-align: center;
-}
 
-.meeting-info h2 {
-  font-size: 28px;
-  font-family: $brand-font-extra-light;
-  color: $md-gray-100;
-}
+  h2 {
```
InteractiveTimmy: good catch for nesting opportunity.

PR #95, opened 2020-03-19
Container styling encapsulation

adamweeks:
After merging #93 we discovered some issues due to expectation of child component classnames.

This PR limits the styles for each component and creates new "container" elements that the parent component styles for layout of the child component only.

One example of the issue this fixes is the usage of the `WebexLocalMedia` component within the `WebexInMeeting` component. The in meeting component was attempting to position the local media component with the class name "remote-media" that comes from the child component. This causes an interdependency between the components and requires strings to be updated in multiple places.

### DIFF
--- a/src/components/WebexInterstitialMeeting/WebexInterstitialMeeting.scss
+++ b/src/components/WebexInterstitialMeeting/WebexInterstitialMeeting.scss
@@ -8,7 +8,7 @@
   width: 100%;
   height: 100%;
 
-  .meeting-info {
+  .interstitial-meeting-info {
     position: absolute;
     align-self: flex-start;
     width: 100%;
--- a/src/components/WebexLocalMedia/WebexLocalMedia.js
+++ b/src/components/WebexLocalMedia/WebexLocalMedia.js
@@ -16,30 +16,33 @@ import './WebexLocalMedia.scss';
  * @param {string} props.meetingID  ID of the meeting from which to obtain local media
  * @returns {Object} JSX of the component
  */
-export default function WebexLocalMedia({meetingID}) {
+export default function WebexLocalMedia({className, meetingID}) {
   const [mediaRef, {width}] = useElementDimensions();
   const {localVideo} = useMeeting(meetingID);
   const {ID} = useMe();
   const videoRef = useStream(localVideo);
 
-  const classBaseName = 'local-media';
-  const classObjects = {};
-
-  classObjects[`${WEBEX_COMPONENTS_CLASS_PREFIX}-${classBaseName}`] = true;
-  classObjects[`${WEBEX_COMPONENTS_CLASS_PREFIX}-${classBaseName}-desktop`] = width >= PHONE_LARGE;
-  classObjects[`${WEBEX_COMPONENTS_CLASS_PREFIX}-no-media`] = localVideo === null;
-
-  const cssClasses = classNames(classObjects);
-
+  const classBaseName = `${WEBEX_COMPONENTS_CLASS_PREFIX}-local-media`;
+  const mainClasses = {
+    [classBaseName]: true,
+    [`${classBaseName}-desktop`]: width >= PHONE_LARGE,
+    [`${classBaseName}-no-media`]: localVideo === null,
+    [className]: !!className,
+  };
   const disabledVideo = ID ? <WebexAvatar personID={ID} displayStatus={false} /> : <Spinner />;
 
   return (
-    <div ref={mediaRef} className={cssClasses}>
+    <div ref={mediaRef} className={classNames(mainClasses)}>
       {localVideo ? <video ref={videoRef} playsInline autoPlay /> : disabledVideo}
     </div>
   );
 }
 
 WebexLocalMedia.propTypes = {
+  className: PropTypes.string,
   meetingID: PropTypes.string.isRequired,
+};
+
+WebexLocalMedia.defaultProps = {
+  className: '',
 };
--- a/src/components/WebexLocalMedia/WebexLocalMedia.test.js
+++ b/src/components/WebexLocalMedia/WebexLocalMedia.test.js
@@ -15,5 +15,9 @@ describe('Webex Local Media component', () => {
     test('matches snapshot of disabled local video', () => {
       expect(shallow(<WebexLocalMedia meetingID="noMedia" />)).toMatchSnapshot();
     });
+
+    test('matches snapshot of local video with custom CSS class', () => {
+      expect(shallow(<WebexLocalMedia className="my-custom-class" meetingID="noMedia" />)).toMatchSnapshot();
+    });
   });
 });
--- a/src/components/WebexLocalMedia/__snapshots__/WebexLocalMedia.test.js.snap
+++ b/src/components/WebexLocalMedia/__snapshots__/WebexLocalMedia.test.js.snap
@@ -2,7 +2,7 @@
 
 exports[`Webex Local Media component snapshot matches snapshot of disabled local video 1`] = `
 <div
-  className="wxc-local-media wxc-no-media"
+  className="wxc-local-media wxc-local-media-no-media"
 >
   <WebexAvatar
     displayStatus={false}
@@ -18,6 +18,17 @@ exports[`Webex Local Media component snapshot matches snapshot of local video 1`
   <video
     autoPlay={true}
     playsInline={true}
+  />
+</div>
+`;
+
+exports[`Webex Local Media component snapshot matches snapshot of local video with custom CSS class 1`] = `
+<div
+  className="wxc-local-media wxc-local-media-no-media my-custom-class"
+>
+  <WebexAvatar
+    displayStatus={false}
+    personID="default"
   />
 </div>
 `;
--- a/src/components/WebexMeeting/WebexMeeting.js
+++ b/src/components/WebexMeeting/WebexMeeting.js
@@ -21,22 +21,22 @@ export default function WebexMeeting({meetingDestination, controls}) {
   const {ID, remoteVideo} = useMeetingDestination(meetingDestination);
   const isActive = remoteVideo !== null;
 
-  const classBaseName = 'meeting';
-  const classObjects = {};
-
-  classObjects[`${WEBEX_COMPONENTS_CLASS_PREFIX}-${classBaseName}`] = true;
-  classObjects[`${WEBEX_COMPONENTS_CLASS_PREFIX}-${classBaseName}-active`] = ID && isActive;
-
-  const cssClasses = classNames(classObjects);
+  const classBaseName = `${WEBEX_COMPONENTS_CLASS_PREFIX}-meeting`;
+  const mainClasses = {
+    [classBaseName]: true,
+    [`${classBaseName}-active`]: ID && isActive,
+  };
 
   const meetingControls = controls(isActive).map((key) => <WebexMeetingControl key={key} type={key} />);
 
   return (
-    <div className={cssClasses}>
+    <div className={classNames(mainClasses)}>
       {ID ? (
         <Fragment>
           {isActive ? <WebexInMeeting meetingID={ID} /> : <WebexInterstitialMeeting meetingID={ID} />}
-          <WebexMeetingControls meetingID={ID}>{meetingControls}</WebexMeetingControls>
+          <WebexMeetingControls className="meeting-controls-container" meetingID={ID}>
+            {meetingControls}
+          </WebexMeetingControls>
         </Fragment>
       ) : (
         <Spinner />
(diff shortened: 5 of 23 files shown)


adamweeks:
> I understand the problem this is trying to solve, but I don't like the approach.
> I really liked that the components have really semantic markup, without extra elements that were not necessary for markup. Adding DOM elements for the sake of styling is a big no-no for me (as it's as un-semantic as it gets).
> Let me brainstorm ideas other than the way the styles were done

Curious to hear what you come up with. Just to be clear, this solution:

* Does not rely on child classnames in parent components
* Makes positional styling of child components specific to their parents
* Allows child components to rename classnames and structure without breaking integrations

adamweeks:
@lalli-flores one solution we could do is pass a classname as a prop to the child component so the parent can still style the child like:

```
<Child classes={['parentPositionalClass']}>
```

That way we could have `.parentPositionalClass` defined in the `Parent.scss` file.

chrisadubois:
To be clear, this is also blocking a prod deploy for web client until a fix goes in. else we have to revert all the meetings functionality

lalli-flores:
> @lalli-flores one solution we could do is pass a classname as a prop to the child component so the parent can still style the child like:
> 
> ```
> <Child classes={['parentPositionalClass']}>
> ```
> 
> That way we could have `.parentPositionalClass` defined in the `Parent.scss` file.

I like this approach, maybe just name the prop `className` so from usage perspective, it feels like styling any other component. Something [like this answer](https://stackoverflow.com/a/46040808).

The other solution would be to use [React Contexts](https://reactjs.org/docs/context.html#examples). This would also pave the way to allow for themes but it's more work right now.

adamweeks:
> I like this approach, maybe just name the prop `className` so from usage perspective, it feels like styling any other component. Something [like this answer](https://stackoverflow.com/a/46040808).
> 
> The other solution would be to use [React Contexts](https://reactjs.org/docs/context.html#examples). This would also pave the way to allow for themes but it's more work right now.

I will go with `className` since that is what [momentum](https://github.com/momentum-design/momentum-ui/blob/master/react/src/lib/AccordionGroup/index.js#L53) is doing as well.

adamweeks:
@lalli-flores ready for another round

src/components/WebexInterstitialMeeting/WebexInterstitialMeeting.scss
```diff
@@ -8,7 +8,7 @@
   width: 100%;
   height: 100%;
 
-  .meeting-info {
+  .meeting-info-container {
```
lalli-flores:
```suggestion
  .interstitial-meeting-info {
```

src/components/WebexLocalMedia/WebexLocalMedia.test.js
```diff
@@ -15,5 +15,9 @@ describe('Webex Local Media component', () => {
     test('matches snapshot of disabled local video', () => {
       expect(shallow(<WebexLocalMedia meetingID="noMedia" />)).toMatchSnapshot();
     });
+
+    test('matches snapshot of local video with custom class', () => {
```
lalli-flores:
```suggestion
    test('matches snapshot of local video with custom CSS class', () => {
```

src/components/WebexMeetingControl/WebexMeetingControls.test.js
```diff
@@ -13,4 +13,15 @@ describe('Webex Meeting Controls component', () => {
 
     expect(component).toMatchSnapshot();
   });
+
+  test('matches snapshot with custom class', () => {
```
lalli-flores:
```suggestion
  test('matches snapshot with custom CSS class', () => {
```

src/components/WebexMeetingControl/WebexMeetingControls.test.js
```diff
@@ -13,4 +13,15 @@ describe('Webex Meeting Controls component', () => {
...
+    const meetingID = 'my-meeting';
+    const component = shallow(
+      <WebexMeetingControls className="my-custom-class" meetingID={meetingID}>
```
lalli-flores: I know you just copy-paste'd but just making an observation that for some reason we create a variable for the meeting ID on these tests, unlike hardcoding the ID directly like everywhere else

src/components/WebexMeetingInfo/WebexMeetingInfo.test.js
```diff
@@ -10,6 +10,10 @@ describe('Webex Meeting Info component', () => {
       expect(shallow(<WebexMeetingInfo meetingID="scheduledMeeting" />)).toMatchSnapshot();
     });
 
+    test('matches snapshot of scheduled meeting with custom class', () => {
```
lalli-flores:
```suggestion
    test('matches snapshot of scheduled meeting with custom CSS class', () => {
```

src/components/WebexRemoteMedia/WebexRemoteMedia.test.js
```diff
@@ -25,5 +26,9 @@ describe('Webex Remote Media component', () => {
     test('matches snapshot on error', () => {
       expect(shallow(<WebexRemoteMedia meetingID="failMeetingID" />)).toMatchSnapshot();
     });
+
+    test('matches snapshot of enabled remote audio & video with custom class', () => {
```
lalli-flores:
```suggestion
    test('matches snapshot of enabled remote audio & video with custom CSS class', () => {
```

src/components/WebexLocalMedia/WebexLocalMedia.js
```diff
@@ -16,30 +16,37 @@ import './WebexLocalMedia.scss';
...
   const videoRef = useStream(localVideo);
 
   const classBaseName = 'local-media';
-  const classObjects = {};
+  const mainClasses = {};
 
-  classObjects[`${WEBEX_COMPONENTS_CLASS_PREFIX}-${classBaseName}`] = true;
-  classObjects[`${WEBEX_COMPONENTS_CLASS_PREFIX}-${classBaseName}-desktop`] = width >= PHONE_LARGE;
-  classObjects[`${WEBEX_COMPONENTS_CLASS_PREFIX}-no-media`] = localVideo === null;
+  mainClasses[`${WEBEX_COMPONENTS_CLASS_PREFIX}-${classBaseName}`] = true;
```
lalli-flores: I know this change was done on a previous PR, but I would like to go back to the way the code looked before because I think that's much more readable (and the reason I felt comfortable with using the `classnames` library). The same can be achieved with computed property names.

```js
const baseCSSClass = `${WEBEX_COMPONENTS_CLASS_PREFIX}-local-media`;
const cssClasses = classNames({
  [baseCSSClass]: true,
  [`${baseCSSClass}-desktop`]: width >= PHONE_LARGE,
  ...,
 className
});
```
This is non-blocking for this since the code already exists, but hope is ok to open a PR later to refactor it.

adamweeks: I honestly had no idea that was an option.

img

adamweeks:
@lalli-flores took your suggestions and made everything **consistent** across all usage of `classNames`. This way there's no ambiguity on how to use the classname helper throughout the repo.

src/components/WebexLocalMedia/WebexLocalMedia.js
```diff
@@ -16,30 +16,37 @@ import './WebexLocalMedia.scss';
...
   const videoRef = useStream(localVideo);
 
-  const classBaseName = 'local-media';
-  const classObjects = {};
+  const classBaseName = `${WEBEX_COMPONENTS_CLASS_PREFIX}-local-media`;
+  const mainClasses = {
+    [classBaseName]: true,
+    desktop: width >= PHONE_LARGE,
+    'no-media': localVideo === null,
+  };
```
lalli-flores:
```suggestion
    className
  };
```

src/components/WebexLocalMedia/WebexLocalMedia.js
```diff
@@ -16,30 +16,37 @@ import './WebexLocalMedia.scss';
...
 
-  classObjects[`${WEBEX_COMPONENTS_CLASS_PREFIX}-${classBaseName}`] = true;
-  classObjects[`${WEBEX_COMPONENTS_CLASS_PREFIX}-${classBaseName}-desktop`] = width >= PHONE_LARGE;
-  classObjects[`${WEBEX_COMPONENTS_CLASS_PREFIX}-no-media`] = localVideo === null;
-
-  const cssClasses = classNames(classObjects);
+  if (className) {
```
lalli-flores: You can just add this as a last item on `mainClasses`, and there is no need to check it. 
Since `className` defaults to `''`, classNames will computed to false and ignore it, otherwise, whatever value it has will be added as-is, which would be the class name we want to add
See my suggestion.

src/components/WebexLocalMedia/WebexLocalMedia.js
```diff
@@ -16,30 +16,37 @@ import './WebexLocalMedia.scss';
...
+    mainClasses[className] = true;
+  }
 
   const disabledVideo = ID ? <WebexAvatar personID={ID} displayStatus={false} /> : <Spinner />;
 
   return (
-    <div ref={mediaRef} className={cssClasses}>
+    <div ref={mediaRef} className={classNames(mainClasses)}>
```
lalli-flores: Is there a reason you didn't invoke `classNames` at the const declaration?
The thing i liked about doing it that way is that is consistent with the way we do all other props, just passing an variable

src/components/WebexLocalMedia/WebexLocalMedia.js
```diff
@@ -16,30 +16,37 @@ import './WebexLocalMedia.scss';
...
   const {localVideo} = useMeeting(meetingID);
   const {ID} = useMe();
   const videoRef = useStream(localVideo);
 
-  const classBaseName = 'local-media';
-  const classObjects = {};
+  const classBaseName = `${WEBEX_COMPONENTS_CLASS_PREFIX}-local-media`;
+  const mainClasses = {
+    [classBaseName]: true,
+    desktop: width >= PHONE_LARGE,
```
lalli-flores: Why is this not a full class name like in In-Meeting? e.g. [`${classBaseName}-desktop`]

src/components/WebexLocalMedia/__snapshots__/WebexLocalMedia.test.js.snap
```diff
@@ -2,7 +2,7 @@
 
 exports[`Webex Local Media component snapshot matches snapshot of disabled local video 1`] = `
 <div
-  className="wxc-local-media wxc-no-media"
+  className="wxc-local-media no-media"
```
lalli-flores: Wouldn't we want `no-media` to also be prefixed? it seems like this could clash with other CSS class names

src/components/WebexMeeting/WebexMeeting.js
```diff
@@ -21,22 +21,23 @@ export default function WebexMeeting({meetingDestination, controls}) {
   const {ID, remoteVideo} = useMeetingDestination(meetingDestination);
   const isActive = remoteVideo !== null;
 
-  const classBaseName = 'meeting';
-  const classObjects = {};
+  const classBaseName = `${WEBEX_COMPONENTS_CLASS_PREFIX}-meeting`;
```
lalli-flores: NITPICK Lalli: can you remove this new line?

src/components/WebexMeetingControl/WebexMeetingControls.js
```diff
@@ -10,15 +15,28 @@ export const MeetingContext = React.createContext();
  * @param {object} props
  * @returns {object} JSX of the component
  */
-export default function WebexMeetingControls({meetingID, children}) {
+export default function WebexMeetingControls({children, className, meetingID}) {
+  const mainClasses = {
+    [`${WEBEX_COMPONENTS_CLASS_PREFIX}-meeting-controls`]: true,
+  };
+
+  if (className) {
```
lalli-flores: Like in my other comment, this check is not necessary, and could be shortened by just adding `className` as a last item in `mainClasses`

src/components/WebexLocalMedia/WebexLocalMedia.js
adamweeks: My reasoning on this was that we could inject some additional classes into `mainClasses` at a later point before the render. It also is the way it is done throughout the original widgets, so I carried it over.

src/components/WebexLocalMedia/WebexLocalMedia.js
adamweeks: I was just using the previous names, I will update

src/components/WebexLocalMedia/WebexLocalMedia.js
adamweeks: Doing this generates the classes as:
`className="wxc-local-media wxc-local-media-no-media className"`

I found that adding `[className]: !!className,` works as expected though

lalli-flores: Interesting that it makes it a string, literally, instead of evaluating the value

PR #102, opened 2020-06-16
feat(WebexMeetingSharing): add screen share meeting control and stories

yusenjeng:
To add screen share meeting control, remote sharing video in the WebexRemoteMedia, related json adapter and stories.

### DIFF
--- a/src/adapters/MeetingsJSONAdapter.js
+++ b/src/adapters/MeetingsJSONAdapter.js
@@ -5,6 +5,7 @@ import {MeetingsAdapter, MeetingControlState} from '@webex/component-adapter-int
 // Defined meeting controls in Meetings JSON Adapter
 export const MUTE_AUDIO_CONTROL = 'mute-audio';
 export const MUTE_VIDEO_CONTROL = 'mute-video';
+export const SHARE_CONTROL = 'share-screen';
 export const JOIN_CONTROL = 'join-meeting';
 export const LEAVE_CONTROL = 'leave-meeting';
 export const DISABLED_MUTE_AUDIO_CONTROL = 'disabled-mute-audio';
@@ -59,6 +60,12 @@ export default class MeetingsJSONAdapter extends MeetingsAdapter {
       ID: MUTE_VIDEO_CONTROL,
       action: this.toggleMuteVideo.bind(this),
       display: this.muteVideoControl.bind(this),
+    };
+
+    this.meetingControls[SHARE_CONTROL] = {
+      ID: SHARE_CONTROL,
+      action: this.handleLocalShare.bind(this),
+      display: this.shareControl.bind(this),
     };
 
     this.meetingControls[JOIN_CONTROL] = {
@@ -158,6 +165,15 @@ export default class MeetingsJSONAdapter extends MeetingsAdapter {
     );
 
     // Send updates on the meeting when an action is triggered
+    const shareEvents$ = fromEvent(document, SHARE_CONTROL).pipe(
+      tap(() => {
+        const meeting = this.datasource[ID];
+
+        // Use local screen share to fake a remote user's screen sharing
+        meeting.remoteShare = meeting.localShare;
+      })
+    );
+
     const audioEvents$ = fromEvent(document, MUTE_AUDIO_CONTROL);
     const videoEvents$ = fromEvent(document, MUTE_VIDEO_CONTROL);
     const joinEvents$ = fromEvent(document, JOIN_CONTROL);
@@ -165,7 +181,7 @@ export default class MeetingsJSONAdapter extends MeetingsAdapter {
       tap(() => end$.next(`Meeting "${ID}" has completed.`))
     );
 
-    const events$ = merge(audioEvents$, videoEvents$, joinEvents$, leaveEvents$).pipe(
+    const events$ = merge(audioEvents$, videoEvents$, joinEvents$, leaveEvents$, shareEvents$).pipe(
       filter((event) => event.detail.ID === ID),
       // Make a copy of the meeting to treat it as if were immutable
       map((event) => ({...event.detail}))
@@ -205,6 +221,26 @@ export default class MeetingsJSONAdapter extends MeetingsAdapter {
     }
 
     return stream;
+  }
+
+  /**
+   * Returns a promise to a MediaStream object that captures the contents of a user local display.
+   *
+   * @returns {Promise.<MediaStream>}
+   * @memberof MeetingJSONAdapter
+   * @private
+   */
+  async getDisplayStream() {
+    let captureStream = null;
+
+    try {
+      captureStream = await navigator.mediaDevices.getDisplayMedia();
+    } catch (reason) {
+      // eslint-disable-next-line no-console
+      console.error('Meetings JSON adapter can not display the local user sharing stream', reason);
+    }
+
+    return captureStream;
   }
 
   /**
@@ -436,6 +472,77 @@ export default class MeetingsJSONAdapter extends MeetingsAdapter {
 
       observer.complete();
     });
+  }
+
+  /**
+   * Handles the starting and stopping of the local screen capture media stream.
+   *
+   * @param {string} ID ID of the meeting for which to update display
+   * @memberof MeetingJSONAdapter
+   * @private
+   */
+  async handleLocalShare(ID) {
+    const meeting = this.datasource[ID];
+
+    if (meeting.localShare) {
+      meeting.localShare.getTracks()[0].stop();
+      meeting.localShare = null;
+    } else {
+      meeting.localShare = await this.getDisplayStream();
+
+      if (meeting.localShare) {
+        // Handle browser's built-in stop Button
+        meeting.localShare.getVideoTracks()[0].onended = () => {
+          meeting.localShare = null;
+          meeting.remoteShare = null;
+          document.dispatchEvent(new CustomEvent(SHARE_CONTROL, {detail: meeting}));
+        };
+      }
+    }
+
+    document.dispatchEvent(new CustomEvent(SHARE_CONTROL, {detail: meeting}));
+  }
+
+  /**
+   * Returns an observable that emits the display data of a meeting control.
+   *
+   * @param {string} ID ID of the meeting for which to update display
+   * @returns {Observable.<MeetingControlDisplay>}
+   * @memberof MeetingJSONAdapter
+   * @private
+   */
+  shareControl(ID) {
+    const inactive = {
+      ID: SHARE_CONTROL,
+      icon: 'share',
+      tooltip: 'Start Sharing',
+      state: MeetingControlState.INACTIVE,
+    };
+    const active = {
+      ID: SHARE_CONTROL,
+      icon: 'share',
+      tooltip: 'Stop Sharing',
+      state: MeetingControlState.ACTIVE,
+    };
+
+    const default$ = Observable.create((observer) => {
+      const meeting = this.datasource[ID];
+
+      if (meeting) {
+        observer.next(meeting.localShare ? active : inactive);
+      } else {
+        observer.error(new Error(`Could not find meeting with ID "${ID}"`));
+      }
+
+      observer.complete();
+    });
+
+    const shareEvent$ = fromEvent(document, SHARE_CONTROL).pipe(
+      filter((event) => event.detail.ID === ID),
+      map((event) => (event.detail.localShare ? active : inactive))
+    );
+
+    return concat(default$, shareEvent$);
   }
 
   /**
--- a/src/components/WebexMeeting/WebexMeeting.stories.js
+++ b/src/components/WebexMeeting/WebexMeeting.stories.js
@@ -11,7 +11,7 @@ const webexAdapter = new WebexJSONAdapter(jsonData);
 
 stories.add('default', () => (
   <WebexDataProvider adapter={webexAdapter}>
-    <WebexMeeting meetingDestination="localMedia" />
+    <WebexMeeting meetingDestination="remote&localMedia" />
   </WebexDataProvider>
 ));
 
--- a/src/components/WebexMeetingControl/WebexMeetingControl.stories.js
+++ b/src/components/WebexMeetingControl/WebexMeetingControl.stories.js
@@ -49,3 +49,13 @@ stories.add('disabled icon', () => (
     </WebexMeetingControls>
   </WebexDataProvider>
 ));
+
+stories.add('multiple icons', () => (
+  <WebexDataProvider adapter={webexAdapter}>
+    <WebexMeetingControls meetingID="localMedia">
+      <WebexMeetingControl type="mute-audio" />
+      <WebexMeetingControl type="mute-video" />
+      <WebexMeetingControl type="share-screen" />
+    </WebexMeetingControls>
+  </WebexDataProvider>
+));
--- a/src/components/WebexRemoteMedia/WebexRemoteMedia.js
+++ b/src/components/WebexRemoteMedia/WebexRemoteMedia.js
@@ -5,7 +5,6 @@ import {Badge, Spinner, AlertBanner} from '@momentum-ui/react';
 
 import {WEBEX_COMPONENTS_CLASS_PREFIX} from '../../constants';
 import {useMeeting, useStream} from '../hooks';
-
 import './WebexRemoteMedia.scss';
 
 /**
@@ -17,13 +16,16 @@ import './WebexRemoteMedia.scss';
  * NOTE: waiting for the UX for a design on what to display if there is no remote video
  */
 export default function WebexRemoteMedia({className, meetingID}) {
-  const {remoteAudio, remoteVideo, error} = useMeeting(meetingID);
+  const {remoteAudio, remoteVideo, remoteShare, error} = useMeeting(meetingID);
   const audioRef = useStream(remoteAudio);
   const videoRef = useStream(remoteVideo);
-  const hasMedia = !!(remoteAudio || remoteVideo);
+  const shareRef = useStream(remoteShare);
+  const hasMedia = !!(remoteAudio || remoteVideo || remoteShare);
+  const hasTwoMedia = remoteVideo && remoteShare;
   const mainClasses = {
     [`${WEBEX_COMPONENTS_CLASS_PREFIX}-remote-media`]: true,
     [className]: !!className,
+    'split-screen': hasTwoMedia,
   };
 
   return (
@@ -40,7 +42,11 @@ export default function WebexRemoteMedia({className, meetingID}) {
               <div>Connecting</div>
             </Badge>
           ) : null}
+
           {remoteVideo ? <video ref={videoRef} playsInline autoPlay /> : null}
+
+          {remoteShare ? <video ref={shareRef} playsInline autoPlay /> : null}
+
           {remoteAudio ? <audio ref={audioRef} autoPlay /> : null}
         </React.Fragment>
       )}
--- a/src/components/WebexRemoteMedia/WebexRemoteMedia.scss
+++ b/src/components/WebexRemoteMedia/WebexRemoteMedia.scss
@@ -27,4 +27,11 @@
     height: 100%;
     width: 100%;
   }
+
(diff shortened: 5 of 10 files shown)


src/components/hooks/useShareControl.js
```diff
@@ -0,0 +1,29 @@
...
+
+import {AdapterContext} from '../../components/';
+
+/**
+ * Custom hook that returns the display state of the meeting share control.
+ *
+ * @param {string} meetingID  ID of the meeting for which to get data
+ * @returns {shareControlDisplay}  Data state of the share control
+ */
+export default function useShareControl(meetingID) {
```
lalli-flores, line 11: Not sure I understand the purpose of this hook? `screen-share` should work like any other control and be useable through the [existing `useMeetingControl` hook](https://github.com/webex/components/blob/master/src/components/hooks/useMeetingControl.js)

src/components/WebexRemoteMedia/WebexRemoteMedia.js
```diff
@@ -17,15 +16,41 @@ import './WebexRemoteMedia.scss';
...
   };
 
+  const isLocalSharingInactive = shareControl.state === 'inactive';
+
+  //
+  // Switch the visibility of the remoteShare video by checking its readyState
+  // Note: Sharing detection can be removed from component, once the sdk provides
+  //       the Start/Stop events about the remote sharing.
+  //
+  const detectRemoteSharing = () => {
```
lalli-flores, line 37: This is not necessary in this component. `WebexMeeting` component handles the layering of local media + remote media + meeting controls. `WebexRemoteMedia` should only be concerned with displaying remote streams.
Do you have suggestions on how to make that clearer in the docs/repo?

src/data/meetings.json
```diff
@@ -120,13 +120,31 @@
   "localMedia": {
     "ID": "localMedia",
     "title": "Ad Hoc Meeting Space",
-    "localVideo": {},
+    "localVideo": {
+      "muted": false
```
lalli-flores, line 124: Is there a reason for adding the `muted` property?

src/adapters/MeetingsJSONAdapter.js
```diff
@@ -157,6 +164,17 @@ export default class MeetingsJSONAdapter extends MeetingsAdapter {
       /* eslint-enable no-confusing-arrow */
     );
 
+    const shareEvents$ = fromEvent(document, SHARE_CONTROL).pipe(
```
lalli-flores, line 167: I see that you're adding the `remoteShare` whenever the share control is activated, however, I think this should follow the pattern of the other streams, where you add local/remote screen share stream based on the JSON data.

src/components/WebexRemoteMedia/WebexRemoteMedia.js
yusenjeng: This is the tricky part. 

The idea of `if (remoteShare) then display <video>` doesn't work for the sdk, because the remoteShare/remoteVideo stream object will be always true value during the entire meeting. The difference is that it receives data or not, and the stream objects were always there.

And, unfortunately, the sdk doesn't provide the remote start/stop events as well, so the adapter can't provide the states of those remote media. That means UI components have no way to know the remote start/stop sharing and cannot do proper UI changes.

This is a sdk design problem plus a missing sdk feature.

But we still found a way to detect the remote event from the DOM video element as a workaround, that's detectRemoteSharing(). DOM element was managed by React, so the code exists in the component.

We can surely remove the detection function, but the UI will lose ability to provide feedbacks to the users about the changes of remote steam (with the current sdk).

src/adapters/MeetingsJSONAdapter.js
```diff
@@ -207,6 +225,24 @@ export default class MeetingsJSONAdapter extends MeetingsAdapter {
     return stream;
   }
 
+  /**
+   * Returns a MediaStream object obtained from user local sharing.
+   * @param {MediaStreamConstraints} displayMediaOptions  an object specifying the types of the media to request
+   * @returns {MediaStream}
+   */
+  async getSharingStream(displayMediaOptions) {
```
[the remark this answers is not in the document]
yusenjeng: got it

src/adapters/MeetingsJSONAdapter.js
```diff
@@ -438,6 +474,77 @@ export default class MeetingsJSONAdapter extends MeetingsAdapter {
...
+      if (meeting) {
+        observer.next(meeting.localShare ? active : inactive);
+      } else {
+        observer.error(new Error(`Could not find meeting with ID "${ID}"`));
+      }
+
+      observer.complete();
+    });
+
+    const muteEvent$ = fromEvent(document, SHARE_CONTROL).pipe(
```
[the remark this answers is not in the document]
yusenjeng: agree

src/components/WebexMeeting/WebexMeeting.stories.js
```diff
@@ -16,11 +16,33 @@ stories.add('default', () => (
 ));
 
 stories.add('custom controls', () => {
-  const controls = (isActive) => (isActive ? ['leave-meeting'] : ['join-meeting']);
+  const controls = (isActive) => {
```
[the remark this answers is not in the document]
yusenjeng: No strong reason for that. Will rollback and focus on the sharing part.

src/data/meetings.json
yusenjeng: No strong reason for that. Will rollback and focus on the sharing part.

src/components/hooks/useMeeting.js
```diff
@@ -16,6 +16,7 @@ export default function useMeeting(meetingID) {
     localVideo: null,
     remoteAudio: null,
```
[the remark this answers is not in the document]
yusenjeng: It's about UX.  There are still many sharing bugs inside the sdk. It fails often, so we are not sure to display or not to display the localShare on the sharing sender's screen at this point.

I prefer leave the application of localShare to the next version.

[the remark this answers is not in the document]
lalli-flores: Sounds good 👍🏻

src/data/meetings.json
```diff
@@ -122,11 +122,27 @@
...
+    },
     "remoteAudio": null,
     "localShare": null,
     "remoteShare": null
   },
+  "localMediaWithSharing": {
+    "ID": "localMediaWithSharing",
+    "title": "Ad Hoc Meeting Space",
+    "localVideo": {
```
lalli-flores, line 135: Since this is supposed to represent a meeting with screen share, the attributes that represent a display media stream should not be `null`, and instead be an object to indicate that "they exist". Like so:
```suggestion
    "localVideo": {},
    "remoteVideo": {},
    "localAudio": {},
    "remoteAudio": {},
    "localShare": {},
    "remoteShare": null
```

In this case, having an object in the `localShare` attribute indicates that the user is sharing their screen. Conversely, if `remoteShare` was an object `{}` instead of `null`, while `localShare` is `null`, this would represent a meeting where a remote participant is sharing their screen.

src/components/hooks/useShareControl.js
yusenjeng: Will reuse the useMeetingControl().

src/components/WebexRemoteMedia/WebexRemoteMedia.js
lalli-flores: A couple things here:
* The JSON adapter code seems to be making changes to `localShare` only, so why is the remote media component being changed?
* I think there might be some rework that needs to happen on the SDK adapter if what you're saying is how things are currently done. SDK issues/design should not be bubbling up to the components, thats why the adapter is there. The SDK adapter should do what it needs to to make sure that if there is something being shared, then it returns the right stream, whereas if nothing is being shared, then the attribute should be `null`. This is why in the SDK adapter there are two objects `meeting` which is the component systems state, and `sdkMeeting`, which is the state of how the SDK handles meetings; the job of the adapter to to translate `sdkMeeting` into `meeting` so that the components don't have to know any of those implementation details.

src/data/meetings.json
```diff
@@ -122,11 +122,27 @@
     "title": "Ad Hoc Meeting Space",
     "localVideo": {},
     "remoteVideo": null,
-    "localAudio": {},
+    "localAudio": {
+      "muted": false
```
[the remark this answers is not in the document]
yusenjeng: sure

src/data/meetings.json
yusenjeng: How about to set both the `localShare` and the `remoteShare` to `null`?  This scenario requires the user to click on the sharing button by themself. The sharing video isn't acquired automatically on starting.

src/adapters/MeetingsJSONAdapter.js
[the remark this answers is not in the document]
yusenjeng: agree

src/components/hooks/useStream.js
```diff
@@ -12,7 +12,7 @@ export default function useStream(stream) {
   useEffect(() => {
     const mediaElement = streamRef.current;
 
-    if (stream instanceof MediaStream) {
+    if (!!mediaElement && stream instanceof MediaStream) {
```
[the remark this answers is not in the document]
yusenjeng: ok

src/data/meetings.json
lalli-flores: I'm ok with that, if it's not necessary to add the share stream. In that case, I would look to the existing objects to see if you can reuse any of them instead of adding new one

yusenjeng: good idea. Will reuse "remote&localMedia" instead of the new one.

src/components/WebexRemoteMedia/WebexRemoteMedia.js
```diff
@@ -17,15 +16,40 @@ import './WebexRemoteMedia.scss';
...
+    if (!shareRef.current) {
+      return false;
+    }
+
+    return shareRef.current.readyState > 1;
+  };
+
+  const remoteShareClasses = {
```
lalli-flores, line 44: Are these necessary? If they are, this should probably be unified at the component level in [`mainClasses`](https://github.com/webex/components/blob/master/src/components/WebexRemoteMedia/WebexRemoteMedia.js#L24)

src/components/WebexMeeting/WebexMeeting.stories.js
```diff
@@ -16,11 +16,33 @@ stories.add('default', () => (
...
+  };
 
   return (
     <WebexDataProvider adapter={webexAdapter}>
       <WebexMeeting meetingDestination="localMedia" controls={controls} />
     </WebexDataProvider>
   );
 });
+
+stories.add('custom controls with sharing', () => {
```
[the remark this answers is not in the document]
lalli-flores: I'd still like to see it done this way

src/components/WebexRemoteMedia/WebexRemoteMedia.js
```diff
@@ -40,7 +64,13 @@ export default function WebexRemoteMedia({className, meetingID}) {
               <div>Connecting</div>
             </Badge>
           ) : null}
-          {remoteVideo ? <video ref={videoRef} playsInline autoPlay /> : null}
+
+          {remoteVideo ? <video ref={videoRef} playsInline autoPlay name="remote-cam" /> : null}
```
lalli-flores, line 68: I don't think [`name` is an attribute of `video`](https://developer.mozilla.org/en-US/docs/Web/API/Element/name) elements

```suggestion
          {remoteVideo ? <video ref={videoRef} playsInline autoPlay /> : null}
```

src/components/WebexRemoteMedia/WebexRemoteMedia.js
```diff
@@ -40,7 +64,13 @@ export default function WebexRemoteMedia({className, meetingID}) {
...
+
+          {remoteShare && isLocalSharingInactive ? (
+            <video ref={shareRef} playsInline autoPlay name="remote-share" className={classNames(remoteShareClasses)} />
```
lalli-flores, line 71: I don't think [`name` is an attribute of `video`](https://developer.mozilla.org/en-US/docs/Web/API/Element/name) elements
```suggestion
            <video ref={shareRef} playsInline autoPlay className={classNames(remoteShareClasses)} />
```

src/components/WebexRemoteMedia/WebexRemoteMedia.scss
```diff
@@ -26,5 +26,17 @@
   video {
     height: 100%;
     width: 100%;
+
+    &.share {
```
lalli-flores, line 30: Is something using this?

package.json
```diff
@@ -39,6 +39,7 @@
     "@juggle/resize-observer": "^3.0.2",
     "@momentum-ui/react": "^23.2.1",
     "@webex/component-adapter-interfaces": "^1.9.0",
+    "@webex/sdk-component-adapter": "file:../sdk-adapter-official",
```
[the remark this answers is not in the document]
yusenjeng: It shouldn't be there.

src/components/WebexInMeeting/WebexInMeeting.js
```diff
@@ -27,7 +27,9 @@ export default function WebexInMeeting({meetingID}) {
 
   return (
     <div ref={meetingRef} className={classNames(mainClasses)}>
-      <WebexRemoteMedia className="remote-media-in-meeting" meetingID={meetingID} />
+      <MeetingContext.Provider value={meetingID}>
```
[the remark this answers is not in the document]
yusenjeng: The useMeetingControl hook needs this to provide the context. <WebexMeetingControls> is a similar scenario. Its sub-components also use useMeetingControl

src/components/WebexRemoteMedia/WebexRemoteMedia.js
yusenjeng: This custom attribute was used to differentiate the two video element when they appeared at the same time. Removing them is fine.

src/components/WebexRemoteMedia/WebexRemoteMedia.scss
yusenjeng: Yes, it's part of remoteShareClasses used to control the visibility of sharing video (not the cam video).

src/components/WebexRemoteMedia/WebexRemoteMedia.js
yusenjeng: It is used to stylized one single video element inside the `<WebexRemoteMedia>`, not the whole component. However, the video element itself is not big enough to be a standalone component.

src/components/WebexMeetingControl/WebexMeetingControl.stories.js
```diff
@@ -49,3 +49,13 @@ stories.add('disabled icon', () => (
     </WebexMeetingControls>
   </WebexDataProvider>
 ));
+
+stories.add('multiple icons', () => (
+  <WebexDataProvider adapter={webexAdapter}>
+    <WebexMeetingControls meetingID="localMediaWithSharing">
```
lalli-flores, line 55: This meeting ID no longer exists, so the story is not displaying properly

src/components/WebexInMeeting/WebexInMeeting.js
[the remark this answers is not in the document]
lalli-flores: Maybe I haven't had enough coffee yet but the `useMeetingControl` hook fetches the appropriate context already, why is this necessary?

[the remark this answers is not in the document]
yusenjeng: The MeetingContext will be removed for simplification.

src/components/WebexRemoteMedia/WebexRemoteMedia.js
yusenjeng: The `remoteShareClasses` will be removed. Instead, we choose a better way to do split screen.

src/components/WebexRemoteMedia/WebexRemoteMedia.scss
yusenjeng: This part is also changed for the removal of `remoteShareClasses`.

src/components/index.js
```diff
@@ -2,11 +2,11 @@ export {default as WebexAvatar} from './WebexAvatar/WebexAvatar';
...
 export {default as WebexDataProvider, AdapterContext} from './WebexDataProvider/WebexDataProvider';
-export {default as WebexInMeeting} from './WebexInMeeting/WebexInMeeting';
-export {default as WebexInterstitialMeeting} from './WebexInterstitialMeeting/WebexInterstitialMeeting';
 export {default as WebexLocalMedia} from './WebexLocalMedia/WebexLocalMedia';
 export {default as WebexMeeting} from './WebexMeeting/WebexMeeting';
 export {default as WebexMeetingControl} from './WebexMeetingControl/WebexMeetingControl';
 export {default as WebexMeetingControls, MeetingContext} from './WebexMeetingControl/WebexMeetingControls';
 export {default as WebexMeetingInfo} from './WebexMeetingInfo/WebexMeetingInfo';
 export {default as WebexRemoteMedia} from './WebexRemoteMedia/WebexRemoteMedia';
+export {default as WebexInMeeting} from './WebexInMeeting/WebexInMeeting';
```
lalli-flores, line 11: Can you revert this change? We wanted to keep the exports alphabetica

package.json
```diff
@@ -149,4 +149,4 @@
       "commit-msg": "commitlint -E HUSKY_GIT_PARAMS"
     }
   }
-}
+}
```
lalli-flores, line 152: Annoying change but can you add a new line at the end?

package-lock.json
```diff
@@ -22227,4 +22227,4 @@
       }
     }
   }
-}
+}
```
lalli-flores, line 22230: Annoying change but can you add a new line at the end?

src/adapters/MeetingsJSONAdapter.js
```diff
@@ -438,6 +474,78 @@ export default class MeetingsJSONAdapter extends MeetingsAdapter {
...
 
+  /**
+   * Handles the starting and stopping of the local screen capture media stream.
+   *
+   * @param {string} ID ID of the meeting for which to update display
+   * @memberof MeetingJSONAdapter
+   * @private
+   */
+  async handleLocalShare(ID) {
+    console.log('handleLocalShare', ID);
```
lalli-flores, line 485: Can you remove this please?

src/components/index.js
yusenjeng: ok

package.json
yusenjeng: sure

package-lock.json
yusenjeng: sure

src/adapters/MeetingsJSONAdapter.js
yusenjeng: already removed

src/components/WebexMeetingControl/WebexMeetingControl.stories.js
yusenjeng: use localMedia instead

package-lock.json
```diff
@@ -3682,6 +3682,18021 @@
       "integrity": "sha512-LfmH444YDTmvc0zqqOm55a1WxVk89XMh/Eq/Sv0YaW8CJCp2m1Ib3jKLl9+Qu9F3dSHpiCxHfjW+k++RX3NwVQ==",
       "dev": true
     },
+    "@webex/sdk-component-adapter": {
```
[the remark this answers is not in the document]
yusenjeng: it's done

src/components/WebexRemoteMedia/WebexRemoteMedia.test.js
```diff
@@ -4,6 +4,7 @@ import WebexRemoteMedia from './WebexRemoteMedia';
 
 jest.mock('../hooks/useMeeting');
 jest.mock('../hooks/useStream');
+jest.mock('../hooks/useMeetingControl');
```
lalli-flores, line 7: Can you remove this? This hook is no longer used in `WebexRemoteMedia`

PR #103, opened 2020-07-01
fix(packages): update momentum-ui

stevem63:
bump up the version of momentum-ui

### DIFF
--- a/package.json
+++ b/package.json
@@ -37,7 +37,7 @@
   },
   "dependencies": {
     "@juggle/resize-observer": "^3.0.2",
-    "@momentum-ui/react": "^23.2.1",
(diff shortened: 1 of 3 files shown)


stevem63:
@lalli-flores fixed the commit message, and yes, I verified the components look correct in the stories

package.json
```diff
@@ -37,7 +37,7 @@
   },
   "dependencies": {
     "@juggle/resize-observer": "^3.0.2",
-    "@momentum-ui/react": "^23.2.1",
+    "@momentum-ui/react": "^23.14.0",
```
lalli-flores, line 40: Seems like the latest version is [23.15.3](https://www.npmjs.com/package/@momentum-ui/react), we should upgrade to that

PR #104, opened 2020-07-06
fix(hooks): cloned members before setting state

yusenjeng:
`useParticipants` hook makes a shallow copy the `members` array of a membership. Without sending a new object, React does not detect that there has been a change to the participant list.

Bug tracked by [SPARK-158264](https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-158264)

### DIFF
--- a/src/components/hooks/useParticipants.js
+++ b/src/components/hooks/useParticipants.js
@@ -20,7 +20,7 @@ export default function useParticipants(membershipID) {
       console.error(error.message);
     };
     const onMembers = (data) => {
-      setParticipants(data.members);
+      setParticipants([...data.members]);
     };
 
     const subscription = membershipsAdapter.getMembers(membershipID).subscribe(onMembers, onError);


src/components/hooks/useParticipants.js
```diff
@@ -20,7 +20,7 @@ export default function useParticipants(membershipID) {
       console.error(error.message);
     };
     const onMembers = (data) => {
-      setParticipants(data.members);
+      setParticipants(data.members.slice());
```
lalli-flores, line 23: Can you use the spread operator instead?
```suggestion
      setParticipants([...data.members]);
```

lalli-flores: Note that either approach makes a shallow copy of `members`, but i think it should be ok

yusenjeng: cool

PR #105, opened 2020-07-07
fix(roster): add key property to WebexParticipant

yusenjeng:
It's a 1 liner PR to fix a bug that the dynamic components (generated from loops) should come with a key property.
JIRA:
https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-158618

### DIFF
--- a/src/components/WebexParticipantRoster/WebexParticipantRoster.js
+++ b/src/components/WebexParticipantRoster/WebexParticipantRoster.js
@@ -15,7 +15,9 @@ import useParticipants from '../hooks/useParticipants';
 export default function WebexParticipantRoster({destination}) {
   const participants = useParticipants(destination);
 
-  const participantList = participants.map((participant) => <WebexParticipant personID={participant.personID} />);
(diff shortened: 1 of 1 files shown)


src/components/WebexParticipantRoster/WebexParticipantRoster.js
```diff
@@ -15,7 +15,9 @@ import useParticipants from '../hooks/useParticipants';
 export default function WebexParticipantRoster({destination}) {
   const participants = useParticipants(destination);
 
-  const participantList = participants.map((participant) => <WebexParticipant personID={participant.personID} />);
+  const participantList = participants.map((participant) => (
+    <WebexParticipant personID={participant.personID} key={participant.personID} />
```
lalli-flores, line 19: Can you have the `key` attribute first, like we do in [activity stream](https://github.com/webex/components/blob/master/src/components/WebexActivityStream/WebexActivityStream.js#L41)?
```suggestion
    <WebexParticipant key={participant.personID} personID={participant.personID} />
```

PR #110, opened 2020-07-31
semantic-release + commitlint + Husky Updates

lalli-flores:
Changes include:
* Updating `semantic-release` + commitlint + Husky dependencies
* Moving release config to `package.json`
* Have Husky run tests before pushing to remote
* Moving `classnames` from a dev dependency to a regular dependency


package.json
```diff
@@ -139,7 +138,38 @@
   "husky": {
     "hooks": {
       "pre-commit": "npm run test:eslint",
+      "pre-push": "npm run test",
```
adamweeks, line 141: Nice!

PR #116, opened 2020-08-25
Momentum UI as peer dependency of UMD bundle

lalli-flores:
Momentum UI as peer dependency of UMD bundle. Momentum UI causes build issues with their Sass. It's pretty heavy too and found that there is an AMD package hosted in [unpkg](https://unpkg.com/browse/@momentum-ui/react).

Moving outside of our bundle for performance purposes, as it is also a x10 reduction in size for the components UMD bundle.
Momentum UI CSS is also moved to its own file (another x10 reduction in size).

Adding `unpkg` to `package.json` so Webex. Components can be hosted properly in [unpkg](https://unpkg.com).
Added back JSON adapter as part of the distribution, as it can be useful for quick testing.

with-momentum
without-momentum

### DIFF
--- a/package.json
+++ b/package.json
@@ -78,6 +78,7 @@
     "rimraf": "^3.0.2",
     "rollup": "^2.26.4",
     "rollup-plugin-copy": "^3.3.0",
+    "rollup-plugin-license": "^2.2.0",
     "rollup-plugin-scss": "^2.6.0",
     "rollup-plugin-terser": "^7.0.0",
     "rollup-plugin-visualizer": "^4.1.0",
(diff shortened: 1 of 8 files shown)


package.json
```diff
@@ -159,11 +161,11 @@
         {
           "assets": [
             {
-              "path": "dist/webex-components.es.min.js",
+              "path": "dist/es/webex-components.es.min.js",
```
adamweeks, line 164: What's the purpose of having both `es` in the folder name and file name?

lalli-flores: I added the folders because it makes the `dist` folder much cleaner.
The bundle type was already part of the name. I left it because I think it's useful when locally copying the bundles for testing purposes, as it's easier to tell what the contents are when the bundle file is moved outside the `dist` folder context.

PR #121, opened 2020-08-31
feat(RemoteMedia): add waiting state

adamweeks:
The front end portion of the https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-136654 ticket.

### DIFF
--- a/src/components/WebexRemoteMedia/WebexRemoteMedia.jsx
+++ b/src/components/WebexRemoteMedia/WebexRemoteMedia.jsx
@@ -1,16 +1,19 @@
+/* eslint-disable jsx-a11y/media-has-caption */
 import React from 'react';
 import PropTypes from 'prop-types';
 import classNames from 'classnames';
 import {Badge, Spinner, AlertBanner} from '@momentum-ui/react';
 
 import {WEBEX_COMPONENTS_CLASS_PREFIX} from '../../constants';
-import {useMeeting, useStream} from '../hooks';
+import {useMeeting, useMemberships, useStream} from '../hooks';
 import './WebexRemoteMedia.scss';
 
 /**
  * Webex Remote Media component displays the meeting's remote video
  *
  * @param {object} props
+ * @param {string} props.className
+ * @param {string} props.meetingID
  * @returns {object} JSX of the component
  *
  * NOTE: waiting for the UX for a design on what to display if there is no remote video
@@ -22,9 +25,11 @@ export default function WebexRemoteMedia({className, meetingID}) {
     remoteShare,
     error,
   } = useMeeting(meetingID);
+  const members = useMemberships(meetingID, 'meeting');
   const audioRef = useStream(remoteAudio);
   const videoRef = useStream(remoteVideo);
   const shareRef = useStream(remoteShare);
+  const hasOtherMembers = members.length > 1;
   const hasMedia = !!(remoteAudio || remoteVideo || remoteShare);
   const hasTwoMedia = remoteVideo && remoteShare;
   const mainClasses = {
@@ -32,29 +37,42 @@ export default function WebexRemoteMedia({className, meetingID}) {
     [className]: !!className,
     'split-screen': hasTwoMedia,
   };
+  let remoteDisplay;
+
+  if (error) {
+    remoteDisplay = (
+      <AlertBanner show type="warning">
+        Having trouble joining the meeting? Please check your connection.
+      </AlertBanner>
+    );
+  } else if (hasMedia && hasOtherMembers) {
+    remoteDisplay = (
+      <>
+        {remoteVideo ? <video ref={videoRef} playsInline autoPlay /> : null}
+
+        {remoteShare ? <video ref={shareRef} playsInline autoPlay /> : null}
+
+        {remoteAudio ? <audio ref={audioRef} autoPlay /> : null}
+      </>
+    );
+  } else if (hasMedia && !hasOtherMembers) {
+    remoteDisplay = (
+      <h4>
+        Waiting for others to join...
+      </h4>
+    );
+  } else {
+    remoteDisplay = (
+      <Badge rounded>
+        <Spinner size={16} />
+        <div>Connecting</div>
+      </Badge>
+    );
+  }
 
   return (
     <div className={classNames(mainClasses)}>
-      {error ? (
-        <AlertBanner show type="warning">
-          Having trouble joining the meeting? Please check your connection.
-        </AlertBanner>
-      ) : (
-        <>
-          {!hasMedia ? (
-            <Badge rounded>
-              <Spinner size={16} />
-              <div>Connecting</div>
-            </Badge>
-          ) : null}
-
-          {remoteVideo ? <video ref={videoRef} playsInline autoPlay /> : null}
-
-          {remoteShare ? <video ref={shareRef} playsInline autoPlay /> : null}
-
-          {remoteAudio ? <audio ref={audioRef} autoPlay /> : null}
-        </>
-      )}
+      {remoteDisplay}
     </div>
   );
 }
--- a/src/components/WebexRemoteMedia/WebexRemoteMedia.test.js
+++ b/src/components/WebexRemoteMedia/WebexRemoteMedia.test.js
@@ -2,6 +2,7 @@ import React from 'react';
 
 import WebexRemoteMedia from './WebexRemoteMedia';
 
+jest.mock('../hooks/useMemberships');
 jest.mock('../hooks/useMeeting');
 jest.mock('../hooks/useStream');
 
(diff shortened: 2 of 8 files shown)


src/components/WebexRemoteMedia/WebexRemoteMedia.jsx
```diff
@@ -22,9 +25,11 @@ export default function WebexRemoteMedia({className, meetingID}) {
     remoteShare,
     error,
   } = useMeeting(meetingID);
+  const members = useMemberships(meetingID, 'meeting');
   const audioRef = useStream(remoteAudio);
   const videoRef = useStream(remoteVideo);
   const shareRef = useStream(remoteShare);
+  const hasMembers = members.length > 1;
```
lalli-flores, line 32: `hasMembers` by itself is a little misleading, as "self" is a current member. We want to know if there are others besides self.

```suggestion
  const hasOtherMembers = members.length > 1;
```

src/components/WebexRemoteMedia/WebexRemoteMedia.jsx
```diff
@@ -41,18 +46,29 @@ export default function WebexRemoteMedia({className, meetingID}) {
         </AlertBanner>
       ) : (
         <>
-          {!hasMedia ? (
+          {hasMedia && hasMembers && (
```
lalli-flores, line 49: This is why I don't like the `&&` format... I personally find it harder to reason about when there are multiple conditions.
Maybe we can rearrange the conditions/JSX?
```
  const remoteDisplay = hasOtherMembers ? <media JSX> : <waiting for others to join JSX>;

...

  hasMedia ? remoteDisplay : <connecting JSX>
```

if that makes sense

src/data/memberships.json
```diff
@@ -4,5 +4,41 @@
     "destinationID": "default_membership",
     "destinationType": "room",
     "members": [{"personID": "default"}, {"personID": "self"}, {"personID": "call"}, {"personID": "presenting"}]
+  },
+  "remoteVideo": {
```
lalli-flores, line 8: Are most of these memberships being used? I only see `noMembers` used in the stories/test
We should remove them if we're not currently using them

src/components/WebexRemoteMedia/WebexRemoteMedia.test.js
```diff
@@ -11,6 +12,10 @@ describe('Webex Remote Media component', () => {
       expect(shallow(<WebexRemoteMedia meetingID="noMedia" />)).toMatchSnapshot();
     });
 
+    test('matches snapshot of no members', () => {
```
lalli-flores, line 15:
```suggestion
    test('matches snapshot of no remote members', () => {
```

src/components/WebexRemoteMedia/WebexRemoteMedia.jsx
adamweeks: There were multiple nested ifs causing eslint issues, but I will re-investigate

src/data/memberships.json
adamweeks: Every test and story gets the membership object for the meeting ID so I had to add all of them. It would be easier if the memberships adapter got the data from the meeting json instead but it doesn't.

lalli-flores: Got it, just wanted to understand since I'm making changes to the data in an upcoming PR. Can probably update that then

src/data/meetings.json
```diff
@@ -146,5 +146,15 @@
     "remoteAudio": {},
     "localShare": null,
     "remoteShare": null
+  },
+  "noMembers": {
+    "ID": "remote&localMedia",
```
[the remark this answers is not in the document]
adamweeks: Yup, copy pasta fail

src/components/WebexRemoteMedia/WebexRemoteMedia.jsx
```diff
@@ -41,18 +46,29 @@ export default function WebexRemoteMedia({className, meetingID}) {
...
+
+              {remoteShare ? <video ref={shareRef} playsInline autoPlay /> : null}
+
+              {remoteAudio ? <audio ref={audioRef} autoPlay /> : null}
+            </>
+          )}
+          {hasMedia && !hasMembers && (
+            <>
+              <Badge rounded>
+                <Spinner size={16} />
```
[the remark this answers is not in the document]
adamweeks: Ok

adamweeks:
@lalli-flores all comments addressed

src/components/WebexRemoteMedia/WebexRemoteMedia.jsx
```diff
@@ -22,16 +25,37 @@ export default function WebexRemoteMedia({className, meetingID}) {
...
   const hasMedia = !!(remoteAudio || remoteVideo || remoteShare);
   const hasTwoMedia = remoteVideo && remoteShare;
   const mainClasses = {
     [`${WEBEX_COMPONENTS_CLASS_PREFIX}-remote-media`]: true,
     [className]: !!className,
     'split-screen': hasTwoMedia,
   };
+  let remoteDisplay;
+
+  if (hasMedia) {
```
lalli-flores, line 42: Is the `if` necessary? I doesn't affect what the JSX in `remoteDisplay` will be right?
Additionally, you have the check [in line 68](https://github.com/webex/components/pull/121/files#diff-e17d52c72e28c31e65dc68aa7ddf2314R68)

src/components/WebexRemoteMedia/WebexRemoteMedia.jsx
```diff
@@ -22,16 +25,37 @@ export default function WebexRemoteMedia({className, meetingID}) {
...
+      <>
+        {remoteVideo ? <video ref={videoRef} playsInline autoPlay /> : null}
+
+        {remoteShare ? <video ref={shareRef} playsInline autoPlay /> : null}
+
+        {remoteAudio ? <audio ref={audioRef} autoPlay /> : null}
+      </>
+    ) : (
+      <>
+        <Badge rounded>
```
lalli-flores, line 53: This still has the badge, and so it renders like this:
with badge

This is what the UI in the client is, which, I'm totally guessing here, but looks like an [`h4` from Momentum typography](https://momentum.design/styles/typography/style)?
client

src/components/WebexRemoteMedia/WebexRemoteMedia.jsx
mwegman: Is it necessary? No. It saves a couple of memory allocations. It could make it easier for the next person looking at this to see `remoteDisplay` will not be rendered if there is no media.

src/components/WebexRemoteMedia/WebexRemoteMedia.jsx
```diff
@@ -22,16 +25,37 @@ export default function WebexRemoteMedia({className, meetingID}) {
...
+    remoteDisplay = hasOtherMembers ? (
+      <>
+        {remoteVideo ? <video ref={videoRef} playsInline autoPlay /> : null}
+
+        {remoteShare ? <video ref={shareRef} playsInline autoPlay /> : null}
+
+        {remoteAudio ? <audio ref={audioRef} autoPlay /> : null}
+      </>
+    ) : (
+      <>
```
mwegman, line 52: The Fragment isn't necessary since you are only rendering a `Badge` element.

src/components/WebexRemoteMedia/WebexRemoteMedia.jsx
lalli-flores: I actually thought it was confusing because the real check to render or not is in line 68, so it felt out of place

src/components/WebexRemoteMedia/WebexRemoteMedia.jsx
```diff
@@ -41,18 +65,12 @@ export default function WebexRemoteMedia({className, meetingID}) {
         </AlertBanner>
       ) : (
         <>
-          {!hasMedia ? (
+          {hasMedia ? remoteDisplay : (
             <Badge rounded>
               <Spinner size={16} />
               <div>Connecting</div>
             </Badge>
```
mwegman, line 72: You could assign this to `remoteDisplay` above and keep your if statement.

src/components/WebexRemoteMedia/WebexRemoteMedia.jsx
```diff
@@ -22,39 +25,54 @@ export default function WebexRemoteMedia({className, meetingID}) {
...
+        {remoteShare ? <video ref={shareRef} playsInline autoPlay /> : null}
 
-          {remoteShare ? <video ref={shareRef} playsInline autoPlay /> : null}
+        {remoteAudio ? <audio ref={audioRef} autoPlay /> : null}
+      </>
+    );
+  } else if (hasMedia && !hasOtherMembers) {
+    remoteDisplay = (
+      <h4>
+        <div>Waiting for others to join...</div>
```
lalli-flores, line 61: Headings are block-type elements. The `div` is not necessary in this case
```suggestion
        Waiting for others to join...
```

PR #124, opened 2020-09-17
fix(WebexRemoteMedia): correct color for waiting text

adamweeks:


mwegman (approved):

lalli-flores (changes requested):
How will the white text show through when there is a white background?

adamweeks:
> How will the white text show through when there is a white background?

@lalli-flores the background is black

image

lalli-flores:
That's incorrect, I specifically mentioned this to you. The screenshot that you have is for the black theme, which we don't support. The components styles are based on the default theme which is white background with gray text.

adamweeks:
> That's incorrect, I specifically mentioned this to you. The screenshot that you have is for the dark theme, which we don't support. The components styles are based on the default theme which is white background with gray text.

@lalli-flores The in meeting component sets the background to black https://github.com/webex/components/blob/master/src/components/WebexInMeeting/WebexInMeeting.scss#L13

lalli-flores:
@adamweeks, In-Meeting is a different component, not sure why you bring it up?

However, the black background in In-Meeting is to "simulate" a video stream feel while the actual video packets start flowing, not because it's using the dark theme.

adamweeks:
@lalli-flores while the in meeting component is different, the remote media component is displayed on top of it and has no background, so the background will be what is behind it which is the in meeting component. What is your suggested change because right now it is dark gray text on a black background.

lalli-flores:
Got it, my suggestion is that if the background is the problem then that's what should be fixed?
Setting the background to white should make the component match the default style.

adamweeks:
> Got it, my suggestion is that if the background is the problem then that's what should be fixed?
> Setting the background to white should make the component match the default style. But do not set the background on the component as a whole, it should only be on the waiting state

@lalli-flores Are you saying we need to switch the "waiting for" display to have a white background over the top of the black background of the in meeting component?

lalli-flores:
@adamweeks, I guess? Components in general should not be concerned with how other components behave, so whether In-Meeting displays a black background or not should not affect how RemoteMedia is supposed to behave.
The default style for the waitings state is a white background, so that should be what this component should look like at that state.
waiting for others

adamweeks:
@lalli-flores color change has been implemented, background for remote media component set to white.

PR #127, opened 2020-09-17
Storybook v6

lalli-flores:
* Updates Storybook version to latest at the time (v6)
* Updates to Storybook stories to use Component Story Format (see [benefits of CSF](https://storybook.js.org/docs/react/api/csf))
* Removes snapshot tests in favor of Storyshots
** By using Storyshots, we save development time as we don't have to write those kinds of tests anymore (they were the majority!)
** By using Storyshots, we are not using the real React hooks (instead of mock hooks) which will allow us to actually catch issues with the components (code quality up!)
** Because the real hooks are now used, our test coverage is up
* Updated data to simulate more real scenarios, hopefully making stories more meaningful


adamweeks (changes requested):
We cannot merge this until the tests are passing.

.storybook/preview.js
```diff
@@ -0,0 +1,54 @@
...
+          'Webex Avatar',
+          'Webex Member Roster',
+          'Webex Member',
+        ],
+        'Messaging',
+        [
+          'Webex Activity Stream',
+          'Webex Activity',
+        ],
+        'Meetings',
```
adamweeks, line 41: I like the grouping!

src/components/WebexMeeting/WebexMeeting.stories.js
```diff
@@ -1,26 +1,29 @@
...
 
-import jsonData from '../../data';
-import {WebexJSONAdapter} from '../../adapters';
+export default {
+  title: 'Meetings/Webex Meeting',
+  component: WebexMeeting,
+};
 
-import {WebexMeeting, WebexDataProvider} from '..';
+const Template = (args) => <WebexMeeting {...args} />;
```
adamweeks, line 9: I've found that the current implementation of the WebexMeeting and child components currently rely on a container of a set size. The min height css does not propagate down and causes it to look like this:

image

Adding the fixed width container gets us a more working example:

```suggestion
const Template = (args) => <div style={{width: '650px', height: '500px'}}><WebexMeeting {...args} /></div>;
```

src/components/WebexInterstitialMeeting/WebexInterstitialMeeting.stories.js
```diff
@@ -1,43 +1,24 @@
...
-import {WebexInterstitialMeeting, WebexDataProvider} from '..';
+export default {
+  title: 'Meetings/Webex Interstitial Meeting',
+  component: WebexInterstitialMeeting,
+};
 
-// Setup for the stories
-const stories = storiesOf('Webex Interstitial Meeting', module);
-const webexAdapter = new WebexJSONAdapter(jsonData);
-const wrapperStyle = {height: '500px', width: '800px', border: '1px solid black'};
```
adamweeks, line 11: The wrapper needs to stay due to the way the meetings components require a fixed sized container.

src/components/WebexMeeting/WebexMeeting.stories.js
lalli-flores: I'm aware of this issue. I also found some other styling issues that I wanted to fix but I wanted to open a separate PR for it since this was too unwieldy already

lalli-flores:
> We cannot merge this until the tests are passing.

@adamweeks, I wouldn't mind an extra pair of eyes

src/components/WebexMeeting/WebexMeeting.stories.js
adamweeks: Since the code was already in for creating a fixed size container, let's continue using that strategy instead of removing it and breaking the display.

adamweeks:
Since the failing tests seem to be the main issue on this, let's just keep this PR at upgrading storybook and the stories. We can investigate story snapshots in another PR.

src/components/WebexMeeting/WebexMeeting.stories.js
lalli-flores: Added the styles back in the mean time

lalli-flores:
@adamweeks, tests are fixed

mwegman:
This is massive. Is there any way to break this up?

lalli-flores:
@mwegman, I kept warning for weeks how massive this was going to be 😬
It's mostly due to the fact that changes need to apply to _all components_ and there's no way to break that up. Some of the changes _could_ be in a different PR, but the tests do not all pass on every commit in-between and it's a lot of work to get them working at this point

adamweeks (approved):

PR #128, opened 2020-09-21
fix(WebexRemoteMedia): move useStream usage to own component

adamweeks:
When joining a meeting with no members, the "waiting" screen is displayed. Once another member joins, the stream reference it attempts to map to the video element is no longer valid. Moving the `useStream` hook into a sub component that only will call it when needed fixes the issue.

Reference: https://github.com/facebook/react/issues/11163#issuecomment-628379291

### DIFF
--- a/src/components/WebexRemoteMedia/RemoteMediaDisplay.jsx
+++ b/src/components/WebexRemoteMedia/RemoteMediaDisplay.jsx
@@ -1,0 +1,42 @@
+/* eslint-disable jsx-a11y/media-has-caption, react/forbid-prop-types */
+import React from 'react';
+import PropTypes from 'prop-types';
+
+import {useStream} from '../hooks';
+
+/**
+ * Remote Media Display component displays the meeting's remote video and audio html elements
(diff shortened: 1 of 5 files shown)


lalli-flores (changes requested):
Not sure I understand how the reference comment and how the issue is explained in the description relates.

>  the stream reference it attempts to map to the video element is no longer valid

What does it mean that it is no longer valid? Not sure that is clear, and if it is what I'm thinking, (the SDK is sending a new stream when someone joins?) then I don't think this is the right solution.

Additionally, can you discuss with the team before making refactoring changes like this?
While I don't think this is the right solution for what _I think_ is the issue, if we do refactor the streams into their own component, I much prefer that we have a `Video` component as suggested in the referenced issue, than `RemoteDisplayMedia`, as `Video` would be more reusable.

mwegman (approved):

lalli-flores:
Solution to issue replaced by https://github.com/webex/components/pull/129

PR #130, opened 2020-09-24
docs(WebexMeeting): add in meeting story

adamweeks:
The interstitial meeting wasn't rendering properly either, it was showing in meeting.


src/components/WebexMeeting/WebexMeeting.stories.js
```diff
@@ -12,17 +12,25 @@ const wrapperStyles = {
...
   meetingDestination: 'meeting8',
 };
 
-// TODO: Add story started while "in meeting"
+export const WaitingForOthers = Template.bind({});
+WaitingForOthers.args = {
+  meetingDestination: 'meeting4',
+};
+
+export const Interstitial = Template.bind({});
```
lalli-flores, line 25: Can you maintain the order of the stories, at least for "interstitial"? The first story is usually taken as the "default" state of a component. For `WebexMeeting`, starting at the interstitial state should be the default

adamweeks: Gotcha! Basically a flow of how the component would proceed: Loading, Interstitial, Waiting, In Meeting?

lalli-flores: So the first item is the default and it's used to showcase the [component in the docs](https://webex.github.io/components/storybook/?path=/docs/meetings-webex-meeting--interstitial). I think the "default/showcase" state for this component should be interstitial, as that's the first useful state (the first interactive state).
From there on, yes, we should have them flow as how the component would flow. If you see the docs, that's also the order in which the "other" states are showcased in the docs

adamweeks:
@lalli-flores fixed

lalli-flores (approved):
🙌🏼

PR #132, opened 2020-09-28
Storybook Build Assets

lalli-flores:
Even though the image assets are hosted:
image
Storybook is still not able to find it, as it seems to be looking for them within the `storybook` directory only.
image

Added avatar image assets as part of the Storybook build, so now they are copied into the storybook directory. Updated paths accordingly.

Other changes:
* Push Storybook changes back to Github

Following steps in https://support.circleci.com/hc/en-us/articles/360018860473-How-to-push-a-commit-back-to-the-same-repository-as-part-of-the-CircleCI-job for push configuration.


.circleci/config.yml
```diff
@@ -64,13 +64,21 @@ jobs:
     executor: node-erbium-executor
     steps:
       - checkout_from_workspace
+      - add_ssh_keys:
+        fingerprints:
```
adamweeks, line 68: You're getting a build error I think around this line

lalli-flores: Yeah, investigating why the config error, since this is [how the docs describe it](https://circleci.com/docs/2.0/configuration-reference/#add_ssh_keys)

adamweeks (approved):

lalli-flores:
Let's hope this one finally works 🤞🏼

PR #133, opened 2020-09-28
Contributing Doc Updates

lalli-flores:


CONTRIBUTING.md
```diff
@@ -17,23 +17,38 @@ If you would like to contribute to this repository by adding features, enhanceme
...
+    2. Wait for Github to finish forking the repository
+    3. Head over to your newly created repository fork
+
+2. Clone your repository fork locally
+
+    ```bash
+    git clone git@github.com:{your username}/components.git
+    ```
+
+3. Install all project dependencies
```
adamweeks, line 44: Does `npm-install-peers` do a full install as well?

lalli-flores: Yes, install peers + others

adamweeks: Could we change this to 
```suggestion
3. Install all project dependencies and peers
```
to avoid confusion?

adamweeks (approved):
One minor comment, but non blocking. Thanks for updating this, it will be helpful for other contributors!

CONTRIBUTING.md
lalli-flores: Updated

PR #137, opened 2020-11-10
Handle state when user leaves a meeting

lalli-flores:
Changing the flow when a user leaves a meeting. Instead of attempting to reset the meeting, we simply let the user know that they have left the meeting.
If they wish to rejoin the meeting, they can restart the meeting flow. This aligns with changes in the SDK where once a user leaves a meeting, a new meetingID is given, and simplifies the flow.


adamweeks (approved):
Thanks for changing this, it should help with the workflow of the meeting. We might need to investigate the auto destroy functionality in the meetings sdk, but this is a good solution until then. 

I'm not a big fan of null vs undefined, perhaps we can revisit this in the future and get more meeting details in the `useMeetingDestination` hook, but this solution works.

LGTM

PR #151, opened 2021-01-14
Checkout repository before running storybook step

lalli-flores:
Checking before running Storybook does two things:
- Bring the CI branch with changes from release
- Adds github.com to ~/.ssh/known_host so we can push back changes to Github


taymoork2 (approved):
LGTM

PR #152, opened 2021-01-14
ci: attach workspace before checkout in storybook step

lalli-flores:
This allows for node_modules to be present when running this step


taymoork2 (approved):
FISI

PR #154, opened 2021-01-19
build(bump): allowing newest momentum ui for node bump

peter7cole:
Missing the caret on the momentum-ui/react dependency, which is breaking a bump in Cisco Online Account Management Portal. Thank you!

### DIFF
--- a/package.json
+++ b/package.json
@@ -32,7 +32,7 @@
   ],
   "peerDependencies": {
     "@babel/runtime": "^7.11.2",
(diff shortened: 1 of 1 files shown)


lalli-flores (changes requested):
Momentum UI is a peer dependency of the components, so we should not include it as part of the dependencies.

taymoork2:
@peter7cole you can squash everything into one commit and rebase to solve the package-lock conflict

lalli-flores (approved):

PR #158, opened 2021-03-02
feat(WebexMember): display icon when member is muted

alinasuciu:
Please give us a feedback for all the changes we added or an overall feedback
https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-154242

### DIFF
--- a/src/adapters/MembershipJSONAdapter.test.js
+++ b/src/adapters/MembershipJSONAdapter.test.js
@@ -1,0 +1,46 @@
+import {isObservable} from 'rxjs';
+
+import JSONData from '../data/memberships';
+import MembershipJSONAdapter from './MembershipJSONAdapter';
+
+describe('Membership JSON Adapter', () => {
+  const destinationID = 'room1';
+  const destinationType = 'room';
+  const membershipID = 'membership1';
+  let membershipJSONAdapter;
+  let testMembership;
+
+  beforeEach(() => {
+    membershipJSONAdapter = new MembershipJSONAdapter(JSONData);
+    testMembership = JSONData[membershipID];
+  });
+
+  afterEach(() => {
+    membershipJSONAdapter = null;
+    testMembership = null;
+  });
+
+  describe('getMembersFromDestination()', () => {
+    test('returns an observable', () => {
+      expect(isObservable(membershipJSONAdapter.getMembersFromDestination())).toBeTruthy();
+    });
+
+    test('emits a Membership object on subscription', (done) => {
+      membershipJSONAdapter.getMembersFromDestination(destinationID, destinationType)
+        .subscribe((membership) => {
+          expect(membership).toMatchObject(testMembership);
+          done();
+        });
+    });
+
+    test('throws error on invalid destinationID', (done) => {
+      membershipJSONAdapter.getMembersFromDestination('invalid').subscribe(
+        () => { },
+        (error) => {
+          expect(error.message).toBe('Could not find members for destination "invalid"');
+          done();
+        },
+      );
+    });
+  });
+});
--- a/src/components/WebexMember/WebexMember.jsx
+++ b/src/components/WebexMember/WebexMember.jsx
@@ -1,30 +1,45 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 
-import {usePerson} from '../hooks';
+import {usePerson, useMemberships} from '../hooks';
 import WebexAvatar from '../WebexAvatar/WebexAvatar';
 import {WEBEX_COMPONENTS_CLASS_PREFIX} from '../../constants';
 
 /**
  * Displays a webex meeting member.
  *
+ * @param {string} props.destinationType Type of destination of the membership roster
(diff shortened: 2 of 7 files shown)


src/components/WebexMember/WebexMember.jsx
```diff
@@ -13,22 +13,25 @@ import {WEBEX_COMPONENTS_CLASS_PREFIX} from '../../constants';
  *
  * @returns {object} JSX of the component
  */
-export default function WebexMember({personID, displayStatus}) {
+export default function WebexMember({personID, displayStatus, isMuted}) {
```
lalli-flores, line 16: Passing `isMuted` as a prop to `WebexMember` does not align with the component system design.
In the component system, each component should be as "self contained" as possible, requiring minimal props. In the design, each component is passed an ID of the resource it presents, and with that ID the component (through 
 a hook that wires in the adapter) fetches the data it needs and listens to changes on that resource to update its UI accordingly.
In this case, instead of passing the mute status as a prop, you need to use the membership hook to get that information about this member.

src/components/WebexMember/WebexMember.stories.js
```diff
@@ -18,3 +18,9 @@ StatusEnabled.args = {
   personID: 'user1',
   displayStatus: true,
 };
+
+export const MicrofonMuted = Template.bind({});
```
lalli-flores, line 22: Simplifying the name:
```suggestion
export const Muted = Template.bind({});
```

src/adapters/MembershipJSONAdapter.test.js
```diff
@@ -0,0 +1,56 @@
+import memberships from '../data/memberships';
```
lalli-flores, line 1: Thanks for adding missing tests! 🙂
I think you followed the format on the people adapter tests, which is a great start, but those tests where written very early on. Using the [`MeetingsJSONAdapter` tests](https://github.com/webex/components/blob/master/src/adapters/MeetingsJSONAdapter.test.js) is a better guide of where we want to take the tests, in terms of what we test and how we word our test descriptions

lalli-flores (changes requested):
Thank you for putting out a draft PR so quickly! Let me know if any of my feedback is not clear.

On your commits, try to squash them if they are making changes for the same feature. One commit == one feature change, as the commits translate to our changelog items. In terms of structure, we usually add a scope to denote where the change is happening, for instance, this change would be `feat(WebexMember):` as we are adding a new feature to the `WebexMember` component. This way, the change description can be shorter as you don't have to specify in that statement where the change happened (we already know from the scope)

src/components/WebexMember/WebexMember.jsx
```diff
@@ -10,23 +10,38 @@ import {WEBEX_COMPONENTS_CLASS_PREFIX} from '../../constants';
...
  */
-export default function WebexMember({personID, displayStatus}) {
+export default function WebexMember({
+  personID,
+  displayStatus,
+  destinationID,
+  destinationType,
+}) {
   const {firstName, lastName} = usePerson(personID);
+  const memberships = useMemberships(destinationID, destinationType);
```
cipak, line 25: actually, `members`, not `memberships`. The hook should also be renamed to `useMembers`, but maybe not part of this PR.

src/components/WebexMember/WebexMember.scss
```diff
@@ -10,4 +11,12 @@
     text-overflow: ellipsis;
     padding-left: 0.25rem;
   }
+
+  .member-muted {
+    flex: 1;
+    text-align: right;
+    .icon {
+      color: #ff0000;
```
cipak, line 19: color can be moved up a level

lalli-flores (commented):
@alinasuciu, can you make sure that the [Storybook stories](https://github.com/webex/components/blob/master/src/components/WebexMember/WebexMember.stories.js) are updated appropriately? 
You will also need to add a new story to showcase a participant muted

src/components/WebexMember/WebexMember.stories.js
```diff
@@ -18,3 +18,10 @@ StatusEnabled.args = {
   personID: 'user1',
   displayStatus: true,
 };
+
+export const MemberMuted = Template.bind({});
```
lalli-flores, line 22:
```suggestion
export const Muted = Template.bind({});
```
Naming this story `muted` should be sufficient, it will display under the `WebexMember` group under Storybook

lalli-flores (commented):
Minor comment, but otherwise good to go!

src/components/WebexMember/__snapshots__/WebexMember.stories.storyshot
```diff
@@ -1,5 +1,115 @@
 // Jest Snapshot v1, https://goo.gl/fbAQLP
 
+exports[`Storyshots Platform/Webex Member Member Muted 1`] = `
```
lalli-flores, line 3: Don't forget to run update snapshot to remove stale snapshots.
```
npm run test -- --updateSnapshot

lalli-flores (approved):

PR #159, opened 2021-03-04
refactor(tests): use local timezone instead of UTC (on windows)

cipak:


jest.global.setup.js
```diff
@@ -1,3 +1,3 @@
 module.exports = async () => {
-  process.env.TZ = 'UTC';
+  process.env.TZ = 'UTC'; // this does not affect Date objects on windows https://stackoverflow.com/a/56482581/210550
```
lalli-flores, line 2: We shouldn't have OS-specific code. If this doesn't work on Windows and it's not helpful in other operating systems, we should just remove it. Make sure to update Jest settings in `package.json` as well

jest.setup.js
```diff
@@ -1,7 +1,8 @@
 const MockDate = require('mockdate');
 
 // Fix "now" date-time for tests
-MockDate.set('August 1, 2020 10:00:00 UTC');
+// use a local timezone date representation so that tests run on windows
```
lalli-flores, line 4: No need to specify this (would this comment make sense to someone that is part of our discussions?), maybe just updating the existing comment should suffice.
```
// Fix "now" date-time for tests but keeping local timezone

lalli-flores (commented):
Minor feedback, otherwise thanks for the find!

lalli-flores (approved):

PR #161, opened 2021-03-10
feat(WebexMeeting): show participant roster from meetings component

alinasuciu:


lalli-flores:
@alinasuciu, I apologize since you put work into this, but the roster should not be shown in the _Meetings component_ but rather in the [Meetings widget](https://github.com/webex/widgets/tree/master/src/widgets/WebexMeeting). I will close this PR and update Jira appropriately.

PR #162, opened 2021-03-11
feat(WebexMemberRoster): separate in meeting and not in meeting users

maloun96:


src/components/WebexMemberRoster/WebexMemberRoster.jsx
```diff
@@ -24,11 +24,37 @@ import useMemberships from '../hooks/useMemberships';
  */
 export default function WebexMemberRoster({destinationID, destinationType}) {
   const memberships = useMemberships(destinationID, destinationType);
-  const members = memberships.map(
+
+  const getMembersNode = (data) => (data || []).map(
```
cipak, line 28: as this function returns a list of nodes, maybe rename it to `getMembers` or `renderMembers`, and also rename `data` to `members`.

src/components/WebexMemberRoster/WebexMemberRoster.jsx
```diff
@@ -24,11 +24,37 @@ import useMemberships from '../hooks/useMemberships';
...
     (member) => <WebexMember key={member.personID} personID={member.personID} />,
   );
 
-  return <div className={`${WEBEX_COMPONENTS_CLASS_PREFIX}-roster`}>{members}</div>;
+  if (destinationType !== 'meeting') {
+    return <div className={`${WEBEX_COMPONENTS_CLASS_PREFIX}-roster`}>{getMembersNode(memberships)}</div>;
+  }
+
+  const groupBy = (items, key) => items.reduce(
```
cipak, line 36: we should put this function in a separate util library or use lodash.

src/components/WebexMemberRoster/WebexMemberRoster.jsx
```diff
@@ -24,11 +24,37 @@ import useMemberships from '../hooks/useMemberships';
...
+      ],
+    }),
+    {},
+  );
+
+  const groupedMembers = groupBy(memberships, 'isInMeeting');
+
+  return (
+    <div className={`${WEBEX_COMPONENTS_CLASS_PREFIX}-roster-container`}>
+      <div className={`${WEBEX_COMPONENTS_CLASS_PREFIX}-roster-separator`}>In Meeting</div>
```
cipak, line 51: more like a section-title than a separator.

src/components/WebexMemberRoster/WebexMemberRoster.jsx
```diff
@@ -24,11 +24,37 @@ import useMemberships from '../hooks/useMemberships';
...
+      <div className={`${WEBEX_COMPONENTS_CLASS_PREFIX}-roster`}>{getMembersNode(groupedMembers.true)}</div>
+
+      <div className={`${WEBEX_COMPONENTS_CLASS_PREFIX}-roster-separator`}>Not in Meeting</div>
```
cipak, line 54: I guess this section should not be displayed at all if there are no not-in-meeting members.

src/data/memberships.js
```diff
@@ -4,30 +4,30 @@ export default {
     destinationID: 'room1',
     destinationType: 'room',
     members: [
-      {personID: 'user1'},
-      {personID: 'user2'},
-      {personID: 'user3'},
-      {personID: 'user4'},
+      {personID: 'user1', isInMeeting: true},
```
cipak, line 7: actually, `isInMeeting` only makes sense for meetings, so let's only add this field for meetings. And it would be nice to have at least one meeting where all the members are active (in meeting).

src/components/WebexMemberRoster/WebexMemberRoster.jsx
```diff
@@ -24,11 +25,26 @@ import useMemberships from '../hooks/useMemberships';
...
+  const renderMembers = (members) => (members || []).map(
     (member) => <WebexMember key={member.personID} personID={member.personID} />,
   );
 
-  return <div className={`${WEBEX_COMPONENTS_CLASS_PREFIX}-roster`}>{members}</div>;
+  if (destinationType !== 'meeting') {
+    return <div className={`${WEBEX_COMPONENTS_CLASS_PREFIX}-roster`}>{renderMembers(memberships)}</div>;
+  }
+
+  const groupedMembers = _.groupBy(memberships, (membership) => membership.isInMeeting);
```
taymoork2, line 37: I'd rather not add a new dep without prior discussion, especially lodash because it's heavy and it looks like there isn't any extra work being done to treeshake it either.
I (personally) do not like Lodash and would rather use a native alternative or just copy/paste the groupBy code directly from lodash source

https://github.com/you-dont-need/You-Dont-Need-Lodash-Underscore#_groupby

taymoork2 (dismissed):

src/components/WebexMemberRoster/WebexMemberRoster.jsx
```diff
@@ -24,11 +24,31 @@ import useMemberships from '../hooks/useMemberships';
  */
 export default function WebexMemberRoster({destinationID, destinationType}) {
   const memberships = useMemberships(destinationID, destinationType);
```
cipak, line 26: let's call this members instead of memberships, they are actually members. We might also rename useMemberships to useMembers in the future.

src/components/WebexMemberRoster/WebexMemberRoster.jsx
```diff
@@ -23,12 +23,32 @@ import useMemberships from '../hooks/useMemberships';
...
-  return <div className={`${WEBEX_COMPONENTS_CLASS_PREFIX}-roster`}>{members}</div>;
+  if (destinationType !== 'meeting') {
+    return <div className={`${WEBEX_COMPONENTS_CLASS_PREFIX}-roster`}>{renderMembers(members)}</div>;
+  }
+
+  const isInMeetingMembers = members.filter(membership => membership.isInMeeting);
+  const isNotInMeetingMembers = members.filter(membership => !membership.isInMeeting);
+
+  return (
+    <div className={`${WEBEX_COMPONENTS_CLASS_PREFIX}-roster-container`}>
```
cipak, line 40: I've noticed that the inner classes of a component are not prefixed in other components, so you can go with:
- `${WEBEX_COMPONENTS_CLASS_PREFIX}-roster-container` as the parent (main) class of this component
- section-title
- section-content (or section-body)

src/components/WebexMemberRoster/WebexMemberRoster.jsx
```diff
@@ -23,12 +23,32 @@ import useMemberships from '../hooks/useMemberships';
...
+  const renderMembers = (members) => (members || []).map(
     (member) => <WebexMember key={member.personID} personID={member.personID} />,
   );
 
-  return <div className={`${WEBEX_COMPONENTS_CLASS_PREFIX}-roster`}>{members}</div>;
+  if (destinationType !== 'meeting') {
+    return <div className={`${WEBEX_COMPONENTS_CLASS_PREFIX}-roster`}>{renderMembers(members)}</div>;
+  }
+
+  const isInMeetingMembers = members.filter(membership => membership.isInMeeting);
```
cipak, line 36: let's use `inMeetingMembers` and `notInMeetingMembers` (without the `is` prefix) so that they don't look like booleans. And the current item in filter should be called `member`.

maloun96:
Done!

src/data/memberships.js
```diff
@@ -25,48 +25,48 @@ export default {
     destinationID: 'meeting1',
     destinationType: 'meeting',
     members: [
-      {personID: 'user1'},
-      {personID: 'user2'},
-      {personID: 'user3'},
+      {personID: 'user1', isInMeeting: false},
```
lalli-flores, line 28: Please make sure that you update the [interface/documentation](https://github.com/webex/component-adapter-interfaces/blob/master/src/MembershipsAdapter.js#L8) before making changes to the data model in other repos

src/components/WebexMemberRoster/WebexMemberRoster.jsx
lalli-flores: @cipak Please avoid adding lodash at all costs [see @taymoork2 comment](https://github.com/webex/components/pull/162#discussion_r592740350). In fact, please discuss with the team any changes to the project's dependencies

src/data/memberships.js
lalli-flores: @cipak, member objects must be consistent, so all properties should exists and have a value

src/components/WebexMemberRoster/WebexMemberRoster.jsx
lalli-flores: @cipak, make sure that all changes to the interface are first done on the [interface repo](https://github.com/webex/component-adapter-interfaces)

src/components/WebexMemberRoster/WebexMemberRoster.jsx
cipak: Yes, we saw his comment, and I agree such changes need to be discussed first. My comment was more like an invitation for comments, I'll be more clear next time. The current implementation of separating in meeting from not in meeting members uses just the standard array filter().

src/data/memberships.js
cipak: ok, but a "room member" and a "meeting member" are like different subclasses of a common "member" class. From this point of view, can they have different/specific fields? `muted` and `isInMeeting` don't make much sense for members of a room. If we must give them a value, though, I would use `undefined` or `null` (expressing a N/A) instead of `false`, what do you think?

src/components/WebexMemberRoster/WebexMemberRoster.jsx
cipak: sure, we don't really need lodash anyway.

src/components/_components.scss
```diff
@@ -7,4 +7,5 @@
 @import 'WebexMeetingControl/WebexMeetingControls';
 @import 'WebexMeetingInfo/WebexMeetingInfo';
 @import 'WebexMember/WebexMember';
+@import 'WebexMemberRoster/WebexMemberRoster';
```
lalli-flores, line 10: Thanks for keeping this alphabetical! 🙂

src/components/WebexMemberRoster/WebexMemberRoster.scss
```diff
@@ -0,0 +1,7 @@
+.#{$WEBEX_COMPONENTS_CLASS_PREFIX}-roster-container {
```
lalli-flores, line 1: For consistency with other components, the top-level CSS class name should match the component name
```suggestion
.#{$WEBEX_COMPONENTS_CLASS_PREFIX}-member-roster {
```

src/components/WebexMemberRoster/WebexMemberRoster.jsx
cipak: ok, we'll create a PR for adding the `isInMeeting` field to the `member` interface. Other than that, the `useMemberships` hook  already returns a list of members, I was just trying to reflect that in this variable's name.

src/components/WebexMemberRoster/WebexMemberRoster.scss
```diff
@@ -0,0 +1,7 @@
...
+  .section-title {
+    @include body-fonts;
+    color: $md-gray-40;
+    border-bottom: 1px solid $md-gray-30;
```
lalli-flores, line 5: Please use `rem` as CSS units

src/components/WebexMemberRoster/WebexMemberRoster.jsx
```diff
@@ -23,12 +23,32 @@ import useMemberships from '../hooks/useMemberships';
...
-  const members = memberships.map(
+  const members = useMemberships(destinationID, destinationType);
+
+  const renderMembers = (members) => (members || []).map(
     (member) => <WebexMember key={member.personID} personID={member.personID} />,
   );
 
-  return <div className={`${WEBEX_COMPONENTS_CLASS_PREFIX}-roster`}>{members}</div>;
+  if (destinationType !== 'meeting') {
+    return <div className={`${WEBEX_COMPONENTS_CLASS_PREFIX}-roster`}>{renderMembers(members)}</div>;
```
lalli-flores, line 33: Let's keep a single exit flow control (e.g. only one return statement per function)

src/data/memberships.js
cipak: all right

src/components/WebexMemberRoster/WebexMemberRoster.jsx
```diff
@@ -23,12 +23,32 @@ import useMemberships from '../hooks/useMemberships';
...
+  }
+
+
+  const inMeetingMembers = members.filter(member => member.isInMeeting);
+  const notInMeetingMembers = members.filter(member => !member.isInMeeting);
+
+  return (
```
lalli-flores, line 40: The problem with this approach, as demonstrated in the description video, is that while there are no members loaded, the user sees the shrunk headers first. We should display all or nothing, maybe a spinner while we gather the data.

src/components/WebexMemberRoster/WebexMemberRoster.jsx
```diff
@@ -23,12 +23,32 @@ import useMemberships from '../hooks/useMemberships';
...
+    <div className={`${WEBEX_COMPONENTS_CLASS_PREFIX}-roster`}>
+      <div className="section-title">In Meeting</div>
```
lalli-flores, line 42: shouldn't this be heading tags?

lalli-flores (changes requested):
Thanks for opening a PR for this!

Left some feedback that I hope gets addressed. Let me know if something is not clear.
You are also missing the Storybook update. Since you're adding a new capability to the component, it should be showcased in Storybook. Adding the story will also autogenerate some snapshot tests

src/components/WebexMemberRoster/WebexMemberRoster.jsx
lalli-flores: I mistook this line as a change to the adapter's method. In this case, you're right, we should update this hook as `useMembers` to reflect its return values 👍🏼

Please make this change in a separate commit `refactor(hooks): ...`

src/data/memberships.js
lalli-flores: I think for now having this subclassing will just complicate things and we should keep it as one member object. I think `false` will make it easier to use when checking for values, but I'm ok with it using `null` if you want to denote a difference.

FYI, in general for this codebase, we should use `null` over `undefined` IF we are assigning values to denote that the lack of value is intentional. `undefined` is ok, if we have never assigned a value.

src/components/WebexMemberRoster/WebexMemberRoster.jsx
```diff
@@ -23,12 +23,47 @@ import useMemberships from '../hooks/useMemberships';
...
+  const inMeetingFilter = (member) => member.inMeeting;
+  const notInMeetingFilter = (member) => !member.inMeeting;
+
+  const renderMembers = (members) => members.map(
     (member) => <WebexMember key={member.personID} personID={member.personID} />,
   );
 
-  return <div className={`${WEBEX_COMPONENTS_CLASS_PREFIX}-roster`}>{members}</div>;
+  const getMembers = () => {
+    if (destinationType !== 'meeting') {
```
lalli-flores, line 35: Please use [the enum](https://github.com/webex/component-adapter-interfaces/blob/master/src/MembershipsAdapter.js#L31) (by importing it) instead of hardcoding strings

lalli-flores (commented):
Code looks good!
You're still missing a couple things:
* Please add a new story to Storybook titled `Meeting`
* Something is off with the commits, there should only be 2 (there are 4 now). Please make sure that each commit has the corresponding changes according to their descriptions

src/components/WebexMemberRoster/WebexMemberRoster.scss
```diff
@@ -0,0 +1,7 @@
+.#{$WEBEX_COMPONENTS_CLASS_PREFIX}-member-roster {
+  .section-title {
+    @include body-fonts;
+    color: $md-gray-40;
+    border-bottom: 0.1rem solid $md-gray-30;
```
cipak, line 5: @lalli-flores there are some issues when using rems for border widths, is this a hard requirement? https://bugs.chromium.org/p/chromium/issues/detail?id=382483

cipak: actually this links to an old (and fixed) bug, but still, as the browsers can only render integer widths, doesn't it make more sense to specify the border as 1px, instead of 0.0625rem? (which is basically converting 1px in rems)

If 1rem==31px, the border is still rendered one pixel wide (in chrome), but if 1rem==32px, the border will be 2 pixels, is this the intended behavior?

lalli-flores: The intended behavior is for all things to scale accordingly. I would prefer to maintain rem units across CSS, less we run into issues.

lalli-flores:
@maloun96 something's gone wrong with the latest push, from your 2 commits one of them shouldn't be there and there are changes to the snapshot tests that shouldn't happen. Changes to a different component should also be placed in a separate commit, this way they'll be 2 entries in the changelog, one for each component.

Did you remove the hook refactor out of the PR? It's ok if it'll be separate but I just want to understand what to expect.

Lastly, sorry if I'm not being clear, but you should add a [new story titled `Meeting`](https://github.com/webex/components/blob/master/src/components/WebexMemberRoster/WebexMemberRoster.stories.js) that showcases the labels, as it would show if the roster was used in a meeting.

src/components/WebexRemoteMedia/WebexRemoteMedia.jsx
```diff
@@ -35,8 +35,8 @@ export default function WebexRemoteMedia({className, meetingID}) {
   const videoRef = useStream(remoteVideo);
   const shareRef = useStream(remoteShare);
   const [remoteMediaRef, {width}] = useElementDimensions();
-
-  const hasOtherMembers = members.length > 1;
+
```
lalli-flores, line 38: Watch out for unintended whitespace characters

lalli-flores (commented):
Code changes look good!
Just update your commit messages and we should be good to go

`feat(membershipssdkadapter): show in meeting and not in meeting members` => `feat(MemberRoster): show in meeting and not in meeting members`

`feat(WebexRemoteMedia): get inMeeting members. Update data to match s...` =>
`feat(WebexRemoteMedia): handle inMeeting members

Update data to match stories and snapshots.`

^^ By adding a secondary comment after a new line character, you create a [commit message body](https://github.com/webex/components/blob/master/CONTRIBUTING.md#body), instead of cutting off the commit header half-sentence

lalli-flores:
@maloun96, actually do you mind creating a new PR for this and link it as a follow up?
For some reason the checks on this PR are not running at all.

PR #163, opened 2021-03-18
docs(Storybook): dropdown for destinationID and destinationType on WebexMemberRoster

BenjaminGolya:
Changed input into dropdown for destinationID and destinationType in storybook, WebexMemberRoster component.

### DIFF
--- a/src/components/WebexMemberRoster/WebexMemberRoster.stories.js
+++ b/src/components/WebexMemberRoster/WebexMemberRoster.stories.js
@@ -1,9 +1,27 @@
 import React from 'react';
+import {DestinationType} from '@webex/component-adapter-interfaces';
 import WebexMemberRoster from './WebexMemberRoster';
+import Memberships from '../../data/memberships';
+
+const destinationIDs = Object.values(Memberships).map((membership) => membership.destinationID);
 
 export default {
   title: 'Platform/Webex Member Roster',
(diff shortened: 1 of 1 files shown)


lalli-flores:
@BenjaminGolya, what's the context of this change? I'm ok with with making the destination type a dropdown but not the destination ID, since we control the `destinationType` enum.
Also, make sure this commit change is of type `docs(Storybook)`, as it is impacting only the Storybook (docs) usage and should not trigger a package release.

src/components/WebexMemberRoster/WebexMemberRoster.stories.js
```diff
@@ -1,9 +1,27 @@
...
 import WebexMemberRoster from './WebexMemberRoster';
+import Memberships from '../../data/memberships';
+
+const destinationIDs = Object.values(Memberships).map((membership) => membership.destinationID);
 
 export default {
   title: 'Platform/Webex Member Roster',
   component: WebexMemberRoster,
+  argTypes: {
+    destinationID: {
```
lalli-flores, line 12: This should be an open field, let's not make the `destinationID` a dropdown

lalli-flores:
@BenjaminGolya, @cipak and I had a discussion on this and we agreed that instead of making all controls into dropdowns, we should just remove the controls panel from Storybook altogether instead.

PR #164, opened 2021-03-18
refactor(MembershipJsonAdapter): change membership to list of members

CernaianuMihai:


lalli-flores (commented):
Code looks good! Would you mind updating your commits though?

* Make sure the commit author matches your Github user, so they can be linked to your profile. You can do this by updating `user.name` in your local git repo to match your Github username and `user.email` to match the email you used for your Github account
* Create two separate commits: one for the adapter changes (commit comment is good) and one for the hook-related changes, so it'd be another `refactor(hooks):...`

src/adapters/MembershipJSONAdapter.js
```diff
@@ -44,13 +44,13 @@ import {Observable} from 'rxjs';
  */
 export default class MembershipJSONAdapter extends MembershipsAdapter {
   /**
-   * Returns an observable that emits membership data for the first destination
+   * Returns an observable that emits a list of members for the first destination
    * found that matches the given ID and type.
-   * For this implementation, once the data is emitted, the observable completes.
+   * For this implementation, once the members is emitted, the observable completes.
```
cipak, line 49: member list / list of members

src/components/hooks/useMembers.js
```diff
@@ -34,8 +34,8 @@ export default function useMemberships(destinationID, destinationType) {
       // eslint-disable-next-line no-console
       console.error(error.message);
     };
-    const onMembers = (data) => {
-      setMembers([...data.members]);
+    const onMembers = (membersList) => {
```
cipak, line 37: just `members` is fine

lalli-flores: @cipak, I would like to keep the `onMember` function as this is the only way the component can guarantee the state stays immutable. The component cannot know if the adapter implementation has mutated the original member array or created a new one, so we just make sure it's always a new one before updating the state. We ran into this issue in the past and it was tricky to find. Although we fixed in in the SDK adapter, we can prevent others extending the adapter from shooting themselves in the foot.

cipak: @lalli-flores I see. Yes, immutability helps with such issues. I see that all hooks use this approach, and provide copies instead of the original data received from the sdk. However, these copies are shallow, so someone could still break something if they modify one of the elements in this array. I guess it's a trade-off, deep copies would be too expensive, right?

lalli-flores: @cipak, correct, deep copies are too expensive. Shallow copies are ok in this case, because we only want to make sure that React understand this is a new state by giving it a new array object (memory address), even if the contents of the array are still pointing to the previous objects.

lalli-flores (approved):

lalli-flores:
@CernaianuMihai seems like test are breaking. Can you please check?

CernaianuMihai:
Hello @lalli-flores, I see that you approved this PR and all of the tests are running fine now. Should I merge it or wait for you to do it?

PR #177, opened 2021-03-24
feat(MemberRoster): show in meeting and not in meeting members

maloun96:
https://user-images.githubusercontent.com/8277264/110772174-0066a700-8264-11eb-9902-e15f644eda4c.mov


lalli-flores (approved):
Everything looks really good! Clean implementation 🙂
Glad to see your checks running as well 👍🏼

I feel like we're going to have to come back to the member roster and add a spinner while we fetch members. It might not be noticeable with the JSON data, but when we have calls over the network, it might take longer than a user will tolerate without some sort of feedback.

PR #179, opened 2021-03-25
feat(WebexMemberRoster): indicate current user if is meeting

maloun96:
https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-220758

https://user-images.githubusercontent.com/8277264/112468341-ccfe3f00-8d70-11eb-970d-c09da8e3d3ff.mov

### DIFF
--- a/src/components/WebexMember/WebexMember.jsx
+++ b/src/components/WebexMember/WebexMember.jsx
@@ -12,6 +12,7 @@ import {WEBEX_COMPONENTS_CLASS_PREFIX} from '../../constants';
  * @param {string} props.destinationID  ID of the destination for which to get members
  * @param {string} props.personID  ID of the person for which to display avatar
  * @param {boolean} props.displayStatus  Whether or not to display the user's status
+ * @param {boolean} props.showMe Indicate who current use id
  *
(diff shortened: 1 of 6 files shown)


src/components/WebexMember/WebexMember.jsx
```diff
@@ -20,6 +21,7 @@ export default function WebexMember({
   destinationID,
   personID,
   displayStatus,
+  showMe,
```
lalli-flores, line 24: This should not be pass as a prop but inferred by the data that is already in `WebexMember`

src/components/WebexMemberRoster/WebexMemberRoster.jsx
```diff
@@ -32,6 +34,7 @@ export default function WebexMemberRoster({destinationID, destinationType}) {
         destinationID={destinationID}
         personID={personID}
         key={personID}
+        showMe={ID === personID && destinationType === DestinationType.MEETING}
```
lalli-flores, line 37: This should probably be moved inside `WebexMember`

lalli-flores (changes requested):

PR #180, opened 2021-03-26
fix(WebexMember): make component icon proportional with text

alinasuciu:
The member component takes the parent width, with a bigger the icon alignment is more visible. Check the roster component

### DIFF
--- a/src/components/WebexMember/WebexMember.jsx
+++ b/src/components/WebexMember/WebexMember.jsx
@@ -32,7 +32,7 @@ export default function WebexMember({
     <div className={`${WEBEX_COMPONENTS_CLASS_PREFIX}-member`}>
       <WebexAvatar personID={personID} displayStatus={displayStatus} />
       <div className="member-name">{`${firstName} ${lastName}`}</div>
-      {isMuted && <div className="member-muted"><span className="icon icon-microphone-muted_24" /></div>}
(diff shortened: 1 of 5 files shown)


src/components/WebexMember/WebexMember.jsx
```diff
@@ -32,7 +32,7 @@ export default function WebexMember({
     <div className={`${WEBEX_COMPONENTS_CLASS_PREFIX}-member`}>
       <WebexAvatar personID={personID} displayStatus={displayStatus} />
       <div className="member-name">{`${firstName} ${lastName}`}</div>
-      {isMuted && <div className="member-muted"><span className="icon icon-microphone-muted_24" /></div>}
+      {isMuted && <div className="member-muted"><span className="icon icon-microphone-muted_16" /></div>}
```
maloun96, line 35: Can we use `<Icon />` component instead of span icon?

src/components/WebexMemberRoster/WebexMemberRoster.scss
```diff
@@ -1,4 +1,5 @@
 .#{$WEBEX_COMPONENTS_CLASS_PREFIX}-member-roster {
+  width: 20rem;
```
maloun96, line 2: I don't think we should define a width for roster

PR #181, opened 2021-03-26
feat(WebexMember): indicate current user if in meeting

maloun96:
https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-220758

https://user-images.githubusercontent.com/8277264/112468341-ccfe3f00-8d70-11eb-970d-c09da8e3d3ff.mov

### DIFF
--- a/src/components/WebexMember/WebexMember.jsx
+++ b/src/components/WebexMember/WebexMember.jsx
@@ -1,8 +1,8 @@
 import React from 'react';
 import PropTypes from 'prop-types';
+import {DestinationType} from '@webex/component-adapter-interfaces';
 import {Icon} from '@momentum-ui/react';
-
-import {usePerson, useMembers} from '../hooks';
+import {usePerson, useMembers, useMe} from '../hooks';
 import WebexAvatar from '../WebexAvatar/WebexAvatar';
 import {WEBEX_COMPONENTS_CLASS_PREFIX} from '../../constants';
 
@@ -23,18 +23,21 @@ export default function WebexMember({
   displayStatus,
 }) {
   const {firstName, lastName} = usePerson(personID);
+  const me = useMe();
   const members = useMembers(destinationID, destinationType);
   const member = members
     .find((itemMember) => itemMember.personID === personID);
 
   const isMuted = member && member.muted;
   const isSharing = member && member.sharing;
+  const showMe = me.ID === personID && destinationType === DestinationType.MEETING;
 
   return (
     <div className={`${WEBEX_COMPONENTS_CLASS_PREFIX}-member`}>
       <WebexAvatar personID={personID} displayStatus={displayStatus} />
       <div className="details">
         <div className="name">{`${firstName} ${lastName}`}</div>
+        {showMe && <div className="subtext">You</div>}
         {isSharing && <div className="presenter">Presenter</div>}
       </div>
       {isSharing && <Icon name="icon-content-share_16" className="sharing" />}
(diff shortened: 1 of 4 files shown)


src/components/WebexMember/WebexMember.jsx
```diff
@@ -31,7 +32,10 @@ export default function WebexMember({
   return (
     <div className={`${WEBEX_COMPONENTS_CLASS_PREFIX}-member`}>
       <WebexAvatar personID={personID} displayStatus={displayStatus} />
-      <div className="member-name">{`${firstName} ${lastName}`}</div>
+      <div className="member-name">
```
cipak, line 35: let's take out the subtitle from the `member-name` class, something like:
```html
<div.member>
    <avatar/>
    <div.description>
        <div.name/>
        <div.subtitle/>
    </div>
</div>
```
Feel free to use a better name than `description` if you find one.

cipak:
@maloun96 please use `WebexMember` instead of `WebexMemberRoster` in the commit message

src/components/WebexMember/WebexMember.jsx
```diff
@@ -22,6 +22,7 @@ export default function WebexMember({
   displayStatus,
 }) {
   const {firstName, lastName} = usePerson(personID);
+  const {ID} = useMe();
```
cipak, line 25: to align with other PRs on this file and minimize the merge conflicts, please use a `me` variable:
```js
const me = useMe();
```

src/components/WebexMember/WebexMember.scss
```diff
@@ -12,6 +12,12 @@
     padding-left: 0.25rem;
   }
 
+  .member-subtitle {
```
cipak, line 15: let's use `subtext`, like in the story, and no `member-` prefix, we are in the member component context anyway

src/components/WebexMember/WebexMember.jsx
```diff
@@ -22,16 +22,21 @@ export default function WebexMember({
...
     .find((itemMember) => itemMember.personID === personID);
 
   const isMuted = member && member.muted;
+  const showMe = me.ID === personID && destinationType === DestinationType.MEETING;
 
   return (
     <div className={`${WEBEX_COMPONENTS_CLASS_PREFIX}-member`}>
       <WebexAvatar personID={personID} displayStatus={displayStatus} />
-      <div className="member-name">{`${firstName} ${lastName}`}</div>
+      <div className="description">
```
lalli-flores, line 36: Is this `description` div necessary?
Can the subtext just be part of the `member-name` div?

lalli-flores (commented):
Minor comment but otherwise almost there 👍🏼

src/components/WebexMember/WebexMember.jsx
maloun96: We agree with the team to use this structure, we consider that member-name should contain the only the name, after that will come 'representor', 'you', 'organization name'

lalli-flores (approved):

src/components/WebexMember/WebexMember.jsx
lalli-flores: Ok, we can keep the structure if the team thinks that's best. I'm not a fan of the class name, not sure it makes sense to me why it's called `description`? Maybe a better name would be `details`

lalli-flores:
@maloun96, if you can fix the conflicts, we should be good to merge

maloun96:
@lalli-flores Done!

lalli-flores:
I'm sorry seems like conflicts are back 😭

PR #183, opened 2021-03-26
fix(WebexMember): make muted icon proportional with text

alinasuciu:


src/components/WebexMember/WebexMember.jsx
```diff
@@ -32,7 +33,7 @@ export default function WebexMember({
     <div className={`${WEBEX_COMPONENTS_CLASS_PREFIX}-member`}>
       <WebexAvatar personID={personID} displayStatus={displayStatus} />
       <div className="member-name">{`${firstName} ${lastName}`}</div>
-      {isMuted && <div className="member-muted"><span className="icon icon-microphone-muted_24" /></div>}
+      {isMuted && <div className="member-muted"><Icon name="icon-microphone-muted_16" /></div>}
```
lalli-flores, line 36: Since you're using Momentum Icon component now, does it make sense to keep the `member-muted` div?

alinasuciu: I didn't remove the div for now in order to prevent future merge conflicts, I also added some extra code for the same reason.
This PR: https://github.com/webex/components/pull/190/files will remove the wrapper div and it also contains some of the changes I added

lalli-flores (approved):

PR #186, opened 2021-03-30
feat(OrganizationsJsonAdapter): implement the organizationsJSONAdapter

CernaianuMihai:
Implement of OrganizationsJSONAdapter as per OrganizationsAdapter interface, create tests for OrganizationsJSONAdapter and create mock data for organizations.

https://jira-eng-gpk2.cisco.com/jira/secure/RapidBoard.jspa?rapidView=6063&view=detail&selectedIssue=SPARK-220422

PR #189, opened 2021-03-31
feat(hooks): add useOrganization hook

BenjaminGolya:
useOrganization hook get an orgID and return an organization name.
depends on https://github.com/webex/components/pull/186

https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-220420


src/components/hooks/useOrganization.js
```diff
@@ -0,0 +1,36 @@
...
+export default function useOrganization(ID) {
+  const [organization, setOrganization] = useState({});
+  const {organizationsAdapter} = useContext(AdapterContext);
+
+  useEffect(() => {
+    const onError = (error) => {
+      // eslint-disable-next-line no-console
+      console.error(error.message);
+    };
+    const subscription = organizationsAdapter.getOrg(ID).subscribe(setOrganization, onError);
```
lalli-flores, line 27: In order to avoid potential state issues in the future, add a callback that creates a shallow copy for the state update, like [we do in the meetings hook](https://github.com/webex/components/blob/master/src/components/hooks/useMeeting.js#L39). As stated in the comment of the meetings hook, React doesn't recognize changes to the attributes of an object state, only a new object state will trigger a state change. By creating a shallow copy before updating the state, we can assure that any emissions from the adapter will for sure trigger a state change

```suggestion
    const onOrganization = (newOrganization) => {
      setOrganization({...newOrganization});
    };
    
    const subscription = organizationsAdapter.getOrg(ID).subscribe(onOrganization, onError);
```

lalli-flores (approved):

PR #190, opened 2021-03-31
feat(WebexMember): display icon when member is sharing

lungancatalin:
Update WebexMember component to display content share icon
https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-220740


src/components/WebexMember/WebexMember.scss
```diff
@@ -12,9 +18,17 @@
...
     color: #ff0000;
   }
+
+  .sharing {
+    padding-right: 2.5rem;
+    line-height: 1.5rem;  
+  }
+
+  .presenter {
+    padding-left: 0.25rem;
```
cipak, line 32: let's move the padding to the `.descripton` div, would this make sense?

lungancatalin: yes, that makes sense. I moved the property to the description class and also turned it into the left margin. ;)

src/data/memberships.js
```diff
@@ -91,13 +159,23 @@ export default {
...
-        personID: 'user2', muted: true, inMeeting: true, orgID: 'org1',
+        personID: 'user2',
+        muted: true,
+        inMeeting: true,
+        sharing: false,
+        orgID: 'org1',
       },
       {
-        personID: 'user3', inMeeting: false, orgID: 'org1',
+        personID: 'user3',
```
cipak, line 176: could you please add `muted: false` for this one so that all members have all fields.

and, only if you're willing, can you put the fields for each member in the same order as the one in the interface:
- personID
- orgID
- muted
- sharing
- inMeeting
https://github.com/webex/component-adapter-interfaces/blob/master/src/MembershipsAdapter.js#L8

thank you!

lungancatalin: yes, I will added the muted property to false for user3 and also, I will arrange them correctly according to the interface.

lalli-flores (approved):

lalli-flores:
@lungancatalin, seems that there are conflicts but once resolved, we should be good to go!

lungancatalin:
> @lungancatalin, seems that there are conflicts but once resolved, we should be good to go!

Hi @lalli-flores , I resolved all of the conflicts and now my commit is ready to be merged. Thanks.

PR #192, opened 2021-04-05
refactor(WebexMember): rename property 'personID' to 'id'

maloun96:
https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-222996


lungancatalin:
I think that you should rename your commit to refactor(WebexMember) instead of feat(WebexMember). :)

lalli-flores (approved):
Looks good! If you could fix conflicts we should be good to merge

PR #193, opened 2021-04-06
feat(WebexMeeting): set meetingID as component prop

alinasuciu:
Update meeting component to use meetingID instead of meetingDestination
https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-223897

### DIFF
--- a/src/components/WebexMeeting/WebexMeeting.jsx
+++ b/src/components/WebexMeeting/WebexMeeting.jsx
@@ -8,17 +8,17 @@ import WebexInterstitialMeeting from '../WebexInterstitialMeeting/WebexInterstit
 import WebexMeetingControl from '../WebexMeetingControl/WebexMeetingControl';
 import WebexMeetingControls from '../WebexMeetingControl/WebexMeetingControls';
 import {WEBEX_COMPONENTS_CLASS_PREFIX} from '../../constants';
-import {useMeetingDestination} from '../hooks';
+import {useMeeting} from '../hooks';
 
 /**
  * Webex Meeting component displays the default Webex meeting experience.
  *
- * @param {string} props.meetingDestination  ID of the virtual meeting location
- * @param {Array} [props.controls]  Array of control names to display
- * @returns {Object} JSX of the component
+ * @param {string} props.meetingID  ID of the meeting
+ * @param {array} [props.controls]  Array of control names to display
+ * @returns {object} JSX of the component
  */
-export default function WebexMeeting({meetingDestination, controls}) {
-  const {ID, remoteVideo} = useMeetingDestination(meetingDestination);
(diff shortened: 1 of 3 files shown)


src/components/WebexMeeting/WebexMeeting.jsx
```diff
@@ -72,8 +72,7 @@ WebexMeeting.defaultProps = {
    * @param {boolean} isActive  Whether or not the meeting is active
    */
   // eslint-disable-next-line no-confusing-arrow
-  controls: (isActive) =>
-    isActive
-      ? ['mute-audio', 'mute-video', 'share-screen', 'leave-meeting']
-      : ['mute-audio', 'mute-video', 'join-meeting'],
+  controls: (isActive) => isActive
```
cipak, line 75: let's not change this if the only change is code formatting

src/components/WebexMeeting/WebexMeeting.jsx
```diff
@@ -8,17 +8,17 @@ import WebexInterstitialMeeting from '../WebexInterstitialMeeting/WebexInterstit
...
 import WebexMeetingControls from '../WebexMeetingControl/WebexMeetingControls';
 import {WEBEX_COMPONENTS_CLASS_PREFIX} from '../../constants';
-import {useMeetingDestination} from '../hooks';
+import {useMeeting} from '../hooks';
 
 /**
  * Webex Meeting component displays the default Webex meeting experience.
  *
- * @param {string} props.meetingDestination  ID of the virtual meeting location
+ * @param {string} props.meetingID  ID of the meeting
```
cipak, line 16: the double space was there before, I know, but a single space should be sufficient. Or maybe a dash to make it more readable https://jsdoc.app/tags-param.html

for now I would go with a single space like in the rest of the project.

alinasuciu: all the files of this kind have double space between the parameter name and description

cipak: ok missed that

lalli-flores (changes requested):
Can you update the [WebexMeeting README](https://github.com/webex/components/blob/e03f64bd572bc10a4b46e8b4379064bf9b5aae60/src/components/WebexMeeting/README.md#embed) with the prop change?

Also, this would technically be a breaking change because a meeting ID is a different  type of value from receiving a meeting destination. While from the JSON data perspective these look the same, with the SDK adapter they are totally different. Anyone out there using `WebexMeeting` should be made aware of the change via a (generated) version bump.

With that said, I don't want to bump the major version because this is technically a "beta" product, so a minor bump using `feat` in the commit instead of `refactor` should do

lalli-flores (approved):
Looks good! 👍🏼

PR #194, opened 2021-04-08
fix(WebexMeetingControl): update screen sharing icon

lungancatalin:
Update the WebexMeetingControl component to use the correct sharing icon. (share-screen-presence-stroke)
Storybook: https://user-images.githubusercontent.com/15773210/114026315-c696b980-987e-11eb-975b-8f1ad9d1850d.png
Link bug in Jira: https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-223771
Subtask for this repo: https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-224251


src/components/WebexMeetingControl/WebexMeetingControl.jsx
```diff
@@ -24,9 +24,11 @@ export default function WebexMeetingControl({type}) {
   );
 
   if (icon) {
+    const size = (icon === 'share-screen-presence-stroke') ? '26' : '28';
```
cipak, line 27: @lalli-flores can we ask momentum for a 28px size for this icon? All the other icons have this size available, while this one only has 12, 14, 26, 30 as @lungancatalin discovered.

Alternatively, I would put complete icon names (with sizes) in the control definitions.

cipak: or add the `_28` suffix only if the last character of the icon name is not already a digit... :)

lalli-flores: Can you try the `size` prop?

lungancatalin: Hi @lalli-flores, Finally, I have used the size prop in Icon but it was necessary to update all of the icons classes with the _28 and _26 suffix. After that, in the Icon I have added the size={28} and all of the icon sizes will be aligned to the 28px as font size. Also, I have made a new PR on sdk-component-adapter to update all of these classes.

cancel_28
microphone-muted_28
camera-muted_28
share-screen-presence-stroke_26
participant-list_28

lalli-flores (approved):

src/components/WebexMeetingControl/WebexMeetingControl.jsx
lalli-flores: Weird that you had to suffix all the names, but it seems that Momentum does that behind the scenes anyway.

PR #195, opened 2021-04-08
feat(WebexMember): display if user is host

karinasigartau0798:
Updated WebexMember to display if user is the host of a meeting.

https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-222973


src/components/WebexMember/WebexMember.scss
```diff
@@ -16,7 +16,12 @@
     text-overflow: ellipsis;
   }
 
-  .subtext {
+  .organization {
```
cipak, line 19: `.subtext` needs to stay (used for "You"), `.presenter` needs to go

cipak: Ok I see now that `.subtext` is still present... please remove `.presenter`

src/components/WebexMember/WebexMember.jsx
```diff
@@ -22,23 +23,32 @@ export default function WebexMember({
   personID,
   displayStatus,
 }) {
-  const {firstName, lastName} = usePerson(personID);
+  const {firstName, lastName, orgID} = usePerson(personID);
   const me = useMe();
   const members = useMembers(destinationID, destinationType);
   const member = members
     .find((itemMember) => itemMember.ID === personID);
+  const organization = useOrganization(orgID);
```
lalli-flores, line 31: Seems like there are some changes from external user story mingled in here?
Can you make sure the changes only relate to "host"?

src/data/memberships.js
```diff
@@ -73,20 +80,31 @@ export default {
...
       {
         ID: 'user3',
         orgID: 'org1',
         muted: true,
         sharing: false,
         inMeeting: true,
+        host: true,
+      },
+      {
+        ID: 'user5',
```
lalli-flores, line 102: I think this change is related to external org users and should be part of a different PR

lalli-flores (changes requested):
The host/ roles changes look good, but there are changes related to external org users that shouldn't be part of this PR. If you could remove those, then we should be good to go

src/components/WebexMember/WebexMember.jsx
karinasigartau0798: Done.

src/data/memberships.js
karinasigartau0798: You're right.
Done.

lalli-flores (approved):
Awesome! You even indirectly fixed a UX bug from having "You" and "presenter" roles in different lines 🥳

PR #196, opened 2021-04-09
feat(WebexMeeting): add an overlay banner on active local share

alinasuciu:
When local share is active a banner will be displayed on the bottom of the meeting pointing out the fact that you are currently sharing

https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-224478


lalli-flores (changes requested):
@alinasuciu, I think this change should go in the [`WebexInMeeting` component](https://github.com/webex/components/blob/master/src/components/WebexInMeeting/WebexInMeeting.jsx) instead of the Webex Meeting component. In-meeting component is meant to handle all states while users are in a meeting.

cipak:
> @alinasuciu, I think this change should go in the [`WebexInMeeting` component](https://github.com/webex/components/blob/master/src/components/WebexInMeeting/WebexInMeeting.jsx) instead of the Webex Meeting component. In-meeting component is meant to handle all states while users are in a meeting.

The problem is that the banner is displayed currently relative to the controls (on top of the controls), and the controls are displayed in the `Meeting` component (not `InMeeting`). As discussed a while ago, we could remove the controls from the `Meeting` component and let the widget add them. This way, the sharing banner could be displayed at the bottom of the `InMeeting` component.

This banner could also be thought of as an extra "highlight" of the sharing button itself, so it could also be part of the `Controls` component, simplifying the display logic.

lalli-flores:
@cipak, I see. Logically, it doesn't make sense for it to be in `WebexMeeting` when we have `WebexInMeeting` to handle in-meeting experience. What we can do is instead of displaying the banner on top of the controls, we can display it on the top portion of the remote video instead. This way you don't need to know anything about the controls

image

lalli-flores (approved):
Thank you for accommodating the changes @alinasuciu 🙌🏼

PR #197, opened 2021-04-12
feat(WebexMember): display if user is guest

BenjaminGolya:
Display if user is guest in meeting. Updated WebexMember component.

https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-222985

guest

### DIFF
--- a/src/components/WebexMember/WebexMember.jsx
+++ b/src/components/WebexMember/WebexMember.jsx
@@ -35,7 +35,8 @@ export default function WebexMember({
   const isSharing = member && member.sharing;
   const showMe = me.ID === personID && destinationType === DestinationType.MEETING;
   const isHost = member && member.host;
-  
+  const isGuest = member && member.guest;
+
   const roles = [
     showMe && 'You',
     isHost && 'Host',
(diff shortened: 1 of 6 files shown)


src/components/WebexMember/WebexMember.jsx
```diff
@@ -44,7 +45,10 @@ export default function WebexMember({
     <div className={`${WEBEX_COMPONENTS_CLASS_PREFIX}-member`}>
       <WebexAvatar personID={personID} displayStatus={displayStatus} />
       <div className="details">
-        <div className="name">{`${firstName} ${lastName}`}</div>
+        <div className="name">
+          {`${firstName} ${lastName}`} 
+          {isGuest ? <span className="guest"> (Guest)</span> : ''}
```
cipak, line 50: you can use the simpler syntax `isGuest && <span>...</span>`

src/components/WebexMember/WebexMember.stories.js
```diff
@@ -34,6 +34,13 @@ Host.args = {
   personID: 'user4',
 };
 
+export const Guest = Template.bind({});
+Guest.args = {
+  destinationType: 'meeting',
+  destinationID: 'meeting2',
+  personID: 'user5',
```
lalli-flores, line 41: Maybe use `user6` from this same meeting? [You've already update the data like that anyway](https://github.com/webex/components/pull/197/files#diff-202791f48524a3e98537bfa5bb91dbf5e6e849322b0b5abda6faafeea9f8e089R170).
The only concern I have with using `user5` for this, is that `user5` is also the external user, and I don't want developers confusing "guest" vs "external"

lalli-flores (commented):
Code looks good, minor comment on the Story. Seems like you got conflicts as well; if you can get those fixed we should be good to go

lalli-flores (approved):

PR #198, opened 2021-04-12
feat(WebexLocalMedia): replace self-view with local share

alinasuciu:
When the user is sharing the screen the self-view is replaced by the screen share

https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-224481

### DIFF
--- a/src/components/WebexLocalMedia/README.md
+++ b/src/components/WebexLocalMedia/README.md
@@ -1,6 +1,6 @@
 # Webex Local Media Component
 
-Webex local media component displays the user local stream, if enabled. If user's local media is disabled the avatar of the current _Webex_ user is displayed.
+Webex local media component displays the local video or the local screen share, if enabled. If disabled, the avatar of the current _Webex_ user is displayed. The user can choose which stream to display (video or screen) using the `mediaType` prop. 
 
 <p align="center">
   <img src="./WebexLocalMedia.png" alt="Default Webex Local Media" />
@@ -9,7 +9,7 @@ Webex local media component displays the user local stream, if enabled. If user'
 ## Preview
 
 To see all the different possible states of the Webex Local Media component,
-you can view our [storybook](https://webex.github.io/components/storybook/?path=/story/webex-local-media--enabled)
+you can view our [storybook](https://webex.github.io/components/storybook/?path=/story/webex-local-media--video)
 or run the following **NPM** command:
 
 ```shell
@@ -24,13 +24,13 @@ or run the following **NPM** command:
     const jsonAdapter = new WebexJSONAdapter(jsonData);
     ```
 
-2. Create a component instance by passing the person ID as a string and enclose it within
+2. Create a component instance by passing the meeting ID as a string and enclose it within
 [a data provider](../WebexDataProvider/WebexDataProvider.js) that takes
-the [component data adapter](../../adapters/WebexJSONAdapter.js) that we created previously
-
+the [component data adapter](../../adapters/WebexJSONAdapter.js) that we created previously.
+Choose the type of media to display by passing a string to `mediaType`
     ```js
     <WebexDataProvider adapter={jsonAdapter}>
-      <WebexLocalMedia meetingID="meetingID" />
+      <WebexLocalMedia meetingID="meetingID" mediaType="video|screen" />
     </WebexDataProvider>
     ```
 
--- a/src/components/WebexLocalMedia/WebexLocalMedia.jsx
+++ b/src/components/WebexLocalMedia/WebexLocalMedia.jsx
@@ -14,29 +14,43 @@ import {
 } from '../hooks';
 
 /**
- * Webex Local Media component displays the user's local video.
+ * Webex Local Media component displays the user's local video or local share.
  *
  * @param {string} props.meetingID  ID of the meeting from which to obtain local media
  * @returns {Object} JSX of the component
  */
-export default function WebexLocalMedia({className, meetingID}) {
+export default function WebexLocalMedia({className, meetingID, mediaType}) {
   const [mediaRef, {width}] = useElementDimensions();
-  const {localVideo} = useMeeting(meetingID);
+  const {localVideo, localShare} = useMeeting(meetingID);
+  let stream;
+
+  switch (mediaType) {
+    case 'video':
+      stream = localVideo;
+      break;
+    case 'screen':
+      stream = localShare;
+      break;
+    default:
+      break;
+  }
+
+  const ref = useStream(stream);
   const {ID} = useMe();
-  const videoRef = useStream(localVideo);
 
   const classBaseName = `${WEBEX_COMPONENTS_CLASS_PREFIX}-local-media`;
   const mainClasses = {
     [classBaseName]: true,
     [`${classBaseName}-desktop`]: width >= PHONE_LARGE,
-    [`${classBaseName}-no-media`]: localVideo === null,
+    [`${classBaseName}-no-media`]: !stream,
     [className]: !!className,
   };
+
   const disabledVideo = ID ? <WebexAvatar personID={ID} displayStatus={false} /> : <Spinner />;
 
   return (
     <div ref={mediaRef} className={classNames(mainClasses)}>
-      {localVideo ? <video ref={videoRef} playsInline autoPlay /> : disabledVideo}
+      {stream ? <video ref={ref} playsInline autoPlay /> : disabledVideo}
     </div>
   );
 }
@@ -44,8 +58,10 @@ export default function WebexLocalMedia({className, meetingID}) {
 WebexLocalMedia.propTypes = {
   className: PropTypes.string,
   meetingID: PropTypes.string.isRequired,
+  mediaType: PropTypes.oneOf(['video', 'screen']),
 };
 
 WebexLocalMedia.defaultProps = {
   className: '',
+  mediaType: 'video',
 };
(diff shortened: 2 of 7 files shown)


src/components/WebexLocalMedia/WebexLocalMedia.jsx
```diff
@@ -14,16 +14,18 @@ import {
 } from '../hooks';
 
 /**
- * Webex Local Media component displays the user's local video.
+ * Webex Local Media component displays the user's local video or local share.
  *
```
cipak, line 18: We should also add an explanatory text of the display logic, eg:

```
If the meeting is active and the current user is sharing the screen, the local share is displayed; otherwise, the local 
video is displayed (if video is enabled).
```

src/components/WebexLocalMedia/WebexLocalMedia.jsx
```diff
@@ -14,16 +14,20 @@ import {
...
+ * the local share is displayed; otherwise, the local video is displayed (if video is enabled).
  *
  * @param {string} props.meetingID  ID of the meeting from which to obtain local media
  * @returns {Object} JSX of the component
  */
 export default function WebexLocalMedia({className, meetingID}) {
   const [mediaRef, {width}] = useElementDimensions();
-  const {localVideo} = useMeeting(meetingID);
+  const {localVideo, localShare, remoteVideo} = useMeeting(meetingID);
+  const isActive = remoteVideo !== null;
```
lalli-flores, line 27: @cipak, @alinasuciu don't forget to add a subtask to update this when refactoring meeting state

src/components/WebexLocalMedia/WebexLocalMedia.jsx
```diff
@@ -14,16 +14,20 @@ import {
...
   const {ID} = useMe();
-  const videoRef = useStream(localVideo);
+  const localStream = (isActive && localShare) || localVideo;
```
lalli-flores, line 29: I think a more scalable solution for this is to move the logic of deciding what type of local media to show to [In-meeting component](https://github.com/webex/components/blob/master/src/components/WebexInMeeting/WebexInMeeting.jsx), as this component decides what the experience should be when you're in a meeting.
I'm thinking that maybe we can add a prop to local media `mediaType` where an user can pass `video`, `audio` or `screen` and that's how we can determine the type of local media to show

lalli-flores (changes requested):

src/components/WebexLocalMedia/WebexLocalMedia.jsx
cipak: Initially, we wanted to have two separate components `LocalVideo` and `LocalShare` so that `InMeeting` could choose one of the 2. However, we then saw that `RemoteMedia` displays both share and video, and has some logic of which to display where. So we updated this `LocalMedia` component in a similar way. I still think that, for now, it would be ok to let this component take care of the logic and maybe move it up when the need arises.

lalli-flores: You're right in that `WebexRemoteMedia` has some meeting logic in it, that looking back, shouldn't be there, it should be in `WebexInMeeting` instead. I don't want to make the same mistake with local media, as I foresee some near-team use cases where users just want to get a hold of media display only. Putting the logic there now would basically diminish the usefulness of having it as a component.

src/components/WebexLocalMedia/WebexLocalMedia.jsx
alinasuciu: This is not needed anymore as we moved the logic in InMeeting component.

src/components/WebexLocalMedia/README.md
```diff
@@ -1,38 +1,38 @@
...
+Webex local media component displays the local video or the local screen share, if enabled. If disabled, the avatar of the current _Webex_ user is displayed. The user can choose which stream to display (video or screen) using the `mediaType` prop. 
+
+<p align="center">
+  <img src="./WebexLocalMedia.png" alt="Default Webex Local Media" />
+</p>
+
+## Preview
+
+To see all the different possible states of the Webex Local Media component,
+you can view our [storybook](https://webex.github.io/components/storybook/?path=/story/webex-local-video--enabled)
```
lalli-flores, line 12:
```suggestion
you can view our [storybook](https://webex.github.io/components/storybook/?path=/story/meetings-webex-local-media--video)

```

src/components/WebexLocalMedia/README.md
```diff
@@ -1,38 +1,38 @@
...
+
+1. Create a component adapter from which the data will be retrieved (See [adapters](../../adapters)). For instance:
+
+    ```js
+    const jsonAdapter = new WebexJSONAdapter(jsonData);
+    ```
+
+2. Create a component instance by passing the meeting ID as a string and enclose it within
+[a data provider](../WebexDataProvider/WebexDataProvider.js) that takes
+the [component data adapter](../../adapters/WebexJSONAdapter.js) that we created previously
```
lalli-flores, line 29:
```suggestion
the [component data adapter](../../adapters/WebexJSONAdapter.js) that we created previously. Choose the type of media to display by passing a string to `mediaType`

```

src/components/WebexLocalMedia/WebexLocalMedia.jsx
```diff
@@ -14,38 +14,54 @@ import {
...
-      {localVideo ? <video ref={videoRef} playsInline autoPlay /> : disabledVideo}
+      {stream ? <video ref={ref} playsInline autoPlay /> : disabledVideo}
     </div>
   );
 }
 
 WebexLocalMedia.propTypes = {
   className: PropTypes.string,
   meetingID: PropTypes.string.isRequired,
+  mediaType: PropTypes.string,
```
lalli-flores, line 61:
```suggestion
  mediaType: PropTypes. oneOf(['video', 'screen']),
```

src/components/WebexLocalMedia/WebexLocalMedia.stories.js
```diff
@@ -8,14 +8,22 @@ export default {
 
 const Template = (args) => <WebexLocalMedia {...args} />;
 
-export const MediaEnabled = Template.bind({});
-MediaEnabled.args = {
+export const VideoEnabled = Template.bind({});
```
lalli-flores, line 11:
```suggestion
export const Video = Template.bind({});
```

src/components/WebexLocalMedia/WebexLocalMedia.stories.js
```diff
@@ -8,14 +8,22 @@ export default {
...
+VideoEnabled.args = {
   meetingID: 'meeting1',
+  mediaType: 'video',
 };
 
 export const MediaDisabled = Template.bind({});
```
lalli-flores, line 17:
```suggestion
export const NoMedia = Template.bind({});
```

src/components/WebexLocalMedia/WebexLocalMedia.stories.js
```diff
@@ -8,14 +8,22 @@ export default {
...
 MediaDisabled.args = {
   meetingID: 'meeting2',
+  mediaType: 'video',
+};
+
+export const ShareEnabled = Template.bind({});
```
lalli-flores, line 23:
```suggestion
export const ScreenShare = Template.bind({});
```

lalli-flores (changes requested):
Few things to address:
* Make sure that the banner and the local video don't overlap. Local video should come under the banner
* Make sure things look proportional on mobile screens
* Arrange the Storybook stories in this order: video, screen share & no media
* Create one commit where you do the local media refactoring / adding the new prop and then another one to do the self-view changes

src/components/WebexInMeeting/WebexInMeeting.scss
```diff
@@ -16,6 +16,12 @@ $shadow-color: rgba(0, 0, 0, 0.2);
     }
   }
 
+  &.local-sharing {
+    .local-media-in-meeting {
+      top: 2.8rem;
```
lalli-flores, line 21: How does the transition look when switching between local view and share preview?

lalli-flores (commented):
@alinasuciu, code looks good. My only concern now is the style override for local sharing preview in In-meeting component. How does the transition look from self-view to preview? does it jump? Maybe all the previews should be aligned at the same level from the top

alinasuciu:
> @alinasuciu, code looks good. My only concern now is the style override for local sharing preview in In-meeting component. How does the transition look from self-view to preview? does it jump? Maybe all the previews should be aligned at the same level from the top

I added the same top distance not just when the local share is activated. As the banner has the same height on all devices I had to change the distance from the top for all devices to be the same.

lalli-flores:
> I added the same top distance not just when the local share is activated. As the banner has the same height on all devices I had to change the distance from the top for all devices to be the same.

Can you share changes on what that would look like? If the screen is small, it might look weird that the local media is pushed so far down. Is there a way to scale the banner?

Please don't forget to update the Storyshots too since the class `local-sharing` should exist anymore

alinasuciu:
> > I added the same top distance not just when the local share is activated. As the banner has the same height on all devices I had to change the distance from the top for all devices to be the same.
> 
> Can you share changes on what that would look like? If the screen is small, it might look weird that the local media is pushed so far down. Is there a way to scale the banner?
> 
> Please don't forget to update the Storyshots too since the class `local-sharing` should exist anymore

@lalli-flores , I updated the stories, I actually forgot to remove the addition of the class if the local share is active, everything should be ok now. I don't see a way to scale the banner, the font is not too big, neither the padding. Doesn't look bad, I would say but it's up to you. 
Screenshot 2021-04-21 140644
Screenshot 2021-04-21 140728

lalli-flores (approved):

lalli-flores:
@alinasuciu code looks good. One test is failing so once it's fixed this can be merged.

PR #200, opened 2021-04-13
feat(hoc): add a withMeeting HOC for class components

lungancatalin:
Create a new withMeeting Higher Order Component, for class components
Link sub-task Jira: https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-224406

### DIFF
--- a/src/components/hoc/withMeeting.jsx
+++ b/src/components/hoc/withMeeting.jsx
@@ -1,0 +1,69 @@
+import React, {useEffect, useContext, useState} from 'react';
+import {concatMap} from 'rxjs/operators';
+import PropTypes from 'prop-types';
+import {AdapterContext} from '../hooks/contexts';
+
+/**
+ * withMeeting is a higher-order component that takes a component (WrappedComponent)
+ * and returns a new component (WithMeeting). The new component needs to be provided with a meeting destination
+ * on the "meetingDestination" prop. The wrapped component will receive the meeting object on the "meeting" prop.
+ *
+ * Example:
+ *
+ * function OriginalComponent {
+ *  return <div>Meeting id is {props.meeting.ID}</div>;
+ *
+ * }
+ *
+ * const EnhancedComponent = withMeeting(OriginalComponent);
+ *
+ * <EnhancedComponent meetingDestination="<MEETING_DESTINATION>" />
+ *
+ * @param {React.Component} WrappedComponent The component to wrap and enhance
+ * @returns {React.Component} The enhanced component
+ */
+export default function withMeeting(WrappedComponent) {
+  /**
+   * @param props Props to be forwarded to WrappedComponent
+   * @param {string} props.meetingDestination Virtual location where the meeting should take place.
+   * @returns {object} JSX of the component
+   */
+  function WithMeeting(props) {
+    const [meeting, setMeeting] = useState({});
+    const {meetingsAdapter} = useContext(AdapterContext);
+    const {meetingDestination} = props;
+
+    useEffect(() => {
+      let cleanup;
+      if (!meetingDestination) {
+        setMeeting({});
+        cleanup = undefined;
+      } else {
+        const meetingSubscription = meetingsAdapter
+        .createMeeting(meetingDestination)
+        .pipe(concatMap(({ID}) => meetingsAdapter.getMeeting(ID)))
+        .subscribe(
+          (meeting) => setMeeting({...meeting}),
+          (error) => {
+            setMeeting({error});
+            console.log(error);
+          },
+        );
+        cleanup = () => {
+          meetingSubscription.unsubscribe();
+        };
+      }
+      return cleanup;
+    }, [meetingDestination]);
+
+    return (
+      <WrappedComponent meeting={meeting} {...props} />
+    );
+  }
+
+  WithMeeting.propTypes = {
+    meetingDestination: PropTypes.string.isRequired,
+  };
+
+  return WithMeeting;
+}
--- a/src/components/index.js
+++ b/src/components/index.js
@@ -13,3 +13,4 @@ export {default as WebexMember} from './WebexMember/WebexMember';
 export {default as WebexMemberRoster} from './WebexMemberRoster/WebexMemberRoster';
 export {default as WebexRemoteMedia} from './WebexRemoteMedia/WebexRemoteMedia';
 export {default as useMeetingDestination} from './hooks/useMeetingDestination';
+export {default as withMeeting} from './hoc/withMeeting';


cipak:
Let's follow the hooks-style commit messages:

> feat(hoc): add a withMeeting HOC for class components

cipak:
you need to export the new hoc to the outside world, see an example here: https://github.com/webex/components/pull/199/files#diff-c7db7feaa21621735b0b41422b0d1ba0aae29948113c1a3bf63575d63217b10aR15

src/components/hoc/withMeeting.jsx
```diff
@@ -0,0 +1,41 @@
+import React from 'react';
+import {concatMap} from 'rxjs/operators';
+import {AdapterContext} from '../hooks/contexts';
+
+/**
+ * @param MeetingComponent
```
cipak, line 6: let's add a bit more documentation:
- function description
- type for the MeetingComponent parameter (React.Component)
- a short usage example

cipak: ...and return type. See an example here: https://terrestris.github.io/react-geo/docs/@terrestris/react-geo/5.6.2/HigherOrderComponent_VisibleComponent_VisibleComponent.jsx.html

cipak (approved):

src/components/hoc/withMeeting.jsx
```diff
@@ -0,0 +1,69 @@
...
+ * const EnhancedComponent = withMeeting(OriginalComponent);
+ *
+ * <EnhancedComponent meetingDestination="<MEETING_DESTINATION>" />
+ *
+ * @param {React.Component} WrappedComponent The component to wrap and enhance
+ * @returns {React.Component} The wrapped component
+ */
+
+/**
+ * @param WrappedComponent
```
lalli-flores, line 28: What's the purpose of this comment? shouldn't the above comment be the doc for the component?

lalli-flores (changes requested):
Minor comment, code looks good otherwise!
There are missing unit tests for the new component, however

src/components/hoc/withMeeting.jsx
```diff
@@ -0,0 +1,69 @@
...
+          },
+        );
+    }
+
+    componentWillUnmount() {
+      this.meetingSubscription.unsubscribe();
+    }
+
+    render(props) {
+      return <WrappedComponent meeting={this.state.meeting} {...props} />;
```
lalli-flores, line 59: Why do you pass the whole meeting object? wouldn't all you need to pass is the meeting ID?

src/components/hoc/withMeeting.jsx
```diff
@@ -0,0 +1,69 @@
...
+        meeting: {},
+      };
+    }
+
+    componentDidMount() {
+      const {meetingsAdapter} = this.context;
+
+      this.meetingSubscription = meetingsAdapter
+        .createMeeting(this.props.meetingDestination)
+        .pipe(concatMap(({ID}) => meetingsAdapter.getMeeting(ID)))
```
lalli-flores, line 44: Do you really need to call and subscribe to `getMeeting`? Why would the HOC need to keep updated with the meeting object? Each wrapped component will already take care of that

src/components/hoc/withMeeting.jsx
```diff
@@ -0,0 +1,69 @@
...
+ *
+ * @param {React.Component} WrappedComponent The component to wrap and enhance
+ * @returns {React.Component} The wrapped component
+ */
+
+/**
+ * @param WrappedComponent
+ */
+export default function withMeeting(WrappedComponent) {
+  class WithMeeting extends React.Component {
```
lalli-flores, line 31: Even though the usage for this is to use it with class components, does the HOC have to use class style? could it be a function component like the rest of the code base?

src/components/hoc/withMeeting.jsx
lungancatalin: Hi @lalli-flores, here is my bad with this comment. I will remove that line and remain only the good comments. Thank you.

src/components/hoc/withMeeting.jsx
lungancatalin: Hi @lalli-flores, I think yes. I do that because this is similar with the Hooks and this make sense. Another reason is that, in most cases, the enhanced component will need to have meeting status and that's why is required to make this call here.

src/components/hoc/withMeeting.jsx
lungancatalin: Hi, @lalli-flores I need to have the whole meeting object because the enhanced component will need the meeting status.

src/components/hoc/withMeeting.jsx
lungancatalin: Hi @lalli-flores Yes, it is possible to create a functional component instead of class component. I will make an amend commit with the functional component. What's your opinion?

src/components/hoc/withMeeting.jsx
```diff
@@ -0,0 +1,66 @@
...
+   * @param {string} props.meetingDestination Virtual location where the meeting should take place.
+   * @param props All the props that can be send to the WrappedComponent
+   * @returns {object} JSX of the component
+   */
+  function WithMeeting(props) {
+    const [meeting, setMeeting] = useState({});
+    const {meetingsAdapter} = useContext(AdapterContext);
+
+    useEffect(() => {
+      if (!props.meetingDestination) {
```
cipak, line 36: to have a single return for this useEffect callback, I would suggest the following approach:
```js
let cleanup;
if (!props.meetingDestination) {
  setMeeting({});
  cleanup = undefined;
} else {
  const meetingSubscription = ...
  ...
  cleanup = () => {
    meetingSubscription.unsubscribe();
  };
}

return cleanup;
```

src/components/hoc/withMeeting.jsx
lalli-flores: I would prefer that @lungancatalin, to keep the code base consistent. Thanks!

lungancatalin: Hi @lalli-flores, the commit with functional component is already done. Please have a look on this commit -> https://github.com/webex/components/commit/1418217c480c9da90d0738f2f2cacd4ced920f8c
Regards

src/components/hoc/withMeeting.jsx
lalli-flores: This I don't understand. Won't each wrapped component have access to the whole meeting object because it will do a subscription to `getMeeting` internally?

src/components/hoc/withMeeting.jsx
lalli-flores: Reposting my reply:

> This I don't understand. Won't each wrapped component have access to the whole meeting object because it will do a subscription to getMeeting internally?

cipak: Hm, maybe we're looking at different approaches here. One approach would be to wrap each component with this HOC, like `WebexMeeting`, `WebexMemberRoster` etc. However, the original intention was to wrap the webex meeting widget itself with the HOC (and likewise a dev-defined component that uses these webex components). This way, the widget (or the custom app) would have easy access to an up-to-date meeting object, from which they can extract:
- the meeting id, to provide it down to the webex components
- the meeting state, to decide how to display the components. For example, the widget could display a spinner while the meeting is being created (no meeting id yet), or it could decide to display the roster only after the meeting is JOINED.
- even the meeting controls could be further customized based on meeting state, for example, display an additional control when the user is sharing their screen.
If we just provide the meeting id, the dev (or the widget) would still need to subscribe to the `getMeeting()` adapter function to get meeting updates. And take care of the unsubscribing.
This "extra" subscription should be really cheap, as we are basically subscribing to a `BehaviorSubject`.
For easier access to the meeting id, we could also provide the wrapped component with a `meetingID` property, in addition to the (whole) meeting object.
If you still want us to only provide the meeting id to the wrapped component, we might need to do the same for the `useMeetingDestination` hook, for consistency reasons.

lalli-flores: Thanks for the thorough explanation @cipak, I appreciate the forward thinking!

lalli-flores (approved):

PR #201, opened 2021-04-14
feat(WebexMeeting): implement meeting state

alinasuciu:
Implement meeting state in MeetingJsonAdapter and WebexMeeting component, update tests and meeting mock data

https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-223894

### DIFF
--- a/src/adapters/MeetingsJSONAdapter.js
+++ b/src/adapters/MeetingsJSONAdapter.js
@@ -4,7 +4,7 @@ import {
 import {
   filter, map, takeUntil, tap,
 } from 'rxjs/operators';
-import {MeetingsAdapter, MeetingControlState} from '@webex/component-adapter-interfaces';
+import {MeetingsAdapter, MeetingControlState, MeetingState} from '@webex/component-adapter-interfaces';
 
 // Meeting control names
 export const MUTE_AUDIO_CONTROL = 'mute-audio';
@@ -54,7 +54,8 @@ export const ROSTER_CONTROL = 'member-roster';
  *     "localShare": null,
  *     "remoteAudio": {},
  *     "remoteVideo": {},
- *     "remoteShare": {}
+ *     "remoteShare": {},
+ *     "state": "JOINED"
  *   },
  *   "meeting-2": {
  *     "ID": "meeting-2",
@@ -64,7 +65,8 @@ export const ROSTER_CONTROL = 'member-roster';
  *     "localShare": {},
  *     "remoteAudio": {},
  *     "remoteVideo": {},
- *     "remoteShare": {}
+ *     "remoteShare": {},
+ *     "state": "NOT_JOINED"
  *   }
  * }
  */
@@ -166,8 +168,8 @@ export default class MeetingsJSONAdapter extends MeetingsAdapter {
   getMeeting(ID) {
     const end$ = new Subject();
     const getMeeting$ = Observable.create((observer) => {
-      // A null ID signifies that the meeting is over, in the past or invalid
-      if (ID === null) {
+      // A falsy ID signifies that the meeting was not yet created, or is invalid
+      if (!ID) {
         observer.next({
           ID: null,
           title: null,
@@ -177,6 +179,7 @@ export default class MeetingsJSONAdapter extends MeetingsAdapter {
           remoteAudio: null,
           remoteVideo: null,
           remoteShare: null,
+          state: null,
           showRoster: null,
         });
       } else if (this.datasource[ID]) {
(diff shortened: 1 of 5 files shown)


src/adapters/MeetingsJSONAdapter.test.js
```diff
@@ -166,6 +170,11 @@ describe('Meetings JSON Adapter', () => {
       expect(testMeeting.remoteVideo).toBeInstanceOf(MediaStream);
     });
 
+    test('sets meeting state to JOINED', async () => {
```
cipak, line 173: let's have an expect() before the action, so that we check that the state really has changed. Otherwise, maybe the state was 'JOINED' from the beginning.

src/adapters/MeetingsJSONAdapter.js
```diff
@@ -177,6 +177,7 @@ export default class MeetingsJSONAdapter extends MeetingsAdapter {
           remoteAudio: null,
           remoteVideo: null,
           remoteShare: null,
+          state: MeetingState.LEFT,
```
cipak, line 180: `EMPTY` seems more suited here. The `ID` won't be null anymore, I think, in case of a "left" meeting, because we'll keep the meeting object with its normal id and state=LEFT.

so the check could just read `if(ID)`, and the comment above should read "A falsy ID signifies that the meeting was not yet created, or is invalid"

src/components/hooks/useMeeting.js
```diff
@@ -49,7 +51,7 @@ export default function useMeeting(meetingID) {
       throw error;
     };
     const onComplete = () => {
-      setMeeting(emptyMeeting);
+      setMeeting({...emptyMeeting, state: MeetingState.LEFT});
```
cipak, line 54: the `LEFT` state is already set by the adapter here https://github.com/webex/components/pull/201/files?file-filters%5B%5D=.js#diff-b63aa3e7e2482c57f9d72c2a873dafae6bdaedaa8c04639c9313b922df39ebd9R273

we don't need to do anything in this `onComplete` handler, it can be removed.

src/adapters/MeetingsJSONAdapter.js
```diff
@@ -166,8 +168,8 @@ export default class MeetingsJSONAdapter extends MeetingsAdapter {
   getMeeting(ID) {
     const end$ = new Subject();
     const getMeeting$ = Observable.create((observer) => {
-      // A null ID signifies that the meeting is over, in the past or invalid
-      if (ID === null) {
+      // A falsy ID signifies that the meeting was not yet created, or is invalid"
```
cipak, line 171: extra quote at the end of the line

package.json
```diff
@@ -40,7 +40,7 @@
   },
   "dependencies": {
     "@juggle/resize-observer": "^3.2.0",
-    "@webex/component-adapter-interfaces": "^1.13.0",
+    "@webex/component-adapter-interfaces": "^1.16.0",
```
cipak, line 43: please remove this, it is handled by another PR https://github.com/webex/components/pull/204

src/adapters/MeetingsJSONAdapter.js
```diff
@@ -185,6 +188,7 @@ export default class MeetingsJSONAdapter extends MeetingsAdapter {
         // Add a video stream as if it were a remote meeting
         if (meeting.remoteVideo instanceof MediaStream) {
           meeting.remoteVideo = this.getVideoStream();
+          meeting.state = MeetingState.JOINED;
```
cipak, line 191: this should not be needed, as `joinMeeting()` takes care of it. And already existing meetings in the datasource should have their `state`s set to JOINED from the start.

src/data/meetings.js
```diff
@@ -1,3 +1,14 @@
+import {MeetingState} from '@webex/component-adapter-interfaces';
+
+MeetingState.NOT_JOINED = 'NOT_JOIND';
```
cipak, line 3: small typo

src/components/hooks/useMeetingDestination.js
```diff
@@ -1,5 +1,6 @@
 import {useEffect, useContext, useState} from 'react';
 import {concatMap} from 'rxjs/operators';
+import {MeetingState} from '@webex/component-adapter-interfaces';
```
cipak, line 3: @alinasuciu please revert the changes in this file, they are done by @lungancatalin in a different PR

cipak (approved):

lalli-flores (approved):
We should also update the follow tests:
* [`getMeeting() completes when the meeting is left`](https://github.com/webex/components/blob/0e322a0c19f02d89a2d46d3f212b0572c3b1ada6/src/adapters/MeetingsJSONAdapter.test.js#L129) expect can be updated to check that the meeting state is `LEFT` instead of checking for truthy

PR #205, opened 2021-04-19
refactor(WebexInMeeting): improve sharing demonstration in story

alinasuciu:
Separate local sharing from remote sharing, add video for remote sharing and create separate stories for local share and remote share

https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-226863


src/adapters/MeetingsJSONAdapter.js
```diff
@@ -342,6 +339,26 @@ export default class MeetingsJSONAdapter extends MeetingsAdapter {
...
 
+  /**
+   * Returns a promise to a MediaStream object that captures the contents of a screen share.
+   *
+   * @returns {Promise.<MediaStream>} Media stream that captures display
+   * @private
+   */
+  // eslint-disable-next-line class-methods-use-this
+  getShareStream() {
+    const share = document.getElementById('remote-share');
```
cipak, line 350: let's factor this logic into a `getStreamFromElement(id)` function, and then we can call:
```js
meeting.remoteVideo = getStreamFromElement('remote-video');
meeting.remoteShare = getStreamFromElement('remote-share');
```

.storybook/preview.js
```diff
@@ -21,6 +21,7 @@ export const decorators = [
       </WebexDataProvider>
 
       <video id="remote-video" src="./video/ongoing-meeting.mp4" muted autoPlay playsInline loop width="0" height="0" />
+      <video id="remote-share" src="./video/ongoing-share.mp4" muted autoPlay playsInline loop width="0" height="0" />
```
cipak, line 24: this file was taken from https://www.videezy.com/abstract/50576-a-data-screen-of-numbers

just an example, we should replace it with something more screen-sharing like.

lalli-flores: Is this PR the same as https://github.com/webex/components/pull/208?
I've added a [video here](https://github.com/webex/components/pull/208#pullrequestreview-640554289) that we should use

PR #213, opened 2021-04-26
JSX Linter Fixes

lalli-flores:
Updates ESLint settings so that JSX files are also linted.

### DIFF
--- a/src/components/hoc/withMeeting.jsx
+++ b/src/components/hoc/withMeeting.jsx
@@ -24,8 +24,8 @@ import {AdapterContext} from '../hooks/contexts';
  */
 export default function withMeeting(WrappedComponent) {
   /**
-   * @param props Props to be forwarded to WrappedComponent
-   * @param {string} props.meetingDestination Virtual location where the meeting should take place.
+   * @param {object} props  Data to be forwarded to WrappedComponent
+   * @param {string} props.meetingDestination Virtual location where the meeting should take place
    * @returns {object} JSX of the component
    */
   function WithMeeting(props) {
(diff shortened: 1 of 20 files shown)


src/components/hoc/withMeeting.jsx
```diff
@@ -24,41 +24,16 @@ import {AdapterContext} from '../hooks/contexts';
...
-        };
-      }
-      return cleanup;
-    }, [meetingDestination]);
-
-    return (
-      <WrappedComponent meeting={meeting} {...props} />
-    );
+    // eslint-disable-next-line react/jsx-props-no-spreading
+    return <WrappedComponent meeting={meeting} {...props} />;
```
cipak, line 36: nice catch:)

The only difference between the hook and the HOC is the handling of the "complete" message from `getMeeting`'s observable. The hook puts a meeting with `ID=null` on the line while the HOC does nothing, because a previous message was a meeting with the same id but `state=MeetingState.LEFT`.

Because components now look at the meeting state field, we should refactor the hook so it also adds the LEFT state when a meeting closes. This is to be done in https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-223895.

Another possible difference is that the HOC does not mind an empty meetingDestination, it just sets the meeting to an empty object without calling `createMeeting(dest)` which, I think, throws an error in this case. This will be handled in the task above, too.

lalli-flores: Both HOC and hooks should behave the same way, hopefully we address those discrepancies as part of the refactor 🙃

taymoork2 (approved):
LGTM

PR #217, opened 2021-04-27
ci: add Storybook previews for PRs

alinasuciu:
Install chromatic for storybook preview and implement CI with CircleCI

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -44,6 +44,15 @@ jobs:
       - store_test_results:
           path: test_results
 
+  storybook_preview: 
+    executor: node-fermium-executor
+    steps:
+      - attach_workspace:
+          at: .
+      - run:
+          name: Generate Storybook Preview in Chromatic
+          command: npm run chromatic --project-token=$CHROMATIC_PROJECT_TOKEN --exit-zero-on-changes
+
   build:
     executor: node-fermium-executor
     steps:
@@ -102,6 +111,9 @@ workflows:
       - unit_test:
           requires:
             - install
+      - storybook_preview:
+          requires:
+            - unit_test
       - build:
           requires:
             - lint
--- a/package.json
+++ b/package.json
@@ -15,7 +15,8 @@
     "storybook:build": "build-storybook -s ./docs/assets -o ./docs/storybook",
     "test": "jest",
     "test:watch": "jest --watch",
-    "test:coverage": "jest --coverage"
+    "test:coverage": "jest --coverage",
+    "chromatic": "chromatic"
   },
   "repository": {
     "type": "git",
@@ -63,6 +64,7 @@
     "@storybook/preset-scss": "^1.0.3",
     "@storybook/react": "^6.1.1",
     "babel-loader": "^8.1.0",
+    "chromatic": "^5.6.2",
     "css-loader": "^4.2.2",
     "eslint": "^7.2.0",
     "eslint-config-airbnb": "^18.2.0",
--- a/package-lock.json
+++ b/package-lock.json
@@ -4,6 +4,33 @@
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
+    "@actions/core": {
+      "version": "1.2.7",
+      "resolved": "https://registry.npmjs.org/@actions/core/-/core-1.2.7.tgz",
+      "integrity": "sha512-kzLFD5BgEvq6ubcxdgPbRKGD2Qrgya/5j+wh4LZzqT915I0V3rED+MvjH6NXghbvk1MXknpNNQ3uKjXSEN00Ig==",
+      "dev": true
+    },
+    "@actions/github": {
+      "version": "4.0.0",
+      "resolved": "https://registry.npmjs.org/@actions/github/-/github-4.0.0.tgz",
+      "integrity": "sha512-Ej/Y2E+VV6sR9X7pWL5F3VgEWrABaT292DRqRU6R4hnQjPtC/zD3nagxVdXWiRQvYDh8kHXo7IDmG42eJ/dOMA==",
+      "dev": true,
+      "requires": {
+        "@actions/http-client": "^1.0.8",
+        "@octokit/core": "^3.0.0",
+        "@octokit/plugin-paginate-rest": "^2.2.3",
+        "@octokit/plugin-rest-endpoint-methods": "^4.0.0"
+      }
+    },
+    "@actions/http-client": {
+      "version": "1.0.11",
+      "resolved": "https://registry.npmjs.org/@actions/http-client/-/http-client-1.0.11.tgz",
+      "integrity": "sha512-VRYHGQV1rqnROJqdMvGUbY/Kn8vriQe/F9HR2AlYHzmKuM/p3kjNuXhmdBfcVgsvRWTz5C5XW5xvndZrVBuAYg==",
+      "dev": true,
+      "requires": {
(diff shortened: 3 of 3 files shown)


alinasuciu:
@lalli-flores, while I waited for the aws credentials I managed to create an implementation with chromatic, from my point of view it satisfy our requests and even more, it's free. The implementation requires the following: 

- creation of a chromatic account (on this step we'll receive a token);
- chromatic installation as dev dependency
- saving the chromatic token in repository's secrets
- creating a github secret on repository 
- workflow file for CI

The current implementation creates a new build on each push on a branch and on PR adds a section which displays if  storybook was successfully deployed and published. From here everyone can access the new storybook by clicking on details link from "storybook publish" section([check this example](https://608681d984568d002107ad37-awqnukhrdk.chromatic.com/?path=/story/platform-webex-avatar--active)). It also has some extra features like ui review and tests but these ones require an account for each user who wants to give a review or accept/decline some changes. Check the attachment.

I wanted to do the same with screener.io, to see if it meets our needs but it seems that they don't have a free version so I couldn't create an account. 
Screenshot 2021-04-27 142343

lalli-flores:
@alinasuciu, we can use chromatic for the PR previews. I went ahead and created a project with the team account and got a project token. The only thing to change is to use CircleCI instead of Github actions for this one since we already have a working workflow there. You would need to create a new job `chromatic` in our [CircleCI configuration](https://github.com/webex/components/blob/master/.circleci/config.yml) and a step in the workflow _after_ the `unit_test`.
I've already setup an env variable `CHROMATIC_PROJECT_TOKEN` in CircleCI, so you can assume that exists

alinasuciu:
> @alinasuciu, we can use chromatic for the PR previews. I went ahead and created a project with the team account and got a project token. The only thing to change is to use CircleCI instead of Github actions for this one since we already have a working workflow there. You would need to create a new job `chromatic` in our [CircleCI configuration](https://github.com/webex/components/blob/master/.circleci/config.yml) and a step in the workflow _after_ the `unit_test`.
> I've already setup an env variable `CHROMATIC_PROJECT_TOKEN` in CircleCI, so you can assume that exists

Hi Lalli, I moved the job in circleCI workflows and I also tested on my local repository. 
Screenshot 2021-04-28 172101

package.json
```diff
@@ -15,7 +15,8 @@
     "storybook:build": "build-storybook -s ./docs/assets -o ./docs/storybook",
     "test": "jest",
     "test:watch": "jest --watch",
-    "test:coverage": "jest --coverage"
+    "test:coverage": "jest --coverage",
+    "chromatic": "chromatic --project-token CHROMATIC_PROJECT_TOKEN --exit-zero-on-changes"
```
cipak, line 19: is this env var evaluated? looks more like a static project token to me

cipak: I guess the arguments given in https://github.com/webex/components/pull/217/files#diff-78a8a19706dbd2a4425dd72bdab0502ed7a2cef16365ab7030a5a0588927bf47R54 overwrite the ones here anyway, so maybe it would be better to just remove them from here:
```json
"chromatic": "chromatic"
```
unless I'm missing something:)

cipak (approved):

cipak:
I would go with this commit message: `build(circleci): storybook preview for each PR using chromatic

package.json
```diff
@@ -15,7 +15,8 @@
     "storybook:build": "build-storybook -s ./docs/assets -o ./docs/storybook",
     "test": "jest",
     "test:watch": "jest --watch",
-    "test:coverage": "jest --coverage"
+    "test:coverage": "jest --coverage",
+    "chromatic": "chromatic --project-token"
```
lalli-flores, line 19: What would the flag `--project-token` do without a value?

```suggestion
    "chromatic": "chromatic"
```

.circleci/config.yml
```diff
@@ -44,6 +44,15 @@ jobs:
       - store_test_results:
           path: test_results
 
+  chromatic:
```
lalli-flores, line 47: Let's name this step more generic in case we change tooling in the future
```suggestion
  storybook_preview: 
```

.circleci/config.yml
```diff
@@ -44,6 +44,15 @@ jobs:
       - store_test_results:
           path: test_results
 
+  chromatic: 
+    executor: node-fermium-executor
+    steps:
+    - attach_workspace:
+        at: .
+    - run:
+        name: Chromatic Build and Deploy
```
lalli-flores, line 53:
```suggestion
        name: Generate Storybook Preview in Chromatic
```

.circleci/config.yml
```diff
@@ -44,6 +44,15 @@ jobs:
...
+        command: npm run chromatic --project-token=79723ccb1e7b --exit-zero-on-changes
```
lalli-flores, line 54: When running NPM script with arguments you need to add `--` before passing flags/arguments'. Also, this is where the env variable in CircleCI would come in

```suggestion
        command: npm run chromatic -- --project-token=$CHROMATIC_PROJECT_TOKEN --exit-zero-on-changes
```

.circleci/config.yml
```diff
@@ -102,6 +111,9 @@ workflows:
       - unit_test:
           requires:
             - install
+      - chromatic:
```
lalli-flores, line 114:
```suggestion
      - storybook_preview:
```

.circleci/config.yml
```diff
@@ -102,6 +111,9 @@ workflows:
...
+          requires:
+            - install
```
lalli-flores, line 116: We should do a preview only if the tests pass
```suggestion
            - unit_test
```

lalli-flores (approved):
We're definitely in the right direction, left some comments on things to address.
Also, please update your commit message as `ci: add Storybook previews for PRs`.

With `feat` a new version will be generated, which is not necessary for changes in tooling

package.json
alinasuciu: I guess I forgot to push the changes. I made some tests and ran the the script from command line so I used the token saved in env variables and I missed to delete the flag as well

.circleci/config.yml
alinasuciu: here as well, I added my token for testing and forgot to push the changes back. sorry again

package.json
alinasuciu: This token is taken from environment variables and it helps to trigger the script manually, I guess we don't need it anymore. I added it just for testing as from here I couldn't access the token saved in circleCI

.circleci/config.yml
lalli-flores: You should probably remove the `install` requirement, it will be met by virtue of the unit tests requiring them. I think this is breaking the CI too, as it's not triggering the workflow?

lalli-flores:
seems like something is wrong with the checks?

alinasuciu:
> seems like something is wrong with the checks?

✖ Missing project token. this is the error, I tried on my local repository with the token I have from chromatic and it worked. I will do some other tests today.

lalli-flores:
The job for this PR is running on the org's CirclCI here https://app.circleci.com/pipelines/github/webex/components?branch=pull%2F217 but it's not triggering. can you try opening another PR?

lalli-flores (commented):
<del>I think I know why CI is not running. The configuration is not valid. Please fix configuration</del>
I ran the validator and it seems like this is not it 🤷🏻‍♀️

lalli-flores:
Replaced by https://github.com/webex/components/pull/223

PR #219, opened 2021-04-27
refactor(WebexMeeting): remove the state detection from the meeting hooks

lungancatalin:
Remove the state detection from the meeting hooks
https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-223895

### DIFF
--- a/src/components/hooks/useMeeting.js
+++ b/src/components/hooks/useMeeting.js
@@ -2,6 +2,19 @@ import {useEffect, useContext, useState} from 'react';
 
 import {AdapterContext} from './contexts';
 
+const emptyMeeting = {
+  ID: null,
+  title: null,
+  localAudio: null,
+  localVideo: null,
+  remoteAudio: null,
+  remoteVideo: null,
+  remoteShare: null,
+  state: null,
+  showRoster: null,
+  error: null,
+};
+
 // TODO: Figure out how to import JS Doc definitions and remove duplication.
 /**
  * A video conference in Webex over WebRTC.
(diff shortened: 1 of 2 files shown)


src/components/hooks/useMeetingDestination.js
```diff
@@ -26,39 +26,42 @@ export default function useMeetingDestination(meetingDestination) {
...
-      setMeeting(emptyMeeting);
-      // eslint-disable-next-line no-console
-      console.log(error);
-    };
-    const onComplete = () => {
-      setMeeting(endedMeeting);
-    };
+    let cleanup;
+
+    if (!meetingDestination) {
```
cipak, line 38: @lungancatalin please do the same modifications on the `useMeeting` hook.

src/components/hooks/useMeeting.js
```diff
@@ -48,13 +48,10 @@ export default function useMeeting(meetingID) {
 
       throw error;
     };
-    const onComplete = () => {
-      setMeeting(emptyMeeting);
-    };
 
     const subscription = meetingsAdapter
       .getMeeting(meetingID)
```
cipak, line 53: let's only call this if the meetingID is set, otherwise we would have to implement this "protection" in each adapter (the json adapter is ok with a falsy meeting id, but the sdk adapter currently isn't). You can use the same approach as for a falsy `meetingDestination` in the `useMeetingDestination` hook

src/components/hooks/useMeeting.js
```diff
@@ -34,31 +34,37 @@ export default function useMeeting(meetingID) {
...
+      cleanup = undefined;
+    } else {
+      const onMeeting = (newMeeting) => {
+        // React won't recognize the meeting attributes have been updated
+        // since the state is the meeting object itself. We need to create a new
+        // meeting object trigger the state change
+        setMeeting({...newMeeting});
+      };
+      const onError = (error) => {
+        setMeeting({
```
cipak, line 50: one liner pls, like in the other hook

src/components/hooks/useMeeting.js
```diff
@@ -34,31 +34,37 @@ export default function useMeeting(meetingID) {
...
+          ...emptyMeeting,
+          error,
+        });
 
-    const subscription = meetingsAdapter
-      .getMeeting(meetingID)
-      .subscribe(onMeeting, onError, onComplete);
+        throw error;
```
cipak, line 55: just logging the error to the console should be enough

cipak (approved):

src/components/hooks/useMeetingDestination.js
```diff
@@ -26,39 +26,42 @@ export default function useMeetingDestination(meetingDestination) {
...
+      cleanup = () => {
+        subscription.unsubscribe();
+      };
+    }
 
-    return () => {
-      subscription.unsubscribe();
-    };
+    return cleanup;
     // eslint-disable-next-line react-hooks/exhaustive-deps
```
lalli-flores, line 65: Let's fix this linter issue instead of disabling. We should re-run the hook if the adapter changes

src/components/hooks/useMeeting.js
```diff
@@ -35,31 +35,34 @@ export default function useMeeting(meetingID) {
...
-    return () => {
-      subscription.unsubscribe();
-    };
+      cleanup = () => {
+        subscription.unsubscribe();
+      };
+    }
+
+    return cleanup;
     // eslint-disable-next-line react-hooks/exhaustive-deps
```
lalli-flores, line 66: Let's fix this linter issue instead of disabling. We should re-run the hook if the adapter changes

lalli-flores (approved):
Minor unrelated comments but we should address them since we're refactoring

PR #220, opened 2021-04-28
build(workflow): update dependency on event

maloun96:


.github/workflows/update-dependencies.yml
```diff
@@ -0,0 +1,34 @@
...
+
+jobs:
+  update-webex-component-adapter-interfaces:
+    runs-on: ubuntu-latest
+    if: github.event.action == 'update-webex-component-adapter-interfaces'
+    steps:
+      - name: checkout
+        uses: actions/checkout@v2-beta
+      - run: |
+          wanted=$(npm outdated @webex/component-adapter-interfaces --json | grep "wanted")
```
cipak, line 13: maybe a simpler version
```sh
npm outdated @webex/components --json | grep wanted | cut -f4 -d\"
1.62.0
```

`npm outdated` has also a `--parseable` option but it separates tokens by colons (:), and on windows the same character is used for the drive name (eg: `C:`)

.github/workflows/update-dependencies.yml
```diff
@@ -0,0 +1,34 @@
...
+          echo ${wanted}
+
+          current=$(npm outdated @webex/component-adapter-interfaces --json | grep "current")
+          IFS='"' read -ra current <<< "$current"
+          current=${current[3]}
+          echo ${current}
+
+          if [ "$current" != "$wanted" ]; then
+            echo "Not Equal"
+            npm install @webex/component-adapter-interfaces@${wanted} --save
```
cipak, line 25: I guess `--save` is implicit now

.github/workflows/update-dependencies.yml
```diff
@@ -0,0 +1,20 @@
...
+    steps:
+      - name: checkout
+        uses: actions/checkout@v2-beta
+      - run: |
+          npm install @webex/component-adapter-interfaces --save
+          if [ -n "$(git status --porcelain)" ]; then
+            git config user.name "GitHub Actions Bot"
+            git config user.email "<>"
+            git add .
+            git commit -m 'fix(@webex/component-adapter-interfaces): update to new version'
```
cipak, line 18: @lalli-flores I would prefer dependency updates to use `build()`, as they should not yet use the features in the new dependency version (eg: a new field in an adapter interface). It's the subsequent commits that start using those features, and it's those commits that should generate new versions on the current repository.

cipak: I would prefer the same commit message we are currently using:
`build(package): update component adapter interfaces to v1.2.3`

.github/workflows/update-dependencies.yml
cipak: so you can remove it

.github/workflows/update-dependencies.yml
```diff
@@ -0,0 +1,20 @@
...
+    runs-on: ubuntu-latest
+    if: github.event.action == 'component-adapter-interfaces:version-changed'
+    steps:
+      - name: checkout
+        uses: actions/checkout@v2-beta
+      - run: |
+          npm install @webex/component-adapter-interfaces
+          if [ -n "$(git status --porcelain)" ]; then
+            git config user.name "GitHub Actions Bot"
+            git config user.email "<>"
```
cipak, line 16: `41898282+github-actions[bot]@users.noreply.github.com`

taken from here: https://github.community/t/github-actions-bot-email-address/17204/6

.github/workflows/update-dependencies.yml
lalli-flores: yeah, I'm good with your commit message suggstion

.github/workflows/update-dependencies.yml
```diff
@@ -0,0 +1,20 @@
...
+
+on: [repository_dispatch]
+
+jobs:
+  update-webex-component-adapter-interfaces:
+    runs-on: ubuntu-latest
+    if: github.event.action == 'component-adapter-interfaces:version-changed'
+    steps:
+      - name: checkout
+        uses: actions/checkout@v2-beta
```
lalli-flores, line 11: Sorry, I just noticed this, but why are we using the beta version here?

lalli-flores (approved):

PR #221, opened 2021-04-28
build(workflow): trigger event on new version

maloun96:
Dispatches an event to notify downstream dependency, [widget](https://github.com/webex/widget) repositories, that a version has changed.
[SPARK-227496](https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-227496)


.github/workflows/dispatch-new-version-event.yml
```diff
@@ -0,0 +1,19 @@
...
+    branches:
+      - master
+  repository_dispatch:
+
+jobs:
+  update-webex-components:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Dispatch initiating repository event
+        # TODO: change urls for repositories
```
cipak, line 14: @lalli-flores if you can approve despite the urls not being updated, we'll update the urls tomorrow and merge

.github/workflows/dispatch-new-version-event.yml
```diff
@@ -0,0 +1,19 @@
...
+  push:
+    branches:
+      - master
+  repository_dispatch:
+
+jobs:
+  update-webex-components:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Dispatch initiating repository event
```
cipak, line 13: > Dispatch new version event

lalli-flores (changes requested):
This workflow needs to look a lot more like the one in the [interfaces repo](https://github.com/webex/component-adapter-interfaces/blob/master/.github/workflows/dispatch-new-version-event.yml). Not only the URLs need to be updated (I know you're aware) but we also need to make sure that 
* This only triggers on push to `master` and not on dispatch
* This only calls the dispatch API if there's actually a version change

Please also update your description to something similar like the [PR in the interfaces repo](https://github.com/webex/component-adapter-interfaces/pull/53)

.github/workflows/dispatch-new-version-event.yml
```diff
@@ -0,0 +1,19 @@
...
+        # TODO: change urls for repositories
+        run: |
+          curl -X POST https://api.github.com/repos/maloun96/widgets/dispatches \
+          -H 'Accept: application/vnd.github.everest-preview+json' \
+          -u ${{ secrets.PERSONAL_TOKEN }} \
```
lalli-flores, line 18:
```suggestion
          -u ${{ secrets.ACCESS_TOKEN }} \
```

lalli-flores (approved):

PR #225, opened 2021-05-04
ci: don't run Storybook previews on master

lalli-flores:
Storybook previews run on all branches for all jobs. Previews should not get generated for `master`


taymoork2 (approved):
FISI

PR #234, opened 2021-05-13
fix(useMeetingDestination): work even if an adapter is not yet present on the context

lungancatalin:
Make the useMeetingDestination hook work even if an adapter is not yet present on the context
https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-233400

### DIFF
--- a/src/components/hooks/useMeetingDestination.js
+++ b/src/components/hooks/useMeetingDestination.js
@@ -30,12 +30,13 @@ const emptyMeeting = {
  */
 export default function useMeetingDestination(meetingDestination) {
   const [meeting, setMeeting] = useState(emptyMeeting);
-  const {meetingsAdapter} = useContext(AdapterContext);
+  const adapter = useContext(AdapterContext);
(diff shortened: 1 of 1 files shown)


src/components/hooks/useMeetingDestination.js
```diff
@@ -30,12 +30,13 @@ const emptyMeeting = {
...
   const [meeting, setMeeting] = useState(emptyMeeting);
-  const {meetingsAdapter} = useContext(AdapterContext);
+  const adapter = useContext(AdapterContext);
+  const meetingsAdapter = adapter && adapter.meetingsAdapter;
 
   useEffect(() => {
     let cleanup;
 
-    if (!meetingDestination) {
+    if (!meetingDestination || !adapter) {
```
cipak, line 39: you can directly check for `meetingsAdapter` here, it's either undefined or set

cipak: and one less dependency in this effect's deps list

lalli-flores (approved):

PR #238, opened 2021-05-18
fix(workflow): dispatch new version also on new release event

maloun96:
After some investigation, the "chor" release does not trigger events (https://github.com/webex/components/actions/workflows/dispatch-new-version-event.yml) Here are the latest triggered events and there is no commit with "chor". This PR is going to fix it.


.github/workflows/dispatch-new-version-event.yml
```diff
@@ -17,6 +19,9 @@ jobs:
           git checkout HEAD~1
           PREVIOUS_VERSION=$(node -p -e "require('./package.json').version")
 
+          echo "Previous version: $PREVIOUS_VERSION"
```
maloun96, line 22: Added this logs to see if it passed and the trigger happens

PR #244, opened 2021-05-24
feat(WebexVideoSettings): show all available cameras in a dropdown

lungancatalin:
List a dropdown for all available cameras
https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-235908

### DIFF
--- a/src/components/WebexVideoSettings/WebexVideoSettings.jsx
+++ b/src/components/WebexVideoSettings/WebexVideoSettings.jsx
@@ -1,0 +1,40 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import webexComponentClasses from '../helpers';
+import WebexLocalMedia from '../WebexLocalMedia/WebexLocalMedia';
+import WebexMeetingControl from '../WebexMeetingControl/WebexMeetingControl';
+
+/**
+ * Webex Video Settings component
+ *
+ * @param {object} props  Data passed to the component
+ * @param {string} props.className  Custom CSS class to apply
+ * @param {string} props.meetingID  ID of the meeting from which to obtain local media
+ * @param {object} props.style  Custom style to apply
+ * @returns {object} JSX of the component
+ */
+export default function WebexVideoSettings({meetingID, className, style}) {
+  const cssClasses = webexComponentClasses('video-settings', className);
+
+  return (
+    <div className={cssClasses} style={style}>
+      <h2>Camera</h2>
+      <WebexMeetingControl type="switch-camera" meetingID={meetingID} />
+      <div className="media">
+        <WebexLocalMedia mediaType="video" meetingID={meetingID} />
+        <div className="preview"><h3>Preview</h3></div>
+      </div>
+    </div>
+  );
+}
+
+WebexVideoSettings.propTypes = {
+  className: PropTypes.string,
+  meetingID: PropTypes.string.isRequired,
+  style: PropTypes.shape(),
+};
+
+WebexVideoSettings.defaultProps = {
+  className: '',
+  style: undefined,
+};
--- a/src/components/WebexVideoSettings/WebexVideoSettings.scss
+++ b/src/components/WebexVideoSettings/WebexVideoSettings.scss
@@ -1,0 +1,15 @@
+.#{$WEBEX_COMPONENTS_CLASS_PREFIX}-video-settings {
+  .media {
+    position: relative;
+    margin-top: 1.5rem;
+  }
+
+  .preview {
+    position: absolute;
+    bottom: 0;
+    width: 100%;
+    text-align: center;
+    background-color: rgba($md-white, .6);
+    color: $md-black-100;
+  }
+}
--- a/src/components/WebexSettings/WebexSettings.jsx
+++ b/src/components/WebexSettings/WebexSettings.jsx
@@ -3,6 +3,7 @@ import PropTypes from 'prop-types';
 import webexComponentClasses from '../helpers';
 import Tabs from './Tabs';
 import WebexMicrophoneSettings from '../WebexMicrophoneSettings/WebexMicrophoneSettings';
+import WebexVideoSettings from '../WebexVideoSettings/WebexVideoSettings';
 
 /**
  * Webex Settings component
@@ -27,13 +28,7 @@ export default function WebexSettings({meetingID, className, style}) {
     key: 'video',
     heading: 'Video',
     content: () => (
-      <div>
-        Video Settings for meeting
-        {' '}
-        {meetingID}
-        {' '}
-        (TBD)
-      </div>
+      <WebexVideoSettings type="switch-camera" meetingID={meetingID} />
     ),
   }];
 
--- a/src/components/WebexSettings/WebexSettings.stories.js
+++ b/src/components/WebexSettings/WebexSettings.stories.js
@@ -6,7 +6,7 @@ export default {
   component: WebexSettings,
 };
 
-const Template = (args) => <WebexSettings {...args} />;
+const Template = (args) => <WebexSettings {...args} style={{flex: 1, alignSelf: 'stretch'}} />;
 
 export const Settings = Template.bind({});
 Settings.args = {
--- a/src/components/WebexSettings/__snapshots__/WebexSettings.stories.storyshot
+++ b/src/components/WebexSettings/__snapshots__/WebexSettings.stories.storyshot
@@ -4,6 +4,12 @@ exports[`Storyshots Meetings/Webex Settings Settings 1`] = `
 Array [
   <div
     className="wxc-settings"
+    style={
+      Object {
+        "alignSelf": "stretch",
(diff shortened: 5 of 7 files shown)


src/components/WebexVideoSettings/WebexVideoSettings.jsx
```diff
@@ -0,0 +1,65 @@
+import React, {useEffect, useState} from 'react';
+import {
```
cipak, line 2: single line pls

src/components/WebexVideoSettings/WebexVideoSettings.jsx
```diff
@@ -0,0 +1,65 @@
...
+  Select, SelectOption,
+} from '@momentum-ui/react';
+import PropTypes from 'prop-types';
+import classNames from 'classnames';
+import {WEBEX_COMPONENTS_CLASS_PREFIX} from '../../constants';
+import WebexLocalMedia from '../WebexLocalMedia/WebexLocalMedia';
+
+/**
+ *
```
cipak, line 11: put the description of the function first "Returns all of available cameras"
and a minimal message in the return, but with the correct type (array):
```js
* @returns {object[]} Available cameras
* ```

File: src/components/WebexVideoSettings/WebexVideoSettings.jsx [hunk 3]
```diff
@@ -0,0 +1,65 @@
...
+ * @returns {object} Returns all of available cameras
+ */
+
+const getAvailableCameras = async (setDevices) => {
+  const cameraDevices = await navigator.mediaDevices.enumerateDevices()
+    .then((cameras) => cameras.filter((camera) => camera.kind === 'videoinput'));
+
+  setDevices([...cameraDevices]);
```
cipak, line 19: this function does not seem to return the camera, it just calls the provided callback (`setDevices`) with the camera list as a parameter.

pls change this function to return the camera list, and use the `setDevices` function in the component itself.

src/components/WebexVideoSettings/WebexVideoSettings.jsx
```diff
@@ -0,0 +1,65 @@
...
+};
+
+/**
+ * Webex Video Settings component
+ *
+ * @param {string} props.className  Custom CSS class to apply
+ * @returns {object} JSX of the component
+ */
+export default function WebexVideoSettings({className, meetingID}) {
+  const [devices, setDevices] = useState([]);
```
cipak, line 29: There might be a state where the user has no available camera, which would translate in the same value for `devices`, an empty array. We need to distinguish between "loading the cameras" state and "no cameras available" state.

Let's use `null` as the initial value, and as an idicator that cameras are currently loading.

In the render function, please display a message if `devices && devices.length === 0` (which means no camera is available). See an example message here: https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-231001

src/components/WebexVideoSettings/WebexVideoSettings.jsx
```diff
@@ -0,0 +1,65 @@
...
+  const classBaseName = `${WEBEX_COMPONENTS_CLASS_PREFIX}-video-settings`;
+  const mainClasses = {
+    [classBaseName]: true,
+    [className]: !!className,
+  };
+
+  const extractCameraName = (label) => label.split('(', 1)[0];
```
cipak, line 36: why is this needed? can you list an example output here please?

src/components/WebexVideoSettings/WebexVideoSettings.jsx
```diff
@@ -0,0 +1,65 @@
...
+import {WEBEX_COMPONENTS_CLASS_PREFIX} from '../../constants';
+import WebexLocalMedia from '../WebexLocalMedia/WebexLocalMedia';
+
+/**
+ *
+ * @returns {object} Returns all of available cameras
+ */
+
+const getAvailableCameras = async (setDevices) => {
+  const cameraDevices = await navigator.mediaDevices.enumerateDevices()
```
cipak, line 16: `navigator.mediaDevices` does not seem to be defined and tests are failing:
> TypeError: Cannot read property 'enumerateDevices' of undefined
    at _callee$ (/home/circleci/webex-components/src/components/WebexVideoSettings/WebexVideoSettings.jsx:16:54)

I think you can mock it in `src/storyshots.test.js` using the technique described here: https://stackoverflow.com/a/65216531/210550

But, instead of an empty value, please return an example list of devices (not just cameras!):
```js

const availableMediaDevices = [
   ...device1...,
   ...device2...,
   ...etc...
];
...
enumerateDevices: jest.fn(() => Promise.resovle(availableMediaDevices)),
```

cipak:
commit message: feat(WebexVideoSettings): show all available cameras in a dropdown

src/components/WebexVideoSettings/WebexVideoSettings.jsx
lungancatalin: Each camera has a code number at the end of its label and that's why I created this feature to remove it.

label: "Logitech HD Webcam C525 (046d:0826)" => Logitech HD Webcam C525

src/storyshots.test.js
```diff
@@ -34,14 +53,21 @@ initStoryshots({
...
-    // Because observables are async, execute snapshot tests on next event loop cycle
-    setTimeout(() => {
-      if (snapshotFilename) {
-        expect(tree.toJSON()).toMatchSpecificSnapshot(snapshotFilename);
-      }
-      done();
-    }, 0);
+    let tree;
+
+    global.navigator.mediaDevices = mediaDevicesMock;
```
alinasuciu, line 58: Maybe we could use Enzyme or render from React Testing Library.
https://www.npmjs.com/package/@storybook/addon-storyshots
Check the addon-storyshots documentation, this is what they recommend:
"By design, react-test-renderer doesn't use a browser environment or JSDOM. Because of this difference, some stories might render in your browser, but not in Storyshots. If you encounter this problem, you may want to switch for an higher level renderer such as mount from Enzyme or render from React Testing Library."

src/components/WebexVideoSettings/WebexVideoSettings.jsx
```diff
@@ -0,0 +1,62 @@
...
+import {useMeetingControl} from '../hooks';
+
+const noCameraMsg = 'Can\'t detect your camera. Make sure your camera is connected and try again.';
+
+/**
+ * Webex Video Settings component
+ *
+ * @param {object} props  Data passed to the component
+ * @param {string} props.meetingID  ID of the meeting from which to obtain local media
+ * @param {string} props.className  Custom CSS class to apply
```
cipak, line 17: className before meetingID, just like in the function declaration

src/components/WebexVideoSettings/WebexVideoSettings.jsx
```diff
@@ -0,0 +1,62 @@
...
+ * @returns {object} JSX of the component
+ */
+export default function WebexVideoSettings({className, meetingID}) {
+  const [, display] = useMeetingControl('switch-camera', meetingID);
```
cipak, line 21: we should not need to look into switch-camera control's internals... let me think of an alternative

src/components/WebexVideoSettings/WebexVideoSettings.scss
```diff
@@ -0,0 +1,26 @@
+.#{$WEBEX_COMPONENTS_CLASS_PREFIX}-video-settings {
+  position: relative;
+  .camera-warning {
+    color: $md-red-60;
+    display: flex;
+    justify-content: space-between;
```
cipak, line 6: does this mean that the extra space will be inserted between the icon and the text?

src/components/WebexVideoSettings/WebexVideoSettings.scss
```diff
@@ -0,0 +1,26 @@
...
+    align-items: start;
+    padding: 0.25rem 0 0.25rem 0;
+
+    .message {
+      margin-left: 0.625rem;
+      font-size: 1rem;
+    }
+  }
+  .preview {
+    position: absolute;
```
cipak, line 16: I don't see a reason for absolutely positioning this element. If the container is a normal div, or a column flex container, the preview should be displayed fine under the control.

cipak: ok, I see now, this is just the "Preview" banner, can you display it relative to the LocalMedia component?

src/components/WebexVideoSettings/WebexVideoSettings.scss
```diff
@@ -0,0 +1,26 @@
...
+    justify-content: space-between;
+    align-items: start;
+    padding: 0.25rem 0 0.25rem 0;
+
+    .message {
+      margin-left: 0.625rem;
+      font-size: 1rem;
+    }
+  }
+  .preview {
```
cipak, line 15: leave a blank line before a new rule, thanks!

cipak (approved):

PR #250, opened 2021-06-02
feat(Select): implement Select component

CernaianuMihai:
This PR is about implementing the Select component

https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-237837


src/components/_components.scss
```diff
@@ -10,3 +10,4 @@
 @import 'WebexMember/WebexMember';
 @import 'WebexMemberRoster/WebexMemberRoster';
 @import 'WebexRemoteMedia/WebexRemoteMedia';
+@import 'WebexSettings/WebexSettings';
```
cipak, line 13: we should not export the (internal) Select component

src/components/WebexSettings/Select.stories.js
```diff
@@ -0,0 +1,14 @@
+import React from 'react';
```
cipak, line 1: there is no need for a dedicated story for the Select component, we are not inviting developers to use it, it's just for internal use.

src/components/WebexSettings/Select.scss
```diff
@@ -0,0 +1,10 @@
+.#{$WEBEX_COMPONENTS_CLASS_PREFIX}-select {
+  select {
+    font-family: "CiscoSansTT Regular","Helvetica Neue",Helvetica,Arial,sans-serif;
```
cipak, line 3: isn't the font family already set globally?

src/components/WebexSettings/Select.jsx
```diff
@@ -0,0 +1,34 @@
...
+export default function Select({className, ...props}) {
+  const mainClasses = {
+    [`${WEBEX_COMPONENTS_CLASS_PREFIX}-select`]: true,
+    [className]: !!className,
+  };
+
+  return (
+    <div className={classNames(mainClasses)}>
+      {/* eslint-disable-next-line react/jsx-props-no-spreading */}
+      <select {...props} />
```
cipak, line 23: maybe send the options as a prop as well?

src/components/WebexSettings/Select.stories.js
CernaianuMihai: Ok, I'll remove it.

src/components/WebexSettings/Select.scss
CernaianuMihai: Yes, it is. I'll remove this line.

src/components/_components.scss
cipak: sorry, I read "export" instead of "import":)

src/components/WebexSettings/Select.jsx
```diff
@@ -0,0 +1,47 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import classNames from 'classnames';
+import {WEBEX_COMPONENTS_CLASS_PREFIX} from '../../constants';
+
+/**
+ * Select component
+ *
+ * @param {object} props Data passed to the component
```
cipak, line 9: use double space between arg name and description `props  Data passed...`

src/components/WebexSettings/Select.jsx
```diff
@@ -0,0 +1,47 @@
...
+ * @param {object[]} props.options Array of HMTL option elements
```
cipak, line 10: they are not (yet) html option elements, they are just simple objects with 2 props: value and label.

I see that you only use value, but value usually is some id (eg: camera device id) so we need a label to display to the user.

Please document these 2 props in the jsdoc, eg:
```js
* @param {string} props.options[].value  The value of the option
```

src/components/WebexSettings/Select.jsx
```diff
@@ -0,0 +1,47 @@
...
+  };
+
+  return (
+    <div className={classNames(mainClasses)}>
+      {/* eslint-disable-next-line react/jsx-props-no-spreading */}
+      <select {...props}>
+        {
+          options?.map((option) => (
+            <option
+              value={option.value}
```
cipak, line 28: you can put this on a single line

src/components/WebexSettings/Select.jsx
```diff
@@ -0,0 +1,47 @@
...
+          ))
+        }
+      </select>
+    </div>
+  );
+}
+
+Select.propTypes = {
+  className: PropTypes.string,
+  options: PropTypes.arrayOf(PropTypes.shape({value: PropTypes.value})),
```
cipak, line 41: nice. Please add the `label` as well.

src/components/WebexSettings/Select.jsx
```diff
@@ -0,0 +1,47 @@
...
+ * @returns {object} JSX of the element
+ */
+export default function Select({className, options, ...props}) {
+  const mainClasses = {
+    [`${WEBEX_COMPONENTS_CLASS_PREFIX}-select`]: true,
+    [className]: !!className,
+  };
+
+  return (
+    <div className={classNames(mainClasses)}>
```
cipak, line 22: if not 100% needed, please remove this `<div>` wrapper and put the class directly on the `<select>` element.

src/components/WebexSettings/Select.jsx
```diff
@@ -0,0 +1,44 @@
...
+import {WEBEX_COMPONENTS_CLASS_PREFIX} from '../../constants';
+
+/**
+ * Select component
+ *
+ * @param {object} props  Data passed to the component
+ * @param {object[]} props.options  Array of option elements
+ * @param {string} props.options[].value  The value of the option
+ * @param {string} props.options[].label  The label of the option
+ * @param {string} props.className  Custom CSS class to apply
```
cipak, line 13: put className before options, just like in the function's signature

src/components/WebexSettings/Select.jsx
```diff
@@ -0,0 +1,44 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import classNames from 'classnames';
+import {WEBEX_COMPONENTS_CLASS_PREFIX} from '../../constants';
+
+/**
+ * Select component
+ *
+ * @param {object} props  Data passed to the component
+ * @param {object[]} props.options  Array of option elements
```
cipak, line 10: Array of options, or List of options

cipak (approved):

PR #253, opened 2021-06-07
feat(WebexInterstitialMeeting): add ask for video access dialog

alinasuciu:
Add access dialog inside the interstitial meeting component and create a new story for it.

https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-231516

### DIFF
--- a/src/components/WebexInterstitialMeeting/WebexInterstitialMeeting.jsx
+++ b/src/components/WebexInterstitialMeeting/WebexInterstitialMeeting.jsx
@@ -4,8 +4,11 @@ import {Spinner} from '@momentum-ui/react';
 
 import {WEBEX_COMPONENTS_CLASS_PREFIX} from '../../constants';
 
+import Modal from '../generic/Modal/Modal';
 import WebexLocalMedia from '../WebexLocalMedia/WebexLocalMedia';
 import WebexMeetingInfo from '../WebexMeetingInfo/WebexMeetingInfo';
+import WebexMediaAccess from '../WebexMediaAccess/WebexMediaAccess';
+import {useMeeting} from '../hooks';
 
 /**
  * Webex Interstitial component displays the user's local video and
@@ -16,12 +19,19 @@ import WebexMeetingInfo from '../WebexMeetingInfo/WebexMeetingInfo';
  * @returns {object} JSX of the component
  */
 export default function WebexInterstitialMeeting({meetingID}) {
+  const {videoPermission} = useMeeting(meetingID);
+
   return (
     <div className={`${WEBEX_COMPONENTS_CLASS_PREFIX}-interstitial-meeting`}>
       {meetingID ? (
         <>
           <WebexMeetingInfo className="interstitial-meeting-info" meetingID={meetingID} />
           <WebexLocalMedia className="interstitial-media" meetingID={meetingID} mediaType="video" />
+          {videoPermission === 'ASKING' && (
+            <Modal>
+              <WebexMediaAccess meetingID={meetingID} />
+            </Modal>
+          )}
         </>
       ) : (
         <Spinner />
--- a/src/components/WebexInterstitialMeeting/WebexInterstitialMeeting.stories.js
+++ b/src/components/WebexInterstitialMeeting/WebexInterstitialMeeting.stories.js
@@ -13,6 +13,14 @@ Loading.args = {
   meetingID: '',
 };
 
+export const VideoAccessDialog = Template.bind({});
+VideoAccessDialog.args = {
+  meetingID: 'meeting2',
+};
+VideoAccessDialog.parameters = {
+  storyshots: {disable: true},
+};
+
 export const VideoEnabled = Template.bind({});
 VideoEnabled.args = {
   meetingID: 'meeting1',
@@ -20,5 +28,5 @@ VideoEnabled.args = {
 
 export const VideoDisabled = Template.bind({});
 VideoDisabled.args = {
-  meetingID: 'meeting2',
+  meetingID: 'meeting5',
 };
--- a/src/data/meetings.js
+++ b/src/data/meetings.js
@@ -14,6 +14,7 @@ export default {
     showRoster: false,
     showSettings: false,
     cameraID: 'cameraID1',
+    videoPermissions: 'ALLOWED',
   },
   meeting2: {
     ID: 'meeting2',
@@ -28,6 +29,7 @@ export default {
     showRoster: false,
     showSettings: false,
     cameraID: 'cameraID2',
+    videoPermission: 'ASKING',
   },
   meeting3: {
     ID: 'meeting3',
@@ -42,6 +44,7 @@ export default {
     showRoster: false,
     showSettings: false,
     cameraID: 'cameraID1',
+    videoPermission: 'ALLOWED',
   },
   meeting4: {
     ID: 'meeting4',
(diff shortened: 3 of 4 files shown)


src/components/WebexInterstitialMeeting/WebexInterstitialMeeting.jsx
```diff
@@ -16,12 +18,29 @@ import WebexMeetingInfo from '../WebexMeetingInfo/WebexMeetingInfo';
  * @returns {object} JSX of the component
  */
 export default function WebexInterstitialMeeting({meetingID}) {
+  const {videoPermission} = useMeeting(meetingID);
+  const showModal = videoPermission === 'asking';
```
cipak, line 22: when https://github.com/webex/component-adapter-interfaces/pull/62 is merged, you can use the value from the enum.

src/components/WebexInterstitialMeeting/WebexInterstitialMeeting.stories.js
```diff
@@ -4,6 +4,9 @@ import WebexInterstitialMeeting from './WebexInterstitialMeeting';
 export default {
   title: 'Meetings/Webex Interstitial Meeting',
   component: WebexInterstitialMeeting,
+  parameters: {
+    storyshots: {disable: true},
```
cipak, line 8: are we disabling all storyshots for this component?

src/components/WebexInterstitialMeeting/WebexInterstitialMeeting.jsx
```diff
@@ -16,12 +18,29 @@ import WebexMeetingInfo from '../WebexMeetingInfo/WebexMeetingInfo';
...
     <div className={`${WEBEX_COMPONENTS_CLASS_PREFIX}-interstitial-meeting`}>
       {meetingID ? (
         <>
           <WebexMeetingInfo className="interstitial-meeting-info" meetingID={meetingID} />
           <WebexLocalMedia className="interstitial-media" meetingID={meetingID} mediaType="video" />
+          {showModal && (
+            <Modal
+              applicationId="app-media-modal"
+              onHide={() => {}}
+              show={showModal}
```
cipak, line 34: I think the condition on line 30 is sufficient

src/components/WebexInterstitialMeeting/WebexInterstitialMeeting.jsx
```diff
@@ -16,12 +18,29 @@ import WebexMeetingInfo from '../WebexMeetingInfo/WebexMeetingInfo';
...
+
   return (
     <div className={`${WEBEX_COMPONENTS_CLASS_PREFIX}-interstitial-meeting`}>
       {meetingID ? (
         <>
           <WebexMeetingInfo className="interstitial-meeting-info" meetingID={meetingID} />
           <WebexLocalMedia className="interstitial-media" meetingID={meetingID} mediaType="video" />
+          {showModal && (
+            <Modal
+              applicationId="app-media-modal"
```
cipak, line 32: do we need this prop? also, `htmlId`? I would prefer to avoid them if possible

src/components/WebexInterstitialMeeting/WebexInterstitialMeeting.stories.js
alinasuciu: yes. there is a problem with the modal testing because the snapshot doesn't contain it, that's the reason why I choose to disable the story shots. or we can do it just for one story but this means that none of the other stories would display a meeting witch has the 'asking' value for video permission.

src/components/WebexInterstitialMeeting/WebexInterstitialMeeting.jsx
alinasuciu: "show" is a mandatory prop

src/components/WebexInterstitialMeeting/WebexInterstitialMeeting.jsx
alinasuciu: yes, they seems to be mandatory, if I don't add them the component will display an error in console telling that they are mandatory props

src/components/WebexInterstitialMeeting/WebexInterstitialMeeting.scss
```diff
@@ -22,3 +22,9 @@
     flex-grow: 1;
   }
 }
+
+.interstitial-media-modal {
```
cipak, line 26: if outside of the main container (`.#{$WEBEX_COMPONENTS_CLASS_PREFIX}-interstitial-meeting`), this needs to be prefixed as well. And maybe call it `{PREFIX}-interstitial-meeting__media-access-modal`

cipak: or, better yet, just send a classname to the `<WebexMediaAccess meetingID={meetingID} />` component, with a `2rem` margin.

src/components/WebexInterstitialMeeting/WebexInterstitialMeeting.stories.js
cipak: I think it's fine to not having any story with "asking".

src/components/WebexInterstitialMeeting/WebexInterstitialMeeting.jsx
```diff
@@ -5,7 +5,10 @@ import {Spinner} from '@momentum-ui/react';
 import {WEBEX_COMPONENTS_CLASS_PREFIX} from '../../constants';
 
 import WebexLocalMedia from '../WebexLocalMedia/WebexLocalMedia';
+import Modal from '../generic/Modal/Modal';
```
cipak, line 8: add this before (or after) the `Webex*` block of imports

cipak: best, keep imports alphabetical

src/components/WebexInterstitialMeeting/WebexInterstitialMeeting.jsx
```diff
@@ -16,12 +19,19 @@ import WebexMeetingInfo from '../WebexMeetingInfo/WebexMeetingInfo';
...
+  const {videoPermission} = useMeeting(meetingID);
+
   return (
     <div className={`${WEBEX_COMPONENTS_CLASS_PREFIX}-interstitial-meeting`}>
       {meetingID ? (
         <>
           <WebexMeetingInfo className="interstitial-meeting-info" meetingID={meetingID} />
           <WebexLocalMedia className="interstitial-media" meetingID={meetingID} mediaType="video" />
+          {videoPermission === 'ASKING' && (
+            <Modal>
```
cipak, line 31: looks clean

src/data/meetings.js
```diff
@@ -98,6 +104,7 @@ export default {
     showRoster: false,
     showSettings: false,
     cameraID: null,
+    videoPermission: 'UNKNOWN',
```
cipak, line 107: use `null` instead of `'UNKNOWN'`, we removed that value from the interface https://github.com/webex/component-adapter-interfaces/blob/master/src/MeetingsAdapter.js#L95

src/components/WebexInterstitialMeeting/WebexInterstitialMeeting.stories.js
```diff
@@ -13,12 +13,20 @@ Loading.args = {
   meetingID: '',
 };
 
+export const VideoAccessDialog = Template.bind({});
+VideoAccessDialog.args = {
+  meetingID: 'meeting2',
+};
+VideoAccessDialog.parameters = {
+  storyshots: {disable: true},
```
alinasuciu, line 21: check: https://www.npmjs.com/package/@storybook/addon-storyshots

cipak:
closing in favor of https://github.com/webex/widgets/pull/101

PR #259, opened 2021-06-11
feat(WebexMeetingIf): add WebexMeetingIf component

karinasigartau0798:
This component is a wrapper that, depending on the truth value obtained after calling the function defined in props.condition, renders the child components.
https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-241477

### DIFF
--- a/src/components/WebexMeetingIf/WebexMeetingIf.jsx
+++ b/src/components/WebexMeetingIf/WebexMeetingIf.jsx
@@ -1,0 +1,35 @@
+import PropTypes from 'prop-types';
+
+import {useMeeting} from '../hooks';
+
+/**
+ * WebexMeetingIf
+ *
+ * This helper component makes it easy for developers to add content conditionally,
+ * depending on the current meeting's state, without obtaining the meeting object with
+ * the adapter getMeeting() function or the useMeeting() hook.
+ *
+ * Example:
+ *
+ * <WebexMeetingIf meetingID="someid" condition={meeting => meeting.state !== 'LEFT'}>
+ *  <WebexMeetingControlBar controls={...} />
+ * </WebexMeetingIf>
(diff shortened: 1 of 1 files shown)


src/components/WebexMeetingIf/WebexMeetingIf.jsx
```diff
@@ -0,0 +1,27 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+import {useMeeting} from '../hooks';
+
+/**
+ * WebexMeetingIf
+ *
```
cipak, line 8: This helper component makes it easy for developers to add content conditionally, depending on the current meeting's state, without obtaining the meeting object with the adapter `getMeeting()` function or the `useMeeting()` hook.

Example:
```js
<WebexMeetingIf meetingID="someid" condition={meeting => meeting.state !== 'LEFT'}>
  <WebexMeetingControlBar controls={...} />
</WebexMeetingIf>
```

src/components/WebexMeetingIf/WebexMeetingIf.jsx
```diff
@@ -0,0 +1,27 @@
...
+ * @param {object} props  Data passed to the component
+ * @param {string} props.meetingID  ID of the meeting to control
+ * @param {object} props.condition  Function that returns true or false
```
cipak, line 11: maybe `{function(meeting: object): boolean}` instead of `{object}`

lalli-flores (changes requested):
This component does not make sense to me. I think this makes things more complicated for a developer. 
Additionally, it doesn't align with how the rest of the components work, as it doesn't represent a piece of UI. Too much abstraction can make code harder to read, so I would be weary of doing something like this.

PR #260, opened 2021-06-11
feat(Modal): add Modal component

CernaianuMihai:
This PR is about implementing the Modal component

https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-241415

### DIFF
--- a/src/components/_components.scss
+++ b/src/components/_components.scss
@@ -11,3 +11,4 @@
 @import 'WebexMemberRoster/WebexMemberRoster';
 @import 'WebexRemoteMedia/WebexRemoteMedia';
 @import 'WebexSettings/WebexSettings';
+@import 'generic';
--- a/src/components/generic/Modal/Modal.jsx
+++ b/src/components/generic/Modal/Modal.jsx
@@ -1,0 +1,64 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import classNames from 'classnames';
+import {Modal as MomentumModal, ModalBody, ModalHeader} from '@momentum-ui/react';
+import {WEBEX_COMPONENTS_CLASS_PREFIX} from '../../../constants';
+
+/**
+ * Modal component
+ *
+ * @param {object} props  Data passed to the component
+ * @param {string} [props.applicationId]  Application DOM id
+ * @param {React.ReactNode[]} props.children  List of children
+ * @param {string} [props.className]  Custom CSS class to apply
+ * @param {Function} [props.onClose]  Callback when the modal is closed
+ * @param {string} [props.title]  Title of the modal
+ *
+ * @returns {object}  JSX of the element
+ */
+export default function Modal({
+  applicationId,
+  children,
+  className,
+  onClose,
+  title,
+  ...props
+}) {
+  const mainClasses = {
+    [`${WEBEX_COMPONENTS_CLASS_PREFIX}-modal`]: true,
+    [className]: !!className,
+  };
+  const testId = `${WEBEX_COMPONENTS_CLASS_PREFIX}-${Math.random().toString().slice(2)}`;
+
+  return (
+    <MomentumModal
+      applicationId={applicationId}
+      className={classNames(mainClasses)}
+      htmlId={testId}
+      onHide={onClose}
+      show
+      // eslint-disable-next-line react/jsx-props-no-spreading
+      {...props}
+    >
+      <ModalHeader headerLabel={title} showCloseButton={!!onClose} />
+      <ModalBody>
+        {children}
+      </ModalBody>
+    </MomentumModal>
+  );
+}
+
+Modal.propTypes = {
+  applicationId: PropTypes.string,
+  children: PropTypes.node.isRequired,
+  className: PropTypes.string,
+  onClose: PropTypes.func,
+  title: PropTypes.string,
+};
+
+Modal.defaultProps = {
+  applicationId: 'app',
+  className: '',
+  onClose: undefined,
+  title: '',
+};
--- a/src/components/generic/Modal/Modal.scss
+++ b/src/components/generic/Modal/Modal.scss
@@ -1,0 +1,5 @@
+.#{$WEBEX_COMPONENTS_CLASS_PREFIX}-modal {
+  .md-modal__content {
+    padding: 2rem;
+  }
+}
--- a/src/components/generic/index.scss
+++ b/src/components/generic/index.scss
@@ -1,0 +1,1 @@
+@import 'Modal/Modal';
--- a/src/components/index.js
+++ b/src/components/index.js
@@ -16,5 +16,6 @@ export {default as WebexSettings} from './WebexSettings/WebexSettings';
 export {default as useMeetingDestination} from './hooks/useMeetingDestination';
 export {default as withMeeting} from './hoc/withMeeting';
 export {default as withAdapter} from './hoc/withAdapter';
+export {default as Modal} from './generic/Modal/Modal';
 
 global.testRendererAct = (fn) => fn();


src/components/generic/Modal/Modal.jsx
```diff
@@ -0,0 +1,70 @@
...
+import PropTypes from 'prop-types';
+import classNames from 'classnames';
+import {Modal, ModalBody, ModalHeader} from '@momentum-ui/react';
+import {v4 as uuidv4} from 'uuid';
+import {WEBEX_COMPONENTS_CLASS_PREFIX} from '../../../constants';
+
+/**
+ * Modal component
+ *
+ * @param {object} props Data passed to the component
```
cipak, line 11: double space

src/components/generic/Modal/Modal.jsx
```diff
@@ -0,0 +1,70 @@
...
+  onHide,
+  show,
+  showCloseButton,
+  ...props
+}) {
+  const mainClasses = {
+    [`${WEBEX_COMPONENTS_CLASS_PREFIX}-modal`]: true,
+    [className]: !!className,
+  };
+  const customId = `${WEBEX_COMPONENTS_CLASS_PREFIX}-${uuidv4()}`;
```
cipak, line 34: let's not introduce a new dependency. Just use `Math.random()` instead of uuid.

src/components/generic/Modal/Modal.jsx
```diff
@@ -0,0 +1,70 @@
...
+      onHide={onHide}
+      show={show}
+      htmlId={customId}
+      className={classNames(mainClasses)}
+      // eslint-disable-next-line react/jsx-props-no-spreading
+      {...props}
+    >
+      <ModalHeader headerLabel={headerLabel} showCloseButton={showCloseButton} />
+      <ModalBody>
+        {children?.map((child) => child)}
```
cipak, line 48: this map seems futile:)

src/components/generic/Modal/Modal.jsx
CernaianuMihai: I just saw that there was already an older version of the uuid installed so I can use that without installing anything new

cipak (approved):

src/components/generic/Modal/Modal.jsx
```diff
@@ -0,0 +1,69 @@
...
+import {Modal, ModalBody, ModalHeader} from '@momentum-ui/react';
+import {WEBEX_COMPONENTS_CLASS_PREFIX} from '../../../constants';
+
+/**
+ * Modal component
+ *
+ * @param {object} props  Data passed to the component
+ * @param {string} props.className   Custom CSS class to apply
+ * @param {string[]} props.children  List of children
+ * @param {string} props.headerLabel  Title of the Modal
```
cipak, line 13: lowercase modal

src/components/generic/Modal/Modal.jsx
```diff
@@ -0,0 +1,69 @@
...
+import classNames from 'classnames';
+import {Modal, ModalBody, ModalHeader} from '@momentum-ui/react';
+import {WEBEX_COMPONENTS_CLASS_PREFIX} from '../../../constants';
+
+/**
+ * Modal component
+ *
+ * @param {object} props  Data passed to the component
+ * @param {string} props.className   Custom CSS class to apply
+ * @param {string[]} props.children  List of children
```
cipak, line 12: children are not necessarily strings, see https://stackoverflow.com/a/57253387/210550 for a better type

src/components/generic/Modal/Modal.jsx
```diff
@@ -0,0 +1,69 @@
...
+import PropTypes from 'prop-types';
+import classNames from 'classnames';
+import {Modal, ModalBody, ModalHeader} from '@momentum-ui/react';
+import {WEBEX_COMPONENTS_CLASS_PREFIX} from '../../../constants';
+
+/**
+ * Modal component
+ *
+ * @param {object} props  Data passed to the component
+ * @param {string} props.className   Custom CSS class to apply
```
cipak, line 11: too many spaces:)

cipak: for optional props, please use the square bracket syntax: `[props.className]`, see https://jsdoc.app/tags-param.html#optional-parameters-and-default-values

src/components/generic/Modal/Modal.jsx
```diff
@@ -0,0 +1,69 @@
...
+ * @param {string[]} props.children  List of children
+ * @param {string} props.headerLabel  Title of the modal
```
cipak, line 13: let's call this `{props.title}`

src/components/generic/Modal/Modal.jsx
```diff
@@ -0,0 +1,69 @@
...
+ * @param {Function} props.onHide  Callback when the modal is closed
+ * @param {boolean} props.show  Show or hide the modal
```
cipak, line 15: this prop should not be needed, the modal should be used like this:
```js
{show && <Modal ... />}
```

src/components/generic/Modal/Modal.jsx
```diff
@@ -0,0 +1,69 @@
...
+import {WEBEX_COMPONENTS_CLASS_PREFIX} from '../../../constants';
+
+/**
+ * Modal component
+ *
+ * @param {object} props  Data passed to the component
+ * @param {string} props.className   Custom CSS class to apply
+ * @param {string[]} props.children  List of children
+ * @param {string} props.headerLabel  Title of the modal
+ * @param {Function} props.onHide  Callback when the modal is closed
```
cipak, line 14: onClose

src/components/generic/Modal/Modal.jsx
```diff
@@ -0,0 +1,69 @@
...
+ * @param {boolean} props.show  Show or hide the modal
+ * @param {boolean} props.showCloseButton  Show or hide the modal close button
```
cipak, line 16: I think the close button (X) should always be shown

src/components/generic/Modal/Modal.jsx
```diff
@@ -0,0 +1,69 @@
...
+}) {
+  const mainClasses = {
+    [`${WEBEX_COMPONENTS_CLASS_PREFIX}-modal`]: true,
+    [className]: !!className,
+  };
+  const customId = `${WEBEX_COMPONENTS_CLASS_PREFIX}-${Math.floor((Math.random() * 100000000) + 1)}`;
+
+  return (
+    <Modal
+      applicationId={customId}
```
cipak, line 37: application id is the id of the main element of the application, see its short description here: https://github.com/momentum-design/momentum-ui/blob/master/react/src/lib/Modal/index.js#L111

As we cannot know in advance what that id is (it depends on each application in which this component is used), please add an optional prop `applicationId`, with a default value of 'app'.

The base component used for rendering the modal does not really need an application node https://github.com/davidtheclark/react-aria-modal/blob/master/src/react-aria-modal.js#L32

src/components/generic/Modal/Modal.jsx
```diff
@@ -0,0 +1,69 @@
...
+  onHide,
+  show,
+  showCloseButton,
+  ...props
+}) {
+  const mainClasses = {
+    [`${WEBEX_COMPONENTS_CLASS_PREFIX}-modal`]: true,
+    [className]: !!className,
+  };
+  const customId = `${WEBEX_COMPONENTS_CLASS_PREFIX}-${Math.floor((Math.random() * 100000000) + 1)}`;
```
cipak, line 33: call this `testId`, please, and use it for `htmlId` https://github.com/momentum-design/momentum-ui/blob/master/react/src/lib/Modal/index.js#L125

src/components/generic/Modal/Modal.jsx
```diff
@@ -0,0 +1,69 @@
...
+
+  return (
+    <Modal
+      applicationId={customId}
+      onHide={onHide}
+      show={show}
+      htmlId={customId}
+      className={classNames(mainClasses)}
```
cipak, line 41: maybe an alphabetic order?

src/components/generic/Modal/Modal.jsx
```diff
@@ -0,0 +1,69 @@
...
+      <ModalBody>
+        {children}
+      </ModalBody>
+    </Modal>
+  );
+}
+
+ModalComponent.propTypes = {
+  className: PropTypes.string,
+  children: PropTypes.arrayOf(PropTypes.shape({
```
cipak, line 55: use `children: PropTypes.node`

https://stackoverflow.com/a/50822607/210550

src/components/generic/Modal/Modal.jsx
```diff
@@ -0,0 +1,69 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import classNames from 'classnames';
+import {Modal, ModalBody, ModalHeader} from '@momentum-ui/react';
```
cipak, line 4: you can import momentum modal as MomentumModal to leave the Modal name for this component.

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/import#import_an_export_with_a_more_convenient_alias

src/components/_components.scss
```diff
@@ -11,3 +11,4 @@
 @import 'WebexMemberRoster/WebexMemberRoster';
 @import 'WebexRemoteMedia/WebexRemoteMedia';
 @import 'WebexSettings/WebexSettings';
+@import 'generic/generic';
```
cipak, line 14: use `components/generic/index.js` instead, so you can import it as `generic`

similar to `components/index.js` (which is not `components/components.js`).

src/components/generic/Modal/Modal.jsx
CernaianuMihai: Is it ok if I just set it's default value on true instead of false?

src/components/generic/Modal/Modal.jsx
```diff
@@ -0,0 +1,66 @@
...
+  className,
+  onClose,
+  title,
+  ...props
+}) {
+  const mainClasses = {
+    [`${WEBEX_COMPONENTS_CLASS_PREFIX}-modal`]: true,
+    [className]: !!className,
+  };
+  const testId = `${WEBEX_COMPONENTS_CLASS_PREFIX}-${Math.floor((Math.random() * 100000000) + 1)}`;
```
cipak, line 31: I would just use Math.random().toString().slice(2)

src/components/generic/Modal/Modal.jsx
```diff
@@ -0,0 +1,66 @@
...
+      <ModalBody>
+        {children}
+      </ModalBody>
+    </MomentumModal>
+  );
+}
+
+Modal.propTypes = {
+  applicationId: PropTypes.string,
+  children: PropTypes.node,
```
cipak, line 54: this should be required (hence, no default value)

cipak (approved):

PR #261, opened 2021-06-14
feat(MeetingsJsonAdapter): create camera selection control

karinasigartau0798:
Create video control for the case when the user has more than one device, add options on it depending on the available devices. create the test accordingly .
https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-238420


karinasigartau0798:
Duplicate

PR #298, opened 2021-07-16
feat(WebexComponents): configure winston logger

patricia0817:


src/components/logger.js
```diff
@@ -0,0 +1,20 @@
+import logform from 'logform';
+import createLogger from '../../node_modules/winston/dist/winston/create-logger';
+import Console from '../../node_modules/winston/dist/winston/transports/console';
```
cipak, line 3: @lalli-flores are you ok with this style of import?

The normal imports don't work because of webpack (storybook).

cipak: add a comment to explain why we're importing like this

cipak (approved):

PR #301, opened 2021-07-20
fix(UseMeetingDestination): chain createMeeting with getMeeting

patricia0817:
Chain dependent observables createMeeting and getMeeting

### DIFF
--- a/src/util.js
+++ b/src/util.js
@@ -1,3 +1,5 @@
+import {Observable} from 'rxjs';
+
 /**
  * Helper function for deep merge on objects.
  *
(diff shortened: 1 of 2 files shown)


patricia0817:
subTask link: https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-251271

src/util.js
```diff
@@ -16,6 +18,46 @@ export function deepMerge(dest, src) {
...
+/**
+ * Helper function for chaining dependent observables.
+ *
+ * @param {Function} factory - Function that creates an observable.
+ * @returns {Observable} observable
+ */
+export function chainWith(factory) {
+  return (source) => new Observable((subscriber) => {
+    let lastValue;
+    let secondSubscriber;
```
cipak, line 30: let's use a single variable for both subcriptions, `subscription`

src/util.js
```diff
@@ -16,6 +18,46 @@ export function deepMerge(dest, src) {
   }
 }
 
+/**
+ * Helper function for chaining dependent observables.
```
cipak, line 22: RXJS custom operator for chaining dependent observables.
Usage:
```js
obs.pipe(
  chainWith((lastMessage) => createDependentObservable(lastMessage),
);
```

src/util.js
```diff
@@ -16,6 +18,46 @@ export function deepMerge(dest, src) {
...
+ *
+ * @param {Function} factory - Function that creates an observable.
```
cipak, line 24: let's use `createDependentObservable` instead of `factory`

and use `{function(lastMessage): Observable}` as a type

double space instead of dash between param name and description

"Function that is passed the last message emitted by the source observable and returns a new observable"

cipak (approved):

PR #309, opened 2021-07-26
feat(WebexMeetingControl): display text on control buttons

CernaianuMihai:
https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-252931


src/components/WebexMeetingControl/WebexMeetingControl.scss
```diff
@@ -0,0 +1,3 @@
+.button-text:not(:first-child) {
```
cipak, line 1: please add a component container class, like the other components have

cipak (approved):

PR #314, opened 2021-08-04
fix(WebexComponentClasses): add support for prefixed optional css classes

BenjaminGolya:
Fix for missing classname prefix

### DIFF
--- a/src/components/helpers.js
+++ b/src/components/helpers.js
@@ -7,14 +7,24 @@ import {WEBEX_COMPONENTS_CLASS_PREFIX} from '../constants';
  * @param {object} props  Data passed to the component
  * @param {string} props.classBaseName  Component class base name
  * @param {string} props.userClassName  Custom CSS class to apply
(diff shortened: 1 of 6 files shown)


src/components/helpers.js
```diff
@@ -12,9 +12,18 @@ import {WEBEX_COMPONENTS_CLASS_PREFIX} from '../constants';
  */
 
 export default function webexComponentClasses(classBaseName, userClassName, otherClasses) {
```
cipak, line 14: `otherClasses` = {}

src/components/helpers.js
```diff
@@ -12,9 +12,18 @@ import {WEBEX_COMPONENTS_CLASS_PREFIX} from '../constants';
...
+  let otherClassesWithPrefix = {};
+
+  if (otherClasses) {
+    otherClassesWithPrefix = Object.fromEntries(
+      Object.entries(otherClasses)
```
cipak, line 19: or here: `otherClasses || {}` to get rid of the if

PR #316, opened 2021-08-04
feat(WebexMeeting): add control bar and roster to meeting component

CernaianuMihai:
https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-255715


src/components/WebexMeeting/WebexMeeting.jsx
```diff
@@ -48,14 +54,26 @@ export default function WebexMeeting({className, meetingID, style}) {
...
-      {showSettings && (
-        <Modal
-          onClose={() => adapter.meetingsAdapter.toggleSettings(ID)}
-          title="Settings"
-        >
-          <WebexSettings meetingID={ID} />
-        </Modal>
+      <div className="webex-meeting__meeting">
+        {meetingDisplay}
+        {showSettings && (
```
cipak, line 59: settings modal should only be shown when one of InMeeting or InterstitialMeeting is shown. So please add it to meetingDisplay.

same for roster

same for controls

the controls don't need a condition anymore, because the `meetingDisplay` has separate cases for no state or left meeting.

CernaianuMihai: done

src/components/WebexMeeting/WebexMeeting.jsx
```diff
@@ -42,21 +48,29 @@ export default function WebexMeeting({className, meetingID, style}) {
...
           : <WebexInterstitialMeeting meetingID={ID} />}
+        {showSettings && (
+          <Modal
+            onClose={() => adapter.meetingsAdapter.toggleSettings(ID)}
+            title="Settings"
+          >
+            <WebexSettings meetingID={ID} />
+          </Modal>
+        )}
+        {showRoster && isActive && (
```
cipak, line 59: isActive is surely true, look on line 48

src/components/WebexMeeting/WebexMeeting.jsx
```diff
@@ -42,21 +48,29 @@ export default function WebexMeeting({className, meetingID, style}) {
...
+          <div className="member-roster-container">
```
cipak, line 60: you can put a class "member-roster" directly on WebexMemberRoster now, before they didn't accept optional classes but now they do

same for control bar

src/components/WebexMeeting/WebexMeeting.scss
```diff
@@ -11,4 +10,17 @@
...
   background: var(--wxc-meeting-background);
+
+  .member-roster-container {
+    align-self: start;
+    width: 20rem;
+    padding: 0.625rem;
+  }
+
+  .control-bar-container {
+    position: absolute;
```
cipak, line 21: i don't think position absolute is required, the flex should take care of laying out this element correctly

CernaianuMihai: the flex would put the control bar side by side with the meeting and i can't add flex-direction: column because the roster has to be side by side with the meeting

src/components/WebexMeeting/WebexMeeting.jsx
CernaianuMihai: done

src/components/WebexMeeting/WebexMeeting.jsx
CernaianuMihai: done

CernaianuMihai: done

src/components/WebexMeeting/WebexMeeting.scss
cipak: ok, then we have to add a parent for the video component (inmeeting, interstitialmeeting) and roster, with horizontal flex

src/components/WebexMeeting/WebexMeeting.jsx
```diff
@@ -39,24 +45,30 @@ export default function WebexMeeting({className, meetingID, style}) {
...
     meetingDisplay = (
       <>
-        {isActive
-          ? <WebexInMeeting meetingID={ID} />
-          : <WebexInterstitialMeeting meetingID={ID} />}
+        <div className="body">
+          {isActive
+            ? <WebexInMeeting meetingID={ID} />
+            : <WebexInterstitialMeeting meetingID={ID} />}
+          {showSettings && (
```
cipak, line 52: settings don't need to belong to .body, please put them at the end, just after the control bar

cipak (approved):

cipak (approved):

PR #323, opened 2021-08-06
feat(WebexMeetingControl): add style for join and leave meeting button

BenjaminGolya:


BenjaminGolya:
Implementation already done in https://github.com/webex/components/pull/327

PR #328, opened 2021-08-11
feat(WebexMeetingControlsBar): show/hide buttons text depending on controls width

alinasuciu:
When controls width is bigger than its container the text should be hidden otherwise this should be shown.

https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-257320


src/components/WebexMeetingControlBar/WebexMeetingControlBar.scss
```diff
@@ -1,9 +1,15 @@
 .#{$WEBEX_COMPONENTS_CLASS_PREFIX}-meeting-control-bar {
+  width: 100%;
```
cipak, line 2: the width information should come from outside, is it needed here?

cipak (approved):

PR #337, opened 2021-08-13
feat(Modal): create modal component

alinasuciu:
Replace momentumUI modal component with a custom component

### DIFF
--- a/src/components/generic/Modal/Modal.jsx
+++ b/src/components/generic/Modal/Modal.jsx
@@ -1,14 +1,11 @@
 import React from 'react';
 import PropTypes from 'prop-types';
-import classNames from 'classnames';
-import {Modal as MomentumModal, ModalBody, ModalHeader} from '@momentum-ui/react';
-import {WEBEX_COMPONENTS_CLASS_PREFIX} from '../../../constants';
+import webexComponentClasses from '../../helpers';
 
 /**
  * Modal component
  *
  * @param {object} props  Data passed to the component
- * @param {string} [props.applicationId]  Application DOM id
  * @param {React.ReactNode[]} props.children  List of children
  * @param {string} [props.className]  Custom CSS class to apply
  * @param {Function} [props.onClose]  Callback when the modal is closed
@@ -17,39 +14,27 @@ import {WEBEX_COMPONENTS_CLASS_PREFIX} from '../../../constants';
  * @returns {object}  JSX of the element
  */
 export default function Modal({
-  applicationId,
   children,
   className,
   onClose,
   title,
-  ...props
 }) {
-  const mainClasses = {
-    [`${WEBEX_COMPONENTS_CLASS_PREFIX}-modal`]: true,
-    [className]: !!className,
-  };
-  const testId = `${WEBEX_COMPONENTS_CLASS_PREFIX}-${Math.random().toString().slice(2)}`;
+  const cssClasses = webexComponentClasses('modal', className);
 
   return (
-    <MomentumModal
-      applicationId={applicationId}
-      className={classNames(mainClasses)}
-      htmlId={testId}
-      onHide={onClose}
-      show
-      // eslint-disable-next-line react/jsx-props-no-spreading
-      {...props}
-    >
-      <ModalHeader headerLabel={title} showCloseButton={!!onClose} />
-      <ModalBody>
-        {children}
-      </ModalBody>
-    </MomentumModal>
+    <div className={cssClasses}>
+      <div className="modal-content">
+        <div className="modal-header">
+          {title && <h3 className="modal-title">{title}</h3>}
+          {onClose && <button type="button" onClick={onClose}>X</button>}
+        </div>
+        <div className="modal-body">{children}</div>
+      </div>
+    </div>
   );
 }
 
 Modal.propTypes = {
-  applicationId: PropTypes.string,
   children: PropTypes.node.isRequired,
   className: PropTypes.string,
   onClose: PropTypes.func,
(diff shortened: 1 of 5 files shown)


src/components/generic/Modal/Modal.scss
```diff
@@ -1,5 +1,34 @@
...
+  bottom: 0;
+  z-index: 9;
+  display: flex;
+  align-items: center;
+  justify-content: center;
+
+  .modal-content {
+    width: 500px;
+    height: 456px;
+    border: 1px solid var(--wxc-modal-border-color);
```
cipak, line 15: maybe just --wxc-modal-border, so someone could change the style of border and width as well

src/components/generic/Modal/Modal.scss
```diff
@@ -1,5 +1,34 @@
...
+  right: 0;
+  top: 0;
+  bottom: 0;
+  z-index: 9;
+  display: flex;
+  align-items: center;
+  justify-content: center;
+
+  .modal-content {
+    width: 500px;
```
cipak, line 13: the size of the component should come from outside, from the optional className prop

for this component, the optional class should be put on the .modal-content div, not on the big container.

src/components/generic/Modal/Modal.jsx
```diff
@@ -17,47 +14,34 @@ import {WEBEX_COMPONENTS_CLASS_PREFIX} from '../../../constants';
...
-      <ModalHeader headerLabel={title} showCloseButton={!!onClose} />
-      <ModalBody>
-        {children}
-      </ModalBody>
-    </MomentumModal>
+    <div className={cssClasses}>
+      <div className="modal-content">
+        <div className="modal-header">
+          <h3 className="modal-title">{title}</h3>
+          <button type="button" onClick={onClose}>Close</button>
```
cipak, line 29: we need an icon here, right?

src/themes/dark.scss
```diff
@@ -1,6 +1,10 @@
 .#{$WEBEX_COMPONENTS_CLASS_PREFIX}-theme-dark {
   --wxc-meeting-background: #{$wxc-gradation-webex-dark-primary};
   --wxc-text-color: #{$wxc-white-alpha-100};
+  // Modal
```
cipak, line 4: no need for comment but please leave a blank line

src/themes/light.scss
```diff
@@ -1,6 +1,10 @@
 .#{$WEBEX_COMPONENTS_CLASS_PREFIX}-theme-light {
   --wxc-meeting-background: #{$wxc-gradation-webex-light-primary};
   --wxc-text-color: #{$wxc-black-alpha-100};
+  // Modal
+  --wxc-modal-background: #{$wxc-white-alpha-100};
```
cipak, line 5: maybe put these component-related variables alphabetically, modal after icon

src/components/generic/Modal/Modal.jsx
alinasuciu: yes. we need the ghost button

src/components/generic/Modal/Modal.jsx
```diff
@@ -17,47 +14,34 @@ import {WEBEX_COMPONENTS_CLASS_PREFIX} from '../../../constants';
...
-      show
-      // eslint-disable-next-line react/jsx-props-no-spreading
-      {...props}
-    >
-      <ModalHeader headerLabel={title} showCloseButton={!!onClose} />
-      <ModalBody>
-        {children}
-      </ModalBody>
-    </MomentumModal>
+    <div className="modal">
```
cipak, line 25: let's call this modal-container and leave the name "modal" for the actual content (what is now named modal-content

it must also be prefixed, otherwise we're likely to get name clashes with other "modal" class defined by devs.

```jsx
<div className={webexComponentClasses('modal-container')}>

Comment (cipak) [reply src/components/generic/Modal/Modal.jsx:25]: if you think it's better to have '.modal' for container and '.modal-content' for content, I'm ok with it

File: src/components/generic/Modal/Modal.jsx [hunk 7]
```diff
@@ -17,47 +14,34 @@ import {WEBEX_COMPONENTS_CLASS_PREFIX} from '../../../constants';
...
+      <div className={cssClasses}>
+        <div className="modal-header">
+          <h3 className="modal-title">{title}</h3>
```
cipak, line 28: don't we have any modal without a title or without a close action? (eg: proceed without)

src/components/generic/Modal/Modal.scss
```diff
@@ -1,5 +1,32 @@
...
+  top: 0;
+  bottom: 0;
+  z-index: 9;
+  display: flex;
+  align-items: center;
+  justify-content: center;
+
+  .#{$WEBEX_COMPONENTS_CLASS_PREFIX}-modal-content {
+    border: var(--wxc-modal-border);
+    box-shadow: 0px 24px 48px rgba(0, 0, 0, 0.2), 0px 0px 1px rgba(0, 0, 0, 0.12);
```
cipak, line 14: we have a color code: $wxc-black-alpha-20

but maybe the whole box shadow should be a css variable to allow customization

src/components/generic/Modal/Modal.jsx
cipak: but it's important to have the top level class of a component prefixed. The rest can be not prefixed.

src/components/generic/Modal/Modal.scss
cipak: and use rems

cipak (approved):

PR #350, opened 2021-08-19
feat(WebexAvatar): Add themed avatar component

bogdanmaier28:
Story: https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-258300


src/components/WebexAvatar/WebexAvatar.jsx
```diff
@@ -26,16 +48,28 @@ export default function WebexAvatar({
   const type = displayStatus ? status : null;
 
   const cssClasses = webexComponentClasses('avatar', className);
+  const oldClasses = webexComponentClasses('old-avatar', className);
```
cipak, line 51: the old avatar is never used with an optional class (search for "<Avatar" and "<WebexAvatar" in the project), and it also did not have a css file before, so I guess, to make it simple to remove the old implementation, it would be ok to just add `<Avatar className="wxc-old-avatar"` below

src/components/WebexAvatar/WebexAvatar.jsx
```diff
@@ -26,16 +48,28 @@ export default function WebexAvatar({
...
+
+  let AvatarComponent = renderAvatarContent(avatar, type);
+
+  if (!avatar || !displayName) {
```
cipak, line 55: it's a bit strange to call renderAvatarContent and then cancel the result here.

I would suggest:
```js
if (!avatar && !displayName) {
  content = ??loading??;
} else if (!avatar && displayName === ' ') {
  content = ??error??;
} else {
   content = <>
     <img ... />
     {displayStatus && <div status-icon>
        ...
     </div>}
  </>;
}
```

if there is no icon name for that status, that's an error that should be reported (logged).

but if it's ok not to have an icon for a particular status, then you can check here for the icon name as well `{displayStatus && iconName && <div>...</div>}`

src/components/WebexAvatar/WebexAvatar.jsx
```diff
@@ -26,16 +48,28 @@ export default function WebexAvatar({
   const type = displayStatus ? status : null;
```
cipak, line 48: status, type, icon name :)

I would suggest:
```js
const iconName = status;
```
this is easy now but might change in the future, I think a map will be needed soon from person statuses to icon names. At least it's obvious that we are using the status string as an icon name verbatim.

src/components/WebexAvatar/WebexAvatar.jsx
```diff
@@ -1,10 +1,32 @@
...
 import {usePerson} from '../hooks';
 
+/**
+ * Renders the person image and status icon.
+ *
+ * @param {string} avatar  URL of the image.
+ * @param {string|null} status  Status of the person.
+ * @returns {object} JSX of the component
+ */
+function renderAvatarContent(avatar, status) {
```
cipak, line 15: this function seems to complicate things a bit, try to see if the code is clearer without it

src/components/WebexAvatar/WebexAvatar.jsx
```diff
@@ -26,16 +48,28 @@ export default function WebexAvatar({
   const type = displayStatus ? status : null;
 
   const cssClasses = webexComponentClasses('avatar', className);
+  const oldClasses = webexComponentClasses('old-avatar', className);
+
+  let AvatarComponent = renderAvatarContent(avatar, type);
+
+  if (!avatar || !displayName) {
+    AvatarComponent = <> </>;
```
cipak, line 56: it's just content (jsx), a component should be a function

src/components/WebexAvatar/WebexAvatar.scss
```diff
@@ -0,0 +1,37 @@
+.#{$WEBEX_COMPONENTS_CLASS_PREFIX}-avatar {
+  display: flex;
+  align-items: center;
+
+  font-size: .875rem;
```
cipak, line 5: does the text scale with the avatar?

src/components/WebexAvatar/WebexAvatar.scss
```diff
@@ -0,0 +1,37 @@
...
+    left: 62.5%;
+    right: -12.5%;
+    top: 56.25%;
+    bottom: -6.25%;
+
+    border-radius: 50%;
+
+    background: var(--wxc-icon--background-color);
+
+    .#{$WEBEX_COMPONENTS_CLASS_PREFIX}-icon {
```
cipak, line 32: please send an optional class name to the Icon component for that, instead of targeting its base class

src/components/WebexAvatar/WebexAvatar.jsx
```diff
@@ -24,18 +34,51 @@ export default function WebexAvatar({
 }) {
   const {avatar, displayName, status} = usePerson(personID);
   const type = displayStatus ? status : null;
-
+  const iconName = status;
```
cipak, line 37: do icons really have the same name as statuses, or did you name them as such? I would prefer to keep the figma icon names if possible, and create a map here, eg `statusIcons`.

the status class names, on the other hand, can be easily computed instead of using a map:
```js
statusClassName = 'avatar-status-' + status;
```

src/components/WebexAvatar/WebexAvatar.jsx
```diff
@@ -24,18 +34,51 @@ export default function WebexAvatar({
...
   const cssClasses = webexComponentClasses('avatar', className);
 
+  let content;
+  let iconClassName;
+
+  if (status) {
```
cipak, line 43: you can move this inside the `else` on line 55

src/components/WebexAvatar/WebexAvatar.jsx
```diff
@@ -24,18 +34,48 @@ export default function WebexAvatar({
 }) {
   const {avatar, displayName, status} = usePerson(personID);
   const type = displayStatus ? status : null;
-
+  const iconName = `avatar-status-${status}`;
```
cipak, line 37: I think we should leave the icon names as they are in figma, and create a map from status names to icon names. The icons are not exclusive to avatar usage.

The status class names can simply be computed based on status, eg:
```js
const statusClassName = `status-${status}`;
```

src/themes/dark.scss
```diff
@@ -62,6 +67,21 @@
   --wxc-button--join--disabled--color: #{$wxc-white-alpha-40};
 }
 
+.#{$WEBEX_COMPONENTS_CLASS_PREFIX}-theme-dark {
+  .avatar-icon-active {
```
cipak, line 71: these belong to the avatar scss file

src/themes/dark.scss
```diff
@@ -8,11 +8,16 @@
   --wxc-banner-background: #{$wxc-gradation-dark-transparent};
   --wxc-banner-text-color: #{$wxc-white-alpha-95};
 
+  --wxc-icon--background-color: #{$wxc-black-alpha-100};
   --wxc-icon--color: #{$wxc-white-alpha-100};
   --wxc-icon--active-color: #{$wxc-green-50};
   --wxc-icon--muted-color: #{$wxc-red-50};
   --wxc-icon--error--gradient-start-color: #{$wxc-pink-50};
   --wxc-icon--error--gradient-stop-color: #{$wxc-red-60};
+  --wxc-icon--avatar-active-color: #{$wxc-green-40};
```
cipak, line 17: let's keep the status colors separated from the icons, and define generic variables that only relate to status:
```scss
--wxc-person-status--active--color
--wxc-person-status--dnd--color
...
```

cipak: they should go at the top with the other generic variables (text-color, video-background)

src/components/WebexAvatar/WebexAvatar.jsx
```diff
@@ -24,18 +44,48 @@ export default function WebexAvatar({
...
   const {avatar, displayName, status} = usePerson(personID);
   const type = displayStatus ? status : null;
-
+  const iconName = statusIcons[status];
   const cssClasses = webexComponentClasses('avatar', className);
 
+  let content;
+
+  if (!avatar && !displayName && !status) {
+    content = <> </>;
```
cipak, line 53:
```js
content = <> </>; // loading
```

src/components/WebexAvatar/WebexAvatar.jsx
```diff
@@ -24,18 +44,48 @@ export default function WebexAvatar({
...
+  } else if (!avatar && displayName === ' ') {
+    content = <> </>;
```
cipak, line 55:
```js
content = <> </>; // error
```

just to let anyone reading this code why we have these conditions.

src/components/WebexAvatar/WebexAvatar.jsx
```diff
@@ -24,18 +44,48 @@ export default function WebexAvatar({
...
+  } else {
+    const iconClassName = statusClassNames[status];
```
cipak, line 57: you can also move const iconName here

cipak (approved):

PR #352, opened 2021-08-20
feat(WebexComponents): theme WebexMember component

patricia0817:


src/components/WebexMember/WebexMember.scss
```diff
@@ -1,33 +1,33 @@
...
     .guest {
       color: $md-gray-70;
     }
   }
 
-  .subtext, .roles {
+  .roles {
     font-size: 0.875rem;
-    color: $md-gray-70;
+    color: var(--wxc-member--role--color);
```
cipak, line 25: isn't this the `--wxc-secondary-text-color`? I would prefer not to have any variables dedicated to a specific webex* component. It's ok to have variables dedicated to generic components, though.

PR #353, opened 2021-08-20
feat(WebexSettings): add theme colors and typography on settings components

lungancatalin:


src/components/WebexSettings/WebexSettings.scss
```diff
@@ -1,2 +1,14 @@
 @import 'Select';
 @import 'Tabs';
+
+.#{$WEBEX_COMPONENTS_CLASS_PREFIX}-settings {
+  background-color: var(--wxc-modal-background);
```
cipak, line 5: modal-background must be set by the modal component, this component is just the content of the modal and it can be displayed in other contexts not only in a modal

src/components/WebexSettings/WebexSettings.scss
```diff
@@ -1,2 +1,14 @@
...
+
+  .#{$WEBEX_COMPONENTS_CLASS_PREFIX}-audio-settings,
```
cipak, line 7: each component needs to have its own css file

src/components/WebexAudioSettings/WebexAudioSettings.scss
```diff
@@ -0,0 +1,8 @@
+.#{$WEBEX_COMPONENTS_CLASS_PREFIX}-audio-settings {
+  h5 {
```
cipak, line 2: please also add the margins, h5 elements have their own margins that we might need to override.

cipak: and you can keep using h2 if everything is specified

src/components/_components.scss
```diff
@@ -15,4 +15,5 @@
 @import 'WebexRemoteMedia/WebexRemoteMedia';
 @import 'WebexSettings/WebexSettings';
 @import 'WebexVideoSettings/WebexVideoSettings';
+@import 'WebexAudioSettings/WebexAudioSettings';
```
cipak, line 18: keep alpha order

src/components/WebexVideoSettings/WebexVideoSettings.scss
```diff
@@ -1,4 +1,12 @@
 .#{$WEBEX_COMPONENTS_CLASS_PREFIX}-video-settings {
+  h5 {
```
cipak, line 2: this looks very similar to the Participants title (only difference is that that one is bold, this one should be medium according to figma).

To avoid duplication, can you create a simple generic component named Title to be used as such:
```jsx
<Title>Speaker</Title>
```

we could then use it for participants as well (in a different PR):
```jsx
<Title bold>Participants ({count})</Title>
```

PR #354, opened 2021-08-20
feat(WebexMeeting): add scrollbar on roster

lungancatalin:
https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-259905


src/components/WebexMemberRoster/WebexMemberRoster.scss
```diff
@@ -23,4 +26,8 @@
     line-height: 1.5rem;
     color: var(--wxc-text-color);
   }
+
+  .members {
+    overflow-y: auto;
```
cipak, line 31: + flex: 1 so that it occupies all available space. Also, please check if the scrollbar appears if there are more members than the available space, if not, try min-height: 0

src/components/WebexMemberRoster/WebexMemberRoster.scss
```diff
@@ -23,4 +26,9 @@
...
+    flex: 1;
```
cipak, line 32: in order to make this visible in storybook, please change the roster story from:
```js
export default {
  title: 'Platform/Webex Member Roster',
  component: WebexMemberRoster,
  decorators: [(Story) => <div style={{width: '20rem'}}><Story /></div>],
};

const Template = (args) => <WebexMemberRoster {...args} />;
```
to
```js
export default {
  title: 'Platform/Webex Member Roster',
  component: WebexMemberRoster,
};

const Template = (args) => {
  const style = {
    width: '20rem',
    maxHeight: '100%',
    ...args.style,
  };
  const props = {...args, style};

  return <WebexMemberRoster {...props} />;
};
```
we're adding a max height so that the roster has limited vertical space. Change the Meeting story to use meeting2 and make the last user of meeting2 to be outside of the meeting (inMeeting: false). The decorator is no longer needed, the roster component can now accept a style prop itself.

cipak (approved):

PR #355, opened 2021-08-20
feat(WebexMeetingControl): add unmuted theme icon

alinasuciu:
Add unmuted icon as per theme design.

https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-259891


src/components/WebexMeeting/__snapshots__/WebexMeeting.stories.storyshot
```diff
@@ -129,12 +129,20 @@ Array [
...
+                fill="var(--wxc-icon--color)"
```
cipak, line 133: remove this, the icon should take its color from the parent/context, the svg already has `fill: currentColor` from the `.wxc-icon` class

cipak (approved):

cipak (approved):

PR #357, opened 2021-08-23
feat(WebexRemoteMedia): crop media inside the meeting

CernaianuMihai:
https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-260359

### DIFF
--- a/src/components/WebexInMeeting/WebexInMeeting.scss
+++ b/src/components/WebexInMeeting/WebexInMeeting.scss
@@ -6,11 +6,20 @@ $shadow-color: rgba(0, 0, 0, 0.2);
 .#{$WEBEX_COMPONENTS_CLASS_PREFIX}-in-meeting {
   position: relative;
   display: flex;
+  justify-content: center;
 
   &.remote-sharing {
     .local-media-in-meeting {
       display: none;
     }
+  }
+
+  .media-container {
+    position: relative;
+    display: flex;
+    flex-direction: column;
+    flex: 1;
+    align-self: stretch;
   }
 
   .share-banner {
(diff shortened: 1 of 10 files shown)


src/styles/_mixins.scss
```diff
@@ -70,3 +78,15 @@ $shadow-color: rgba(0, 0, 0, 0.2);
 
   z-index: 1;
 }
+
+@mixin tablet {
+  @media (min-width: #{$tablet-width}) {
```
cipak, line 83: I don't think we can use media queries, our components can be much smaller than the screen. We need to use useElementDimensions to take the width and add classes conditionally.

CernaianuMihai: I reversed these changes.

src/components/WebexInMeeting/WebexInMeeting.scss
```diff
@@ -6,13 +6,19 @@ $shadow-color: rgba(0, 0, 0, 0.2);
...
 
   &.remote-sharing {
     .local-media-in-meeting {
       display: none;
     }
   }
 
+  .media-container {
+    position: relative;
+    flex-direction: column;
```
cipak, line 19: add `display: flex` before `flex-direction`

src/components/WebexInMeeting/WebexInMeeting.scss
```diff
@@ -21,6 +27,7 @@ $shadow-color: rgba(0, 0, 0, 0.2);
   }
 
   .remote-media-in-meeting {
+    height: 100%;
```
cipak, line 30: just flex: 1 is enough if you add display: flex on parent

src/components/WebexInMeeting/WebexInMeeting.scss
CernaianuMihai: done

src/components/WebexInMeeting/WebexInMeeting.scss
CernaianuMihai: done

cipak (approved):

PR #361, opened 2021-08-23
feat(Modal): add ghost button to the modal

lungancatalin:
https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-260267

### DIFF
--- a/src/themes/dark.scss
+++ b/src/themes/dark.scss
@@ -70,4 +70,5 @@
   .wxc-old-icon { display: none; }
   .wxc-icon { display: block; }
   .wxc-in-meeting { background: transparent; }
+  .wxc-modal .old-close-button { display: none; }
 }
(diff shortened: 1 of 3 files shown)


src/themes/dark.scss
```diff
@@ -69,4 +69,5 @@
   .wxc-button { display: flex; }
   .wxc-old-icon { display: none; }
   .wxc-icon { display: block; }
+  .modal-header .old-close-icon { display: none; }
```
cipak, line 72: `.wxc-modal .old-close-icon`

all css rules must have at least one prefixed (`.wxc-`) css class

cipak: .old-close-button would be a better name

cipak: you also have to hide the new close button on legacy, but if it doesn't show because the icon does not show, it's fine

cipak (approved):

PR #362, opened 2021-08-23
feat(WebexComponents): get the icon size based on the name of the icon

patricia0817:
Task link: https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-260334

### DIFF
--- a/src/components/generic/Icon/Icon.jsx
+++ b/src/components/generic/Icon/Icon.jsx
@@ -32,7 +32,7 @@ const icons = {
  *
  * @param {object} props  Data passed to the component
  * @param {string} props.name  Icon name
- * @param {number} [props.size=24]  Icon width and height
+ * @param {number} [props.size]  Icon width and height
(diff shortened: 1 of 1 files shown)


src/components/generic/Icon/Icon.jsx
```diff
@@ -63,7 +69,7 @@ Icon.propTypes = {
 };
 
 Icon.defaultProps = {
-  size: 24,
+  size: 0,
```
cipak, line 72: undefined

src/components/generic/Icon/Icon.jsx
```diff
@@ -43,12 +43,18 @@ export default function Icon({
   name, size, className, indicator, style,
 }) {
   const baseName = name.split('_')[0];
+  const baseSize = name.split('_')[1];
```
cipak, line 46: in a single shot:
```js
const [baseName, baseSize] = name.split('_');

cipak (approved):

PR #376, opened 2021-08-27
feat(icons): disabled state for icons

karinasigartau0798:
Task link: https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-261627

PR #378, opened 2021-08-30
fix(Icons): update muted/unmuted microphone and camera icons to use currentColor

BenjaminGolya:


BenjaminGolya:
Close because the implementation has changed, and already done in: https://github.com/webex/components/pull/376

PR #384, opened 2021-08-31
fix(WebexAvatar): resize avatar and add margins to it

CoroDaniel:


src/components/WebexMember/WebexMember.scss
```diff
@@ -3,6 +3,13 @@
   align-items: center;
   color: var(--wxc-text-color);
 
+  .avatar{
+    height: 2rem;
```
cipak, line 7: `Avatar` has `size`, use that instead

src/components/WebexMember/WebexMember.scss
```diff
@@ -3,6 +3,13 @@
   align-items: center;
   color: var(--wxc-text-color);
 
+  .avatar{
```
cipak, line 6: space before curly brace

src/components/WebexMember/WebexMember.scss
```diff
@@ -3,6 +3,13 @@
...
+    height: 2rem;
+    width: 2rem;
+    margin-top: 0.5rem;
```
cipak, line 9: this margin should come from the roster on the whole `WebexMember` component

src/components/WebexMember/WebexMember.scss
CoroDaniel: Avatar hasn't size yet

cipak (approved):

PR #394, opened 2021-09-02
fix(themes): add the correct path for loading logo

lungancatalin:


src/themes/legacy.scss
```diff
@@ -1,6 +1,6 @@
 :root {
   --wxc-meeting-background: transparent;
-  --wxc-loading-logo: url(../../src/themes/dark/loading.gif);
+  --wxc-loading-logo: url(../themes/dark/loading.gif);
```
cipak, line 3: this won't look good on legacy

cipak (approved):

PR #419, opened 2021-09-09
fix(Icons): use correct color for accents

CoroDaniel:


src/components/icons/_icons.scss
```diff
@@ -5,7 +5,7 @@
   fill: currentColor;
 
   .accent {
-    fill: var(--wxc-icon--active-color);
+    fill: var(--wxc-unmuted-color);
```
cipak, line 8: please use --wxc-icon--accent-color, it should have the same value as --wxc-unmuted-color

src/themes/dark.scss
```diff
@@ -42,6 +42,7 @@
   --wxc-icon--color: #{$wxc-white-alpha-100};
   --wxc-icon--active-color: #{$wxc-green-50};
   --wxc-icon--muted-color: #{$wxc-red-50};
+  --wxc-icon--accent-color: #{$wxc-green-40};
```
cipak, line 45: "active" should be renamed to "accent".

commit: fix(Icons): use correct color for accents

cipak (approved):

PR #421, opened 2021-09-09
feat(styles): reset browser default styles

alinasuciu:
This is needed because once with the momentum removal we need our own reset file.

https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-265508


lalli-flores (commented):
@alinasuciu where did you get this CSS? should be use something more standard [like normalize](https://github.com/necolas/normalize.css/#normalizecss)?

lalli-flores:
We talked about this further and decided that it's better to reset as-needed within each component instead of a global reset. A global reset will leak and affect whatever page the widget is placed in.

PR #422, opened 2021-09-09
fix(OptionsList): selecting a dropdown option moves options below

alinasuciu:
Selecting a dropdown option moves options below 2 pixels.

https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-265451

### DIFF
--- a/src/components/generic/OptionsList/OptionsList.scss
+++ b/src/components/generic/OptionsList/OptionsList.scss
@@ -17,6 +17,7 @@
       align-items: center;
       cursor: pointer;
       color: var(--wxc-text-color);
+      border: 1px solid transparent;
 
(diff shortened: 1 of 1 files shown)


src/components/generic/OptionsList/OptionsList.scss
```diff
@@ -17,6 +17,7 @@
       align-items: center;
       cursor: pointer;
       color: var(--wxc-text-color);
+      border: 1px solid transparent;
```
cipak, line 20: you can remove this pixel from the vertical padding to preserve the total height pf the element

lalli-flores:
Replaced by https://github.com/webex/components/pull/439

PR #424, opened 2021-09-10
fix(Select): make options list the same width as select box

lungancatalin:


src/components/generic/Select/Select.scss
```diff
@@ -64,10 +64,12 @@
   }
 
   .options-list {
-    width: 100%;
     position: absolute;
     z-index: 1;
     top: calc(100% + 0.125rem);
```
cipak, line 69: is box-sizing still needed?

lungancatalin: Not. You're right. Works without it. :)

cipak (approved):

PR #432, opened 2021-09-13
fix(MeetingsJsonAdapter): don't enable camera and microphone on switch

BenjaminGolya:


src/adapters/MeetingsJSONAdapter.js
```diff
@@ -558,11 +558,13 @@ export default class MeetingsJSONAdapter extends MeetingsAdapter {
    */
   async switchCamera(ID, cameraID) {
     await this.updateMeeting(ID, async (meeting) => {
-      this.stopStream(meeting.localVideo.stream);
+      if (meeting.localVideo.stream) this.stopStream(meeting.localVideo.stream);
 
       return {
         localVideo: {
-          stream: await this.getStream({video: {deviceId: {exact: cameraID}}}),
+          stream: meeting.localVideo.stream ? await this.getStream({
```
cipak, line 565: I think this is more complicated, because with these changes the user won't be able to see a preview and switch camera if it is already stopped. We need another place to store the preview stream and the preferred device (when camera is stopped).

lalli-flores (commented):
You're on the right track but here is how I would like to see it implemented:
`showSettings` should become an object => `settings: {}`

`settings` object should include:
```
{
  visible: true/false, // This replaces the current showSettings
  preview: {
    video: {}  // MediaStream of the video preview 
 }
```
This way, `WebexLocalMedia` doesn't have to be modified because you can check `settings.visible` to determine whether you need to show a preview or the `localVideo`/`localShare`

I would approach this in 2 steps:
1. Update the showSettings to settings (1 commit) - this makes sure that things don't break current functionality
2. Enable preview on settings component (1 commit) - similar to have you have done

Things to consider/test for:
* Preview is not muted if my video is muted when I click settings
* Changing preview doesn't change `localVideo` UNTIL I exit settings modal
* if my video was muted it needs to stay muted even if I change it
* if my video was unmuted, it needs to stay unmuted even if I change it
* If I don't make any changes, `localVideo` should not be affected

PR #433, opened 2021-09-13
fix(OptionsList): show "selected" icon in its own column

CoroDaniel:
https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-267148

### DIFF
--- a/src/components/generic/OptionsList/OptionsList.jsx
+++ b/src/components/generic/OptionsList/OptionsList.jsx
@@ -1,5 +1,6 @@
 import React from 'react';
 import PropTypes from 'prop-types';
+import classNames from 'classnames';
 
 import Icon from '../Icon/Icon';
 import webexComponentClasses from '../../helpers';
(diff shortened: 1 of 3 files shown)


src/components/generic/OptionsList/OptionsList.jsx
```diff
@@ -39,6 +39,7 @@ export default function OptionsList({
             {option.icon && <Icon name={option.icon} />}
             <span className="label">{option.label}</span>
             {selected && selected === option.value && <Icon className="check" size={16} name="check" />}
+            {selected && selected !== option.value && <div className="uncheck" />}
```
cipak, line 42: add a `check-invisible` class to the icon with `visibility: hidden`

cipak: commit: fix(OL): show "selected" icon in its own column

cipak (approved):

PR #442, opened 2021-09-17
fix(MeetingsJsonAdapter): use real streams for local media on storybook from the first interaction

karinasigartau0798:


src/adapters/MeetingsJSONAdapter.js
```diff
@@ -462,6 +467,17 @@ export default class MeetingsJSONAdapter extends MeetingsAdapter {
     }
   }
 
+  /**
+   * A function that checks whether or not a Mediastream object is empty (no tracks).
+   *
+   * @param {MediaStream} stream  Media stream object
```
cipak, line 473: `{MediaStream|null} [stream]`, because it's optional and it can also take null

src/adapters/MeetingsJSONAdapter.js
```diff
@@ -462,6 +467,17 @@ export default class MeetingsJSONAdapter extends MeetingsAdapter {
...
+   * @returns {boolean} Whether or not the Mediastream object is empty
```
cipak, line 474: True if the received param is a MediaStream object with no tracks

PR #443, opened 2021-09-20
fix(WebexMeeting): resize dark and light loading logos

lungancatalin:
https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-267914

### DIFF
--- a/src/components/WebexMeeting/WebexMeeting.scss
+++ b/src/components/WebexMeeting/WebexMeeting.scss
@@ -27,7 +27,7 @@
   }
 
   .loading-logo {
-    background: var(--wxc-loading-logo) no-repeat center / 6.5rem 4.627rem;
+    background: var(--wxc-loading-logo) no-repeat center / var(--wxc-loading-logo-size);
     position: absolute;
     top: 0;
     left: 0;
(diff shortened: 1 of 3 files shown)


src/components/WebexMeeting/WebexMeeting.scss
```diff
@@ -27,7 +27,7 @@
   }
 
   .loading-logo {
-    background: var(--wxc-loading-logo) no-repeat center / 6.5rem 4.627rem;
+    background: var(--wxc-loading-logo) no-repeat center / 11.6875rem;
```
lalli-flores, line 30: Can you make this a variable or formulate as to how you arrived to this number?

lungancatalin: Hi @lalli-flores,

I received in Jira, from Hannah, the correct dimensions for loading the logo, which should be 187 * 187 px. I transformed these pixels into rem 187/16 = 11.6875 and used the size in rem. That's it. I will add a new variable for that dimension and add it in both themes. Thank you.

lalli-flores: Thanks for the clarification @lungancatalin. Someone without access or knowledge of the Figma spec would look at the number and wonder where this seemingly random number came from.

lalli-flores (approved):

PR #445, opened 2021-09-20
feat(WebexAudioSettings): add message for no speakers in Firefox and …

lungancatalin:
…Safari browser

The project uses the setSinkId() method to allows you to switch between speakers but this method it seems to be quite restricted for Firefox and Safari browser. Check here in order to see the browser compatibility: https://developer.mozilla.org/en-US/docs/Web/API/HTMLMediaElement/setSinkId

As a solution, when a user is in Firefox or Safari, we can display a disabled dropdown menu for speaker with a message: 'No available speakers'

image

Jira: https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-267752

### DIFF
--- a/src/components/WebexMeetingControl/WebexMeetingControl.jsx
+++ b/src/components/WebexMeetingControl/WebexMeetingControl.jsx
@@ -68,10 +68,22 @@ function renderButton(action, display, style, showText, asItem) {
  * @param {Function} action  Adapter control callback
  * @param {object} display  Display data of the control
  * @param {object} style  Custom style to apply
+ * @param {string} type  Name of the control as defined in adapter
  * @returns {object} JSX of the component
  */
-function renderDropdown(action, display, style) {
-  const {options, noOptionsMessage, selected} = display;
+function renderDropdown(action, display, style, type) {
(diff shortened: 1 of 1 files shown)


src/components/WebexMeetingControl/WebexMeetingControl.jsx
```diff
@@ -68,19 +68,31 @@ function renderButton(action, display, style, showText, asItem) {
...
  * @param {object} style  Custom style to apply
+ * @param {string} type  Name of the control as defined in adapter
  * @returns {object} JSX of the component
  */
-function renderDropdown(action, display, style) {
-  const {options, noOptionsMessage, selected} = display;
+function renderDropdown(action, display, style, type) {
+  const {noOptionsMessage} = display;
+  let {options, selected} = display;
+  // The browser api setSinkId() does not work properly on Firefox and Safari and that's why we need to overwrite/reset received options on the control.
```
alinasuciu, line 77: // The browser api setSinkId() does not work properly on Firefox and Safari browsers so we need to treat them separately by resetting the options we receive with a default no media label.

PR #448, opened 2021-09-20
fix(WebexSettings): correct the wrong font weights in settings modal

alinasuciu:
Correct the wrong font weights in settings modal.

https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-267927


src/components/WebexVideoSettings/WebexVideoSettings.scss
```diff
@@ -2,6 +2,10 @@
   display: flex;
   flex-direction: column;
 
+  .title {
+    font-family: $brand-font-medium;
```
lalli-flores, line 6: Do you need to specify `font-size: 1rem`??

alinasuciu: Sorry, I don't understand the question. It's not specified.

PR #452, opened 2021-09-21
feat(WebexComponents): add meeting control tooltips

patricia0817:


src/components/generic/Tooltip/Tooltip.scss
```diff
@@ -0,0 +1,44 @@
+.#{$WEBEX_COMPONENTS_CLASS_PREFIX}-tooltip-container {
+  position: absolute;
+  width: 110px;
```
alinasuciu, line 3: rem

src/components/generic/Button/Button.jsx
```diff
@@ -20,16 +22,39 @@ export default function Button({
...
   isDisabled,
   onClick,
-  title,
+  tooltip,
   style,
   type,
 }) {
   const cssClasses = webexComponentClasses('button', className, {[`-${type}`]: true});
+  const [isTooltipVisible, setTooltipVisible] = useState(false);
+  const btnRef = useRef(null);
```
alinasuciu, line 31: maybe buttonRef, would be more suggestive

lalli-flores:
@patricia0817 let me think about your implementation. It works, but it doesn't look great when there is long lines of text

src/components/generic/Portal/Portal.jsx
```diff
@@ -0,0 +1,23 @@
...
+/**
+ * Provides a first-class way to render children into a DOM node that exist
+ * outside the DOM hierarchy of the parent component
+ *
+ * @param {object} props  Data passed to the component
+ * @param {React.ReactNode[]} props.children  List of children
+ * @returns {object}  JSX of the element
+ */
+const Portal = ({children}) => {
+  const mount = document.getElementById('portal-root');
```
cipak, line 14: this means that devs must add this id to their dom... can we default to the `body` element?

src/components/generic/Tooltip/Tooltip.jsx
```diff
@@ -0,0 +1,76 @@
...
+  target,
+}) => {
+  const cssClasses = webexComponentClasses('tooltip-container', className);
+  const [tooltipVisible, setTooltipVisible] = useState(false);
+  const rect = target.getBoundingClientRect();
+  const coords = {
+    left: rect.x + rect.width / 2, // add half the width of the button for centering
+    top: rect.y + window.scrollY, // add scrollY offset, as soon as getBountingClientRect takes on screen coords
+  };
+  const handleMouseEnter = () => {
```
cipak, line 29: you can create these functions inside the `useEffect` block, so that they are not created when not needed

src/components/generic/Tooltip/Tooltip.jsx
```diff
@@ -0,0 +1,76 @@
...
+    target.addEventListener('mouseleave', handleMouseLeave);
+
+    return () => {
+      target.removeEventListener('mouseenter', handleMouseEnter);
+      target.removeEventListener('mouseleave', handleMouseLeave);
+    };
+  }, [target]);
+
+  return (
+    <Portal>
```
cipak, line 47: can the portal be too under the condition?

src/components/generic/Button/Button.jsx
```diff
@@ -20,16 +21,31 @@ export default function Button({
...
+        className={cssClasses}
+        disabled={isDisabled}
+        type="button"
+        onClick={onClick}
+        ref={buttonRef}
+        style={style}
+      >
+        {children}
+      </button>
+      {(buttonRef.current && tooltip) && (
```
cipak, line 43: the `Tooltip` component could handle a falsey target, to make its use easier (and more robust)

PR #457, opened 2021-09-21
fix(WebexMember): set higher font color contrast for "guest" role

CoroDaniel:
Updates roster guest font color so it is complaint with contrast accessibility requirements

### DIFF
--- a/src/components/WebexMember/WebexMember.scss
+++ b/src/components/WebexMember/WebexMember.scss
@@ -24,7 +24,7 @@
     margin-top: 0.063rem;
 
     .guest {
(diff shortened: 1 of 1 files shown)


lalli-flores (changes requested):
Can you update the commit message so it more accurate reflects the change?
```
fix(WebexMember): set higher font color contrast for "guest" role

CoroDaniel:
> Can you update the commit message so it more accurate reflects the change?
> 
> ```
> fix(WebexMember): set higher font color contrast for "guest" role
> ```

done

lalli-flores (approved):

PR #461, opened 2021-09-23
fix(icons): remove indicator prop from icons because it is not used anywhere

karinasigartau0798:
https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-269354


alinasuciu (approved):

lalli-flores (approved):
Good catch!

PR #466, opened 2021-09-27
feat(WebexAudioSettings): displays the error for no speakers in Firef…

lungancatalin:
…ox and Safari browser


alinasuciu (approved):

src/components/WebexNoMedia/WebexNoMedia.jsx
```diff
@@ -32,6 +32,12 @@ const SCREENS = {
     ),
     /* eslint-enable react/jsx-one-expression-per-line */
   },
+  speaker: {
+    icon: '',
+    title: 'Speaker can\'t be changed on your browser',
+    /* eslint-disable react/jsx-one-expression-per-line */
```
alinasuciu, line 38: if the message is not on multiple lines we don't need to disable the linter

alinasuciu (approved):

src/components/WebexAudioSettings/WebexAudioSettings.jsx
```diff
@@ -18,14 +18,21 @@ import WebexNoMedia from '../WebexNoMedia/WebexNoMedia';
 export default function WebexAudioSettings({className, meetingID, style}) {
   const cssClasses = webexComponentClasses('audio-settings', className);
   const [, display] = useMeetingControl('switch-microphone', meetingID);
+  // The browser api setSinkId() does not work properly on Firefox and Safari browsers.
+  // https://developer.mozilla.org/en-US/docs/Web/API/HTMLMediaElement/setSinkId
+  const isFirefox = navigator.userAgent.indexOf('Firefox') !== -1; // detect Firefox browser
+  const isSafari = navigator.userAgent.indexOf('Safari') !== -1 && navigator.userAgent.indexOf('Chrome') === -1; // detect Safari browser
```
lalli-flores, line 24: I think you should move these into the [helpers module](https://github.com/webex/components/blob/master/src/components/helpers.js), they could be useful for other components in the future.
You would make them into functions instead, e.g. `isFirefox()` `isSafari()` and they would return a boolean

src/components/WebexAudioSettings/WebexAudioSettings.jsx
```diff
@@ -18,14 +18,21 @@ import WebexNoMedia from '../WebexNoMedia/WebexNoMedia';
 export default function WebexAudioSettings({className, meetingID, style}) {
   const cssClasses = webexComponentClasses('audio-settings', className);
   const [, display] = useMeetingControl('switch-microphone', meetingID);
+  // The browser api setSinkId() does not work properly on Firefox and Safari browsers.
```
lalli-flores, line 21: Thanks for adding the comment! Since it's relevant to the code here, let's keep the comment in the component

src/components/WebexAudioSettings/WebexAudioSettings.jsx
```diff
@@ -18,14 +18,21 @@ import WebexNoMedia from '../WebexNoMedia/WebexNoMedia';
...
+  // https://developer.mozilla.org/en-US/docs/Web/API/HTMLMediaElement/setSinkId
+  const isFirefox = navigator.userAgent.indexOf('Firefox') !== -1; // detect Firefox browser
+  const isSafari = navigator.userAgent.indexOf('Safari') !== -1 && navigator.userAgent.indexOf('Chrome') === -1; // detect Safari browser
+  const checkBrowser = isFirefox || isSafari;
```
lalli-flores, line 25:
```suggestion
  const isSpeakerSupported = !(isFirefox() || isSafari());
```

src/components/WebexAudioSettings/WebexAudioSettings.jsx
```diff
@@ -18,14 +18,21 @@ import WebexNoMedia from '../WebexNoMedia/WebexNoMedia';
...
 
   return (
     <div className={cssClasses} style={style}>
       {display.options?.length !== 0
         ? (
           <>
             <Title>Speaker</Title>
-            <WebexMeetingControl type="switch-speaker" meetingID={meetingID} />
+            {!checkBrowser
```
lalli-flores, line 33:
```suggestion
            {isSpeakerSupported
```

lalli-flores (changes requested):
This doesn't look like UX mockup. In UX mockup, the dropdown is still shown but disabled.
When user hovers over disabled dropdown, tooltip shows with the message

PR #470, opened 2021-09-29
feat(SwitchSpeakerControl): displays a tooltip message for no speaker…

lungancatalin:
…s in Firefox and Safari


src/components/WebexMeetingControl/WebexMeetingControl.jsx
```diff
@@ -70,20 +70,26 @@ function renderButton(action, display, style, showText, asItem) {
  * @param {object} style  Custom style to apply
  * @returns {object} JSX of the component
  */
-function renderDropdown(action, display, style) {
-  const {options, noOptionsMessage, selected} = display;
+const renderDropdown = (action, display, style) => {
```
cipak, line 73: why the change to an arrow function?

src/components/WebexMeetingControl/WebexMeetingControl.jsx
```diff
@@ -70,20 +70,26 @@ function renderButton(action, display, style, showText, asItem) {
...
   return (
-    <Select
-      className="control-select"
-      style={style}
-      value={selected || ''}
-      onChange={(id) => action(id)}
-      options={options?.length === 0 ? [{value: '', label: noOptionsMessage}] : options}
-      disabled={!options?.length}
-    />
+    <>
```
cipak, line 79: is this wrapper needed?

src/components/helpers.js
```diff
@@ -28,3 +28,8 @@ export default function webexComponentClasses(
     ...otherClasses,
   });
 }
+
+const isFirefox = () => navigator.userAgent.indexOf('Firefox') !== -1; // detect Firefox browser
+const isSafari = () => navigator.userAgent.indexOf('Safari') !== -1 && navigator.userAgent.indexOf('Chrome') === -1; // detect Safari browser
+
+export const isSpeakerSupported = !(isFirefox() || isSafari());
```
cipak, line 35: I think a safer (and future-proof, eg when setSinkId will be available in this browsers, probably soon in firefox) would be to check for the existence of setSinkId on a media element.

``js
!!document.createElement('audio').setSinkId;

src/components/generic/Select/Select.jsx
```diff
@@ -84,11 +87,13 @@ Select.propTypes = {
...
   disabled: PropTypes.bool,
+  tooltip: PropTypes.string,
 };
 
 Select.defaultProps = {
   className: '',
   value: '',
   options: [],
   disabled: false,
+  tooltip: '',
```
cipak, line 98: undefined

src/components/WebexMeetingControl/WebexMeetingControl.jsx
```diff
@@ -81,6 +83,8 @@ function renderDropdown(action, display, style) {
       onChange={(id) => action(id)}
       options={options?.length === 0 ? [{value: '', label: noOptionsMessage}] : options}
       disabled={!options?.length}
+      tooltip={tooltip}
+      title={tooltip}
```
cipak, line 87: is `title` needed here?

lungancatalin: You are right. There is no need for it. I took it out.

src/components/helpers.js
```diff
@@ -28,3 +28,6 @@ export default function webexComponentClasses(
     ...otherClasses,
   });
 }
+
+// Checks for the existence of setSinkId on  media element
```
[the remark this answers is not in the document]
cipak: checks for support for setSinkId https://developer.mozilla.org/en-US/docs/Web/API/HTMLMediaElement/setSinkId

PR #471, opened 2021-09-30
feat(WebexSettings): select tabs with arrow keys

bogdanmaier28:
The user is able to go trough the tabs using the arrow keys. This is possible only when the tabs are focused.

Screenshot 2021-09-30 134241


src/components/WebexSettings/Tabs.jsx
```diff
@@ -18,22 +19,52 @@ import webexComponentClasses from '../helpers';
  * @param {string} props.selected  Specifies which tab is active
  * @param {object} props.style  Custom style to apply
  * @param {Tab[]} props.tabs  The tabs to be rendered
+ * @param props.tabIndex
```
cipak, line 22: needs a type

src/components/WebexSettings/Tabs.jsx
```diff
@@ -18,22 +19,52 @@ import webexComponentClasses from '../helpers';
...
   const {content} = tabs.find((tab) => tab.key === selected);
   const cssClasses = webexComponentClasses('tabs', className);
+  const tabListRef = useRef();
+  const leftPress = useKeyPress('ArrowLeft', tabListRef);
+  const rightPress = useKeyPress('ArrowRight', tabListRef);
+
+  useEffect(() => {
+    const selectedTabIndex = tabs.findIndex((tab) => tab.key === selected);
+
+    if (tabs && leftPress) {
```
cipak, line 42: a little late to check if `tabs` is defined, `tabs.findIndex` would already throw an error. You could just use:
```js
(tabs || []).findIndex(...);
```

src/components/WebexSettings/Tabs.jsx
```diff
@@ -43,6 +74,7 @@ export default function Tabs({
               <button
                 onClick={() => onSelect(tab.key)}
                 type="button"
+                tabIndex={-1}
```
cipak, line 77: don't tabs need to also be navigated with the TAB key?

src/components/WebexSettings/Tabs.jsx
```diff
@@ -63,9 +95,11 @@ Tabs.propTypes = {
   selected: PropTypes.string.isRequired,
   style: PropTypes.shape(),
   tabs: PropTypes.arrayOf(PropTypes.object).isRequired,
+  tabIndex: PropTypes.number,
 };
 
 Tabs.defaultProps = {
   className: '',
   style: undefined,
+  tabIndex: '-1',
```
cipak, line 104: `-1` without quotes as it is a number

src/components/WebexSettings/Tabs.jsx
bogdanmaier28: From the documentation we have on Figma it looks like these can be navigated only with arrow keys

src/components/WebexSettings/Tabs.jsx
```diff
@@ -18,22 +19,52 @@ import webexComponentClasses from '../helpers';
  * @param {string} props.selected  Specifies which tab is active
  * @param {object} props.style  Custom style to apply
  * @param {Tab[]} props.tabs  The tabs to be rendered
+ * @param {number} props.tabIndex  Tab index
```
cipak, line 22: a little confusing, this is the html attribute, right? maybe "Tab index html attribute"

src/components/WebexSettings/Tabs.jsx
```diff
@@ -18,22 +19,52 @@ import webexComponentClasses from '../helpers';
...
+  const tabListRef = useRef();
+  const leftPress = useKeyPress('ArrowLeft', tabListRef);
+  const rightPress = useKeyPress('ArrowRight', tabListRef);
+
+  useEffect(() => {
+    const selectedTabIndex = (tabs || []).findIndex((tab) => tab.key === selected);
+
+    if (leftPress) {
+      if (selectedTabIndex > 0) {
+        onSelect(tabs[selectedTabIndex - 1].key);
```
cipak, line 44: the `if` is not needed
```js
const prevTabIndex = (selectedTabIndex + tabs.length - 1) % tabs.length;
onSelect(prevTabIndex);

File: src/components/WebexSettings/Tabs.jsx [hunk 7]
```diff
@@ -18,22 +19,52 @@ import webexComponentClasses from '../helpers';
...
+  tabIndex,
 }) {
   const {content} = tabs.find((tab) => tab.key === selected);
   const cssClasses = webexComponentClasses('tabs', className);
+  const tabListRef = useRef();
+  const leftPress = useKeyPress('ArrowLeft', tabListRef);
+  const rightPress = useKeyPress('ArrowRight', tabListRef);
+
+  useEffect(() => {
+    const selectedTabIndex = (tabs || []).findIndex((tab) => tab.key === selected);
```
cipak, line 40: this can be computed outside the useEffect

src/components/WebexSettings/Tabs.jsx
```diff
@@ -18,22 +19,52 @@ import webexComponentClasses from '../helpers';
...
+        onSelect(tabs[0].key);
+      }
+    }
+    // eslint-disable-next-line
+  }, [rightPress]);
 
   return (
     <div className={cssClasses} style={style}>
-      <ul className="tabs-list">
+      <ul className="tabs-list" tabIndex={tabIndex} ref={tabListRef}>
```
cipak, line 67: I'm not sure why the `ul` needs a tabIndex

src/components/hooks/useKeyPress.js
```diff
@@ -0,0 +1,42 @@
...
+
+    return () => {
+      if (elementRef.current) {
+        elementRef.current.removeEventListener('keydown', downHandler);
+        // eslint-disable-next-line
+        elementRef.current.removeEventListener('keyup', upHandler);
+      }
+    };
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, []);
```
cipak, line 39: this means the effect is only executed once, what if the ref (elementRef.current) is not set yet at that moment?

src/components/hooks/useKeyPress.js
```diff
@@ -0,0 +1,42 @@
+import {useState, useEffect} from 'react';
+
+/**
+ * Custom hook that returns wether a key is pressed on an element.
+ *
+ * @param {string} targetKey  Target key to listen for.
+ * @param {RefObject} elementRef  Element on which to listen for event.
+ * @returns {boolean} Wehter or not a key has been pressed.
```
cipak, line 8: maybe just return the pressed key to simplify the usage, only one listener for all keys.

src/components/WebexSettings/Tabs.jsx
bogdanmaier28: We need the tabIndex for the user to be able to focus the whole list of tabs by pressing tab so he can move trough them with arrow keys. This will also be useful for hints as we will need to let the user know he jumped on the tabs section.

PR #472, opened 2021-09-30
feat(InputField): create and style input field component

alinasuciu:
Create and style input field component.

https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-270413

### DIFF
--- a/src/components/generic/InputField/InputField.jsx
+++ b/src/components/generic/InputField/InputField.jsx
@@ -1,0 +1,65 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import webexComponentClasses from '../../helpers';
+
+/**
+ * Input Field component.
+ *
+ * @param {object} props  Data passed to the component
+ * @param {string} props.className  Custom CSS class to apply
+ * @param {object} props.style  Custom style to apply
+ * @param {string} props.type  Input type
+ * @param {string} props.name  Input name
+ * @param {string} props.value  Input value
+ * @param {string} props.placeholder  Input placeholder
+ * @param {Function} props.onChange  Action to perform on input change
+ * @param {boolean} props.disabled  Flag indicating input disabled
+ * @returns {object} JSX of the component
+ */
+export default function InputField({
+  className,
+  style,
+  type,
+  value,
+  name,
+  placeholder,
+  onChange,
+  disabled,
+}) {
+  const cssClasses = webexComponentClasses('input-field', className);
+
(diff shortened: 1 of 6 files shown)


src/components/generic/InputField/InputField.jsx
```diff
@@ -0,0 +1,70 @@
...
+import PropTypes from 'prop-types';
+import webexComponentClasses from '../../helpers';
+
+/**
+ * Input Field component.
+ *
+ * @param {object} props  Data passed to the component
+ * @param {string} props.className  Custom CSS class to apply
+ * @param {object} props.style  Custom style to apply
+ * @param {string} props.label Input label
```
cipak, line 11: double spaces

src/components/generic/InputField/InputField.jsx
```diff
@@ -0,0 +1,70 @@
...
+  name,
+  placeholder,
+  onChange,
+  isDisabled,
+}) {
+  const cssClasses = webexComponentClasses('input-field', className);
+
+  return (
+    <div className={cssClasses} style={style}>
+      {label && <label htmlFor="input-field">{label}</label>}
```
cipak, line 35: I think we can remove this, `<Select>` does not have an optional label and it needs to be linked to the actual control (htmlfor="input-field" implies that the input has the id `input-field` which it doesn't).

src/components/generic/InputField/InputField.jsx
```diff
@@ -0,0 +1,70 @@
...
+export default function InputField({
+  className,
+  style,
+  label,
+  type,
+  value,
+  name,
+  placeholder,
+  onChange,
+  isDisabled,
```
cipak, line 29: just `disabled`

src/components/generic/InputField/InputField.jsx
```diff
@@ -0,0 +1,70 @@
...
+      />
+    </div>
+  );
+}
+
+InputField.propTypes = {
+  className: PropTypes.string,
+  style: PropTypes.shape(),
+  label: PropTypes.string,
+  type: PropTypes.oneOf(['text', 'number', 'password']),
```
cipak, line 52: there are more types that relate to text (email, tel, search etc.) see here https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input

I would suggest just leaving this as `PropTypes.string` instead of trying to list them all

cipak (approved):

PR #479, opened 2021-10-04
fix(Select): set default value for dropdown

BenjaminGolya:


src/components/WebexMeetingControl/WebexMeetingControl.jsx
```diff
@@ -72,12 +72,13 @@ function renderButton(action, display, style, showText, asItem) {
...
 function renderDropdown(action, display, style) {
   const {options, noOptionsMessage, selected} = display;
+  const selectedValue = !selected ? (options?.length && options[0].value) : selected;
 
   return (
     <Select
       className="control-select"
       style={style}
-      value={selected || ''}
+      value={selectedValue || ''}
```
cipak, line 81: `selected || options?.[0].value || ''`

PR #499, opened 2021-10-11
fix(styles): update remote media styles

jonniespratley:
- Fixes bug with remote share sizing.


jonniespratley:
This should be the right fix. Verified on storybook.

PR #551, opened 2021-11-08
feat(WebexMeetingGuestAuthentication): add custom tab order

patricia0817:
Task link: https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-281286

PR #554, opened 2021-11-09
fix(WebexMeetingAuthentication): regenerate snaphots on the master

karinasigartau0798:


karinasigartau0798:
It is no longer necessary because it was solved in PR #551

PR #556, opened 2021-11-09
fix(WebexMeetingControlBar): set meeting controls tab indexes based on collapseStart

patricia0817:
Task link: https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-282018

### DIFF
--- a/src/components/WebexMeetingControlBar/WebexMeetingControlBar.jsx
+++ b/src/components/WebexMeetingControlBar/WebexMeetingControlBar.jsx
@@ -121,11 +121,13 @@ export default function WebexMeetingControlBar({
   ]);
   /* eslint-enable react-hooks/exhaustive-deps */
 
-  const renderControls = (names) => names.map(
-    (key, i) => (
+  const range = (start, end) => Array(end - start).fill().map((_, i) => start + i);
+
+  const renderControls = (start, end) => range(start, end).map(
+    (i) => (
       <WebexMeetingControl
-        key={key}
-        type={key}
+        key={controlNames[i]}
+        type={controlNames[i]}
         meetingID={meetingID}
         showText={controlTexts[i]}
         autoFocus={controlsTabIndexes[i] === Math.min(...controlsTabIndexes) || false}
@@ -151,9 +153,9 @@ export default function WebexMeetingControlBar({
     );
   };
 
-  const renderCollapseButton = (shown, onClick, ref) => (
+  const renderCollapseButton = (shown, onClick, ref, tabIndex) => (
     <div ref={ref} key="collapse-button" className={sc('collapse-button')}>
-      <Button onClick={onClick} type="toggle" size={48} pressed={shown} tabIndex={controlsTabIndexes[collapseStart]}>
+      <Button onClick={onClick} type="toggle" size={48} pressed={shown} tabIndex={tabIndex}>
         <Icon name={shown ? 'more-adr' : 'more'} />
       </Button>
     </div>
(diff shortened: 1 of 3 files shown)


src/components/WebexMeetingControlBar/WebexMeetingControlBar.jsx
```diff
@@ -193,7 +208,7 @@ export default function WebexMeetingControlBar({
             </div>
           </React.Fragment>
         ))}
-        {renderCollapseButton(collapsedShown, () => {}, collapseButtonRef)}
+        {renderCollapseButton(collapsedShown, () => {}, collapseButtonRef, undefined, false)}
```
cipak, line 211: you have too many params here (5)

src/components/WebexMeetingControlBar/WebexMeetingControlBar.jsx
```diff
@@ -151,9 +158,9 @@ export default function WebexMeetingControlBar({
     );
   };
 
-  const renderCollapseButton = (shown, onClick, ref) => (
+  const renderCollapseButton = (shown, onClick, ref, isVisible) => (
```
cipak, line 161: instead of `isVisible`, you can just send it its tabIndex, which may be undefined

src/components/WebexMeetingControlBar/WebexMeetingControlBar.jsx
```diff
@@ -121,15 +128,15 @@ export default function WebexMeetingControlBar({
   ]);
   /* eslint-enable react-hooks/exhaustive-deps */
 
-  const renderControls = (names) => names.map(
+  const renderControls = (names, controlIndexes) => names.map(
     (key, i) => (
```
cipak, line 132: the `i` here is an index in the sub-array of controls produced by `slice`, however I see that we're using `controlTexts[i]` and `controlsTabIndexes[i]` for computing `showText` and `autoFocus`, which does not seems right as controlTexts and controlsTabIndexes are containing the full list of controls.

My suggestion would be to change this function (renderControls) to receive the start and end indexes of the controls to render (not a sub-array), and use these indexes to extract information from controlNames, controlTexts and controlsTabIndexes.

```js
renderControls(start, end) => range(start, end).map((i) => (
  key = {names[i]}
  showText = {controlTexts[i]}
  tabIndex = {controlsTabIndexes[i]}
)
```
where `range` is a function that produces a list of numbers from `start` to `end-1`, eg:
```js
range = (start, end) => Array(end-start).fill().map((_, i) => start+i)

Comment (cipak) [reply src/components/WebexMeetingControlBar/WebexMeetingControlBar.jsx:132]: And I think it's ok, for simplicity, to have a gap in the tab indexes list when some are collapsed, they don't really need to be consecutive: 1, 2, 5, 6, 7. The collapse button would have the tab index 3 in this case.

File: src/components/WebexMeetingControlBar/WebexMeetingControlBar.jsx [hunk 4]
```diff
@@ -177,9 +179,17 @@ export default function WebexMeetingControlBar({
...
+          {
+            collapseStart < collapseEnd
+            && renderCollapseButton(
+              collapsedShown,
+              toggleCollapsed,
+              undefined,
+              controlsTabIndexes[collapseStart],
+            )
+          }
+          {renderControls(collapseEnd, controlNames.length)}
```
cipak, line 192: use the existing variable controlCount

cipak (approved):

PR #557, opened 2021-11-09
feat(Select): toggle/collapse dropdown when pressing enter or space on the keyboard

karinasigartau0798:
https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-282311

### DIFF
--- a/src/components/generic/Select/Select.jsx
+++ b/src/components/generic/Select/Select.jsx
@@ -46,6 +46,16 @@ export default function Select({
 
   const onOutsideClick = () => setExpanded(false);
 
+  const handleKeyUp = (event) => {
+    if ((event.key === 'Enter' || event.key === ' ') && event.target === event.currentTarget) {
+      toggleExpanded();
+    }
+  };
+
+  const handleBlur = () => {
+    setExpanded(false);
+  };
+
   useEffect(() => {
     let cleanup;
 
@@ -67,6 +77,8 @@ export default function Select({
         tabIndex={tabIndex}
         title={tooltip}
         aria-label={`${label ? `${label}. ` : ''}${ariaLabel}`}
+        onKeyUp={handleKeyUp}
+        onBlur={handleBlur}
       >
         <span className={sc('label')}>{label || value}</span>
         <Icon name={expanded ? 'arrow-up' : 'arrow-down'} size={13} />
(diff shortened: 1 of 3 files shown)


src/components/generic/Select/Select.jsx
```diff
@@ -46,6 +46,18 @@ export default function Select({
...
   const onOutsideClick = () => setExpanded(false);
 
+  const handleKeyDown = (event) => {
+    if (event.key === 'Enter' || event.key === ' ') {
+      setExpanded(!expanded);
+    }
+
+    if (event.key === 'Tab') {
+      if (expanded) {
+        setExpanded(false);
```
cipak, line 56: there's a `toggleExpanded` function you can use

src/components/generic/Select/Select.jsx
```diff
@@ -46,6 +46,18 @@ export default function Select({
 
   const onOutsideClick = () => setExpanded(false);
 
+  const handleKeyDown = (event) => {
+    if (event.key === 'Enter' || event.key === ' ') {
+      setExpanded(!expanded);
```
cipak, line 51: there's a `toggleExpanded` function you can use

src/components/generic/Select/Select.jsx
```diff
@@ -46,6 +46,18 @@ export default function Select({
 
   const onOutsideClick = () => setExpanded(false);
 
+  const handleKeyDown = (event) => {
+    if (event.key === 'Enter' || event.key === ' ') {
```
cipak, line 50: maybe you should also check that the event originated in this element, not a child element (an option), because clicking Enter on an option should select that option, which I guess collapses the dropdown anyway, and an additional toggleExpanded would re-open it.
```js
event.target === event.currentTarget

File: src/components/generic/Select/Select.jsx [hunk 4]
```diff
@@ -67,6 +79,7 @@ export default function Select({
         tabIndex={tabIndex}
         title={tooltip}
         aria-label={`${label ? `${label}. ` : ''}${ariaLabel}`}
+        onKeyDown={handleKeyDown}
```
cipak, line 82: as the change is practically done on the `Select` element, please update the commit message:
> feat(Select): toggle/collapse dropdown when pressing enter or space on the keyboard

cipak (approved):

cipak (approved):

PR #561, opened 2021-11-11
refactor(OptionsList): use state for hovered option instead of :hover pseudoclass

CoroDaniel:
https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-282835

### DIFF
--- a/src/components/generic/OptionsList/OptionsList.jsx
+++ b/src/components/generic/OptionsList/OptionsList.jsx
@@ -1,4 +1,4 @@
-import React from 'react';
+import React, {useState} from 'react';
 import PropTypes from 'prop-types';
 import classNames from 'classnames';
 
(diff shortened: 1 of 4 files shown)


src/components/generic/OptionsList/OptionsList.jsx
```diff
@@ -23,14 +23,17 @@ export default function OptionsList({
 }) {
   const [cssClasses, sc] = webexComponentClasses('options-list', className);
 
+  const [currentOptionIndex, setCurrentOptionIndex] = useState();
```
cipak, line 26: focusedOptionIndex

src/components/generic/OptionsList/OptionsList.jsx
```diff
@@ -23,14 +23,17 @@ export default function OptionsList({
...
       <ul role="menu" className={sc('list')}>
-        {options.map((option) => (
+        {options.map((option, index) => (
           <li
-            className={sc('option')}
+            className={`${sc('option')} ${(currentOptionIndex === index) && sc('option--focused')}`}
             key={option.value}
             onClick={() => onSelect(option)}
+            onMouseEnter={() => setCurrentOptionIndex(index)}
             aria-hidden="true"
```
cipak, line 37: remove this pls

src/components/generic/OptionsList/OptionsList.jsx
```diff
@@ -23,14 +23,17 @@ export default function OptionsList({
...
             role="option"
             aria-selected
```
cipak, line 39: `aria-selected={selected === option.value}`

src/components/generic/OptionsList/OptionsList.jsx
CoroDaniel: I will delete when I add onKeyDown event for enter because it would give error if I only remove that aria-hidden

cipak (approved):

PR #569, opened 2021-11-15
feat(Select): close the option list on blur

patricia0817:
Task link: https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-283354

### DIFF
--- a/src/components/generic/OptionsList/OptionsList.jsx
+++ b/src/components/generic/OptionsList/OptionsList.jsx
@@ -13,6 +13,7 @@ import {useRef, useAutoFocus} from '../../hooks';
  * @param {object[]} props.options  Array of options
  * @param {string} props.className  Custom CSS class to apply
  * @param {string} props.selected  Selected option label
+ * @param {Function} [props.onBlur]  Called when this component loses focus
  * @param {Function} props.onSelect  A function which will be triggerd on option selection
  * @param {number} props.tabIndex  Value of the parent's tabIndex
  * @returns {object}  JSX of the element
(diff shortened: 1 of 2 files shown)


src/components/generic/OptionsList/OptionsList.jsx
```diff
@@ -15,6 +15,7 @@ import {useRef, useAutoFocus} from '../../hooks';
  * @param {string} props.selected  Selected option label
  * @param {Function} props.onSelect  A function which will be triggerd on option selection
  * @param {number} props.tabIndex  Value of the parent's tabIndex
+ * @param {Function} props.onBlur  A function that will close the optionList
```
cipak, line 18: - square brackets as this is optional
- "Called when this component loses focus"

src/components/generic/OptionsList/OptionsList.jsx
```diff
@@ -80,11 +82,13 @@ OptionsList.propTypes = {
   onSelect: PropTypes.func.isRequired,
   selected: PropTypes.string,
   tabIndex: PropTypes.number,
+  onBlur: PropTypes.func,
```
cipak, line 85: alpha order

src/components/generic/Select/Select.jsx
```diff
@@ -89,6 +89,7 @@ export default function Select({
           onSelect={(option) => handleOptionSelect(option)}
           selected={value}
           tabIndex={tabIndex}
+          onBlur={onOutsideClick}
```
cipak, line 92: although the effect is the same, on blur should not call a function dedicated for outside clicks. The reason is that, in the future, someone might  add/change code inside `onOutsideClick` and that should not affect what happens when the option list is blurred.

I would advise renaming the function `onOutsideClick` to `collapse`.

cipak (approved):

PR #574, opened 2021-11-15
feat(WebexMeetingControl): add styled tooltip on collapsed meeting controls

patricia0817:
Task link: https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-283385


src/components/WebexMeetingControl/WebexMeetingControl.jsx
```diff
@@ -44,10 +44,17 @@ function renderButton(sc, action, display, style, showText, asItem, autoFocus, t
 
   if (asItem) {
     output = (
-      <div title={tooltip} style={style}>
+      <Button
+        type="ghost"
+        style={style}
+        size={24}
+        className={sc('collapsed-control-button')}
```
cipak, line 51: use the existing terminology: `as-item`

or rename asItem to collapsed if you prefer, but we need a single term.

src/components/WebexMeetingControl/WebexMeetingControl.jsx
```diff
@@ -44,10 +44,17 @@ function renderButton(sc, action, display, style, showText, asItem, autoFocus, t
...
+        tooltip={tooltip}
+        verticalOffset={15}
```
cipak, line 53: this is the vertical offset of the tooltip, right? as it's a property of the Button, it seems that it specifies the vertical offset of the button, not of the tooltip. Maybe `tooltipVerticalOffset`.

cipak (approved):

PR #580, opened 2021-11-17
feat(WebexComponents): prefix aria label value with button text

karinasigartau0798:
https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-284360

### DIFF
--- a/src/components/WebexMediaAccess/WebexMediaAccess.jsx
+++ b/src/components/WebexMediaAccess/WebexMediaAccess.jsx
@@ -15,7 +15,7 @@ const SCREENS = {
     buttonLabel: 'Proceed without camera',
     buttonAction: 'ignoreVideoAccessPrompt',
     buttonTooltip: 'This setting cannot be changed once the meeting starts.',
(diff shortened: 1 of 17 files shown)


src/components/WebexMediaAccess/WebexMediaAccess.jsx
```diff
@@ -70,7 +70,7 @@ export default function WebexMediaAccess({
         size={40}
         onClick={() => adapter.meetingsAdapter[screen.buttonAction](ID)}
         tooltip={screen.buttonTooltip}
-        ariaLabel={screen.buttonHint}
+        ariaLabel={`${screen.buttonLabel}. ${screen.buttonHint}`}
```
cipak, line 73:
```js
buttonHint: 'Proceed without camera. This setting cannot be changed once the meeting starts.',

cipak: the idea is to put the full text in the buttonHint and not do any concatenation here

cipak (approved):

PR #589, opened 2021-11-23
fix(OptionsList): move the Options component and change the focus logic

CoroDaniel:
https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-285611


src/components/WebexMeetingControlBar/WebexMeetingControlBar.jsx
```diff
@@ -146,7 +146,14 @@ export default function WebexMeetingControlBar({
     ),
   );
 
-  const toggleCollapsed = () => setCollapsedShown(!collapsedShown);
+  const [withKey, setWithKey] = useState(false);
+
+  const toggleCollapsed = (withKeyboard) => {
```
cipak, line 151: let's put everything in a single state:

```js
const toggleCollapsed = (withKey) => {
  setCollapsedShown((shown) => shown ? undefined : {withKey});
};
```

this means `collapsedShown` will be either `undefined` (collapsed controls not shown) or an object with the property `withKey` set to true or false.

cipak (approved):

PR #590, opened 2021-11-23
fix(icons): style audio/video mute icons for disable state of the control to reflect Figma design

karinasigartau0798:


src/components/icons/AudioMicrophoneMutedIcon.scss
```diff
@@ -1,3 +1,7 @@
 .#{$WEBEX_COMPONENTS_CLASS_PREFIX}-audio-microphone-muted-icon {
   path { fill: var(--wxc-icon--muted-color); }
+
+  :disabled & {
```
cipak, line 4: this should be a generic rule for all icons, we could use the common `.wxc-icon` class

cipak (approved):

PR #591, opened 2021-11-23
feat(WebexComponents): don't show the share screen prompt for stories where it is not needed

karinasigartau0798:


src/data/meetings.js
```diff
@@ -313,4 +313,33 @@ export default {
...
+      stream: new MediaStream(),
+    },
+    remoteAudio: null,
+    remoteVideo: null,
+    remoteShare: null,
+    state: MeetingState.JOINED,
+    showRoster: false,
+    settings: {
+      visible: false,
+      preview: {
```
cipak, line 336: when `visible` is false, `preview` does not need to exist

cipak (approved):

PR #608, opened 2021-12-10
fix(Tabs): change focus on tab instead of container

CoroDaniel:
https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-285592

### DIFF
--- a/src/components/WebexSettings/Tabs.jsx
+++ b/src/components/WebexSettings/Tabs.jsx
@@ -1,4 +1,4 @@
-import React from 'react';
+import React, {useState} from 'react';
 import PropTypes from 'prop-types';
 import classNames from 'classnames';
 import {useAutoFocus, useRef} from '../hooks';
@@ -14,6 +14,7 @@ import webexComponentClasses from '../helpers';
  * Tabs component
  *
  * @param {object} props  Data passed to the component
+ * @param {boolean} props.autoFocus   Flag indicating whether to add autofocus on selected tab
  * @param {string} props.className  Custom CSS class to apply
  * @param {Function} props.onSelect  A callback to be executed on selecting active tab
  * @param {string} props.selected  Specifies which tab is active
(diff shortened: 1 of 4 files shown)


src/components/WebexSettings/Tabs.jsx
```diff
@@ -29,49 +30,51 @@ export default function Tabs({
   style,
   tabs,
   tabIndex,
+  autoFocus,
```
cipak, line 33: alpha order

src/components/WebexSettings/Tabs.jsx
```diff
@@ -29,49 +30,51 @@ export default function Tabs({
...
     <div className={cssClasses} style={style}>
-      {/* eslint-disable-next-line */}
       <ul
         className={sc('list')}
-        tabIndex={tabIndex}
-        onKeyUp={handleKeyUp}
-        aria-label="Use arrow keys to navigate between setting options"
         role="tablist"
-        ref={ref}
+        aria-label="Use arrow keys to navigate between setting options."
```
cipak, line 62: is this needed?

cipak: maybe it should read "use left and right arrow keys to change the tab"

src/components/WebexSettings/Tabs.jsx
```diff
@@ -29,49 +30,51 @@ export default function Tabs({
   style,
   tabs,
   tabIndex,
+  autoFocus,
 }) {
   const {content} = tabs.find((tab) => tab.key === selected);
   const [cssClasses, sc] = webexComponentClasses('tabs', className);
   const selectedTabIndex = (tabs || []).findIndex((tab) => tab.key === selected);
   const ref = useRef();
+  const [focus, setFocus] = useState(false);
```
cipak, line 39: this state seems to be a flag indicating that the tab has changed since the initial rendering.

Let's rename it for better semantics:
```js
const [tabChangedWithKey, setTabChangedWithKey] = useState(false);

cipak (approved):

PR #612, opened 2021-12-10
feat(RadioButton): create radio button component

karinasigartau0798:
https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-289976

### DIFF
--- a/src/components/inputs/RadioButton/RadioButton.jsx
+++ b/src/components/inputs/RadioButton/RadioButton.jsx
@@ -1,0 +1,91 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import webexComponentClasses from '../../helpers';
+
+/**
+ * RadioButton component.
+ *
+ * @param {object} props  React props passed to the component
+ * @param {string} [props.ariaLabel]  Radio button aria label
+ * @param {string} [props.className]  Custom CSS class to apply
+ * @param {boolean} [props.selected=false]  Flag indicating whether radio button is selected
+ * @param {boolean} [props.disabled=false]  Flag indicating whether radio button is disabled
+ * @param {boolean} [props.required=false]  Flag indicating whether radio button is required
+ * @param {string} [props.error]  Error text
+ * @param {number} [props.tabIndex]  Value of the tabIndex
+ * @param {string} [props.title]  Radio button title
+ * @param {Function} [props.onChange]  Action to perform on radio button change
+ * @returns {object} JSX of the component
+ */
+export default function RadioButton({
+  ariaLabel,
+  className,
+  selected,
+  disabled,
+  required,
+  error,
+  onChange,
+  tabIndex,
+  title,
+}) {
+  const enabled = !disabled;
+  const SPACE_KEY = ' ';
+  const ENTER_KEY = 'Enter';
+
+  const [cssClasses, sc] = webexComponentClasses('radio-button', className, {
+    enabled,
+    disabled,
+    error,
+  });
+
+  const handleClick = () => enabled && onChange(!selected);
+  const handleKeyDown = (event) => {
+    event.preventDefault();
+    if (enabled && (event.key === SPACE_KEY || event.key === ENTER_KEY)) {
+      onChange(!selected);
+    }
+  };
+
+  return (
+    <div
+      className={cssClasses}
+      role="radio"
+      aria-checked={selected}
+      aria-label={ariaLabel}
+      onClick={handleClick}
+      onKeyDown={handleKeyDown}
+      required={required}
+      aria-disabled={disabled}
+      tabIndex={tabIndex}
+    >
+      <div className={sc('background')}>
+        <div className={sc('foreground')} />
+      </div>
+      {title && <span className={sc('title')}>{title}</span>}
+    </div>
+  );
+}
+
+RadioButton.propTypes = {
+  ariaLabel: PropTypes.string,
+  className: PropTypes.string,
+  selected: PropTypes.bool,
+  disabled: PropTypes.bool,
+  required: PropTypes.bool,
+  error: PropTypes.string,
+  onChange: PropTypes.func,
+  tabIndex: PropTypes.number,
+  title: PropTypes.string,
+};
+
+RadioButton.defaultProps = {
+  ariaLabel: undefined,
+  className: undefined,
+  selected: false,
+  disabled: true,
+  required: false,
+  error: undefined,
+  onChange: undefined,
+  tabIndex: 0,
+  title: undefined,
+};
--- a/src/components/inputs/RadioButton/RadioButton.scss
+++ b/src/components/inputs/RadioButton/RadioButton.scss
@@ -1,0 +1,113 @@
+$C: #{$WEBEX_COMPONENTS_CLASS_PREFIX}-radio-button;
+
+.#{$C} {
+  display: flex;
+  position: relative;
+
+  &__title {
+    color: var(--wxc-input--title--normal);
+    @include body-fonts;
+    margin-left: 0.5rem;
+  }
+
+  &__background {
+    position: relative;
+    display: inline-block;
+    content: '';
+    cursor: pointer;
+    box-sizing: border-box;
+    border-radius: 50%;
+    min-width: 1rem;
+    width: 1rem;
+    height: 1rem;
+  }
+
+  &__foreground {
+    position: absolute;
+    display: inline-block;
+    content: '';
+    border-radius: 50%;
+    width: 0.375rem;
+    height: 0.375rem;
+    top: 0.3rem;
+    left: 0.3rem;
+  }
+
+  &[aria-disabled='false'] {
+    &[aria-checked='false'] {
+      .#{$C}__background {
+        border: var(--wxc-input--off--normal--border);
+        background-color: var(--wxc-input--off--normal--background);
+        .#{$C}__foreground {
+          background-color: var(--wxc-input--off--normal--foreground);
+        }
+        &:hover {
+          border: var(--wxc-input--off--hover--border);
+          background-color: var(--wxc-input--off--hover--background);
+          .#{$C}__foreground {
+            background-color: var(--wxc-input--off--hover--foreground);
+          }
+        }
+        &:active {
+          border: var(--wxc-input--off--active--border);
+          background-color: var(--wxc-input--off--active--background);
+          .#{$C}__foreground {
+            background-color: var(--wxc-input--off--active--foreground);
+          }
+        }
+      }
+    }
+
+    &[aria-checked='true'] {
+      .#{$C}__background {
+        border: var(--wxc-input--on--normal--border);
+        background-color: var(--wxc-input--on--normal--background);
+        .#{$C}__foreground {
+          background-color: var(--wxc-input--on--normal--foreground);
+        }
+        &:hover {
+          border: var(--wxc-input--on--hover--border);
+          background-color: var(--wxc-input--on--hover--background);
+          .#{$C}__foreground {
+            background-color: var(--wxc-input--on--hover--foreground);
+          }
+        }
+        &:active {
+          border: var(--wxc-input--on--active--border);
+          background-color: var(--wxc-input--on--active--background);
+          .#{$C}__foreground {
+            background-color: var(--wxc-input--on--active--foreground);
+          }
+        }
+      }
+    }
+  }
+
+  &[aria-disabled='true'] {
+    cursor: default;
+
+    .#{$C}__title {
+      color: var(--wxc-input--title--disabled);
+    }
+
+    &[aria-checked='false'] {
+      .#{$C}__background {
+        border: var(--wxc-input--off--disabled--border);
+        background-color: var(--wxc-input--off--disabled--background);
+        .#{$C}__foreground {
+          background-color: var(--wxc-input--off--disabled--foreground);
+        }
+      }
+    }
+
+    &[aria-checked='true'] {
+      .#{$C}__background {
+        border: var(--wxc-input--on--disabled--border);
+        background-color: var(--wxc-input--on--disabled--background);
+        .#{$C}__foreground {
+          background-color: var(--wxc-input--on--disabled--foreground);
+        }
+      }
+    }
+  }
+}
--- a/src/themes/dark.scss
+++ b/src/themes/dark.scss
@@ -214,6 +214,9 @@
 
   --wxc-adaptive-cards--highlight-color: #{$wxc-gray-40};
 
+  --wxc-input--title--normal: #{$wxc-white-alpha-95};
+  --wxc-input--title--disabled: #{$wxc-white-alpha-40};
+
   --wxc-input--on--normal--border: none;
   --wxc-input--on--normal--background: #{$wxc-blue-40};
   --wxc-input--on--normal--foreground: #{$wxc-black-alpha-95};
--- a/src/themes/light.scss
+++ b/src/themes/light.scss
@@ -214,6 +214,9 @@
 
   --wxc-adaptive-cards--highlight-color: #{$wxc-gray-70};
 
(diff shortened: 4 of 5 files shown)


src/components/inputs/RadioButton/RadioButton.jsx
```diff
@@ -0,0 +1,85 @@
...
+    disabled,
+    error,
+  });
+
+  return (
+    <div className={cssClasses}>
+      <input
+        role="switch"
+        aria-checked={selected}
+        aria-label={`${name}.${value}`}
```
cipak, line 47: I think the screen readers can properly handle (read) a standard radio button, without us specifying an aria-label. Please check with a screen reader.

src/components/inputs/RadioButton/RadioButton.jsx
```diff
@@ -0,0 +1,85 @@
...
+  selected,
+  disabled,
+  required,
+  error,
+  onChange,
+  tabIndex,
+}) {
+  const enabled = !disabled;
+  const SPACE_KEY = ' ';
+  const ENTER_KEY = 'ENTER';
```
cipak, line 34: I think it's `Enter` not `ENTER`. https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/key/Key_Values#whitespace_keys

But doesn't the browser handle key presses on radio buttons by itself?

src/components/inputs/RadioButton/RadioButton.jsx
```diff
@@ -0,0 +1,85 @@
...
+
+  const [cssClasses] = webexComponentClasses('radio-button', className, {
+    enabled,
+    disabled,
+    error,
+  });
+
+  return (
+    <div className={cssClasses}>
```
cipak, line 43: do we really need this container?

src/components/inputs/RadioButton/RadioButton.scss
```diff
@@ -0,0 +1,81 @@
...
+  input[type=radio] {
+    all:unset;
+  }
+
+  input[type=radio]:before {
+    content: '';
+    box-sizing: border-box;
+    border-radius: 50%;
+    display: inline-block;
+    width: 16px;
```
cipak, line 16: I would prefer relative dimensions (percentages), so one can easily scale the radio button.

src/components/inputs/RadioButton/RadioButton.scss
```diff
@@ -0,0 +1,81 @@
...
+  }
+
+  input[type=radio]:after {
+    content: '';
+    border-radius: 50%;
+    display: inline-block;
+    width: 6px;
+    height: 6px;
+    position: absolute;
+    top:5.5px;
```
cipak, line 27: space after colon

src/components/inputs/RadioButton/RadioButton.scss
```diff
@@ -0,0 +1,81 @@
...
+      }
+      &:active:before {
+        background-color: var(--wxc-input-on--active--background);
+      }
+    }
+
+    input[type=radio]:not(:checked) {
+      &:before {
+        background-color: var(--wxc-input-off--normal--background);
+        border: 0.063rem solid var(--wxc-input-off--normal--border-color);
```
cipak, line 52: please put borders on all states, otherwise the control will have different dimensions when on or off. The border color should be the same as the background for some states.

src/themes/light.scss
```diff
@@ -185,4 +185,22 @@
...
+
+  --wxc-input-off--normal--background: #{$wxc-black-alpha-20};
+  --wxc-input-off--hover--background: #{$wxc-black-alpha-30};
+  --wxc-input-off--active--background: #{$wxc-black-alpha-40};
+  --wxc-input-off--disabled--background: #{$wxc-black-alpha-11};
+
+  --wxc-input-off--normal--foreground: #{$wxc-black-alpha-95};
+  --wxc-input-off--disabled--foreground: #{$wxc-black-alpha-40};
+
+  --wxc-input-off--normal--border-color: #{$wxc-black-alpha-50};
```
cipak, line 205: group by state first (normal) and then property (border):
```css
--wxc-input--on--normal--background
--wxc-input--on--normal--border

--wxc-input--on--hover--background
--wxc-input--on--hover--border

--wxc-input--off--normal--background
etc

Comment (cipak) [reply src/themes/light.scss:205]: use just `--border` instead of `--border-color`, after all, background and foreground are colors too and we did not suffix them

File: src/components/inputs/RadioButton/RadioButton.jsx (hunk 1 above)
Comment (karinasigartau0798) [reply src/components/inputs/RadioButton/RadioButton.jsx:47]: The name and value are not read without adding aria-label.

File: src/components/inputs/RadioButton/RadioButton.jsx (hunk 3 above)
Comment (karinasigartau0798) [reply src/components/inputs/RadioButton/RadioButton.jsx:43]: I think yes, we need, because at this level are set values for the properties such as position, width and height.

Comment (cipak) [reply src/components/inputs/RadioButton/RadioButton.jsx:43]: can't they be set on the input itself?
```jsx
<input className={cssClasses}
  role="switch"
  ...
/>

File: src/components/inputs/RadioButton/RadioButton.jsx [hunk 8]
```diff
@@ -0,0 +1,88 @@
...
+  const ENTER_KEY = 'Enter';
+
+  const [cssClasses] = webexComponentClasses('radio-button', className, {
+    enabled,
+    disabled,
+    error,
+  });
+
+  return (
+    <div
```
cipak, line 43: still wondering if this container is really needed. Can we move all its props on the `<input type="radio">` element? This way, the role/aria attributes won't be needed anymore. I think not even the keyboard handling, as the radio button natively has that.

src/components/inputs/RadioButton/RadioButton.jsx
```diff
@@ -0,0 +1,88 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import webexComponentClasses from '../../helpers';
+
+/**
+ * RadioButton component.
+ *
+ * @param {object} props  React props passed to the component
+ * @param {string} [props.className]  Custom CSS class to apply
+ * @param {string} props.name  Radio button name
```
cipak, line 10: remove name and value

src/components/inputs/RadioButton/RadioButton.jsx
```diff
@@ -0,0 +1,88 @@
...
+    disabled,
+    error,
+  });
+
+  return (
+    <div
+      className={cssClasses}
+      role="radio"
+      aria-checked={selected}
+      aria-label={`${name}.${value}`}
```
cipak, line 47: add a dedicated prop, `ariaLabel`

src/components/inputs/RadioButton/RadioButton.jsx
```diff
@@ -0,0 +1,88 @@
...
+      onClick={enabled ? onChange : undefined}
```
cipak, line 48: create a `handleClick` function that calls `onChange` with the new `selected` value, only if the radio is enabled

```js
handleClick = event => enabled && onChange(!selected)
```

src/components/inputs/RadioButton/RadioButton.jsx
```diff
@@ -0,0 +1,88 @@
...
+};
+
+RadioButton.defaultProps = {
+  className: '',
+  selected: false,
+  disabled: false,
+  required: false,
+  error: undefined,
+  onChange: undefined,
+  tabIndex: undefined,
```
cipak, line 87: the default should be 0, so that the component is focusable

src/components/inputs/RadioButton/RadioButton.jsx
```diff
@@ -0,0 +1,88 @@
...
+      aria-checked={selected}
+      aria-label={`${name}.${value}`}
+      onClick={enabled ? onChange : undefined}
+      onKeyDown={(event) => ((enabled && (event.key === SPACE_KEY || event.key === ENTER_KEY))
+        ? onChange
+        : undefined
+      )}
+      tabIndex={tabIndex}
+    >
+      <input
```
cipak, line 55: tabIndex should be -1 for this element, so that it is not focusable

src/components/inputs/RadioButton/RadioButton.jsx
```diff
@@ -0,0 +1,88 @@
...
+  selected: PropTypes.bool,
+  disabled: PropTypes.bool,
+  required: PropTypes.bool,
+  error: PropTypes.string,
+  onChange: PropTypes.func,
+  tabIndex: PropTypes.number,
+};
+
+RadioButton.defaultProps = {
+  className: '',
```
cipak, line 81: undefined

src/components/inputs/RadioButton/RadioButton.jsx
```diff
@@ -0,0 +1,88 @@
...
+  });
+
+  return (
+    <div
+      className={cssClasses}
+      role="radio"
+      aria-checked={selected}
+      aria-label={`${name}.${value}`}
+      onClick={enabled ? onChange : undefined}
+      onKeyDown={(event) => ((enabled && (event.key === SPACE_KEY || event.key === ENTER_KEY))
```
cipak, line 49: please add a handleKeyDown function to make this code simpler

src/components/inputs/RadioButton/RadioButton.scss
```diff
@@ -0,0 +1,84 @@
+$C: #{$WEBEX_COMPONENTS_CLASS_PREFIX}-radio-button;
+
+.#{$C} {
+  position: relative;
+  width: 1rem;
+  height: 1rem;
+
+  input[type=radio] {
+    all:unset;
```
cipak, line 9: space

src/components/inputs/RadioButton/RadioButton.scss
```diff
@@ -0,0 +1,84 @@
...
+    display: inline-block;
+    width: 37.5%;
+    height: 37.5%;
+    position: absolute;
+    top: 34.4%;
+    left: 34.4%;
+  }
+
+  &--enabled {
+    cursor:pointer;
```
cipak, line 35: space

src/themes/dark.scss
```diff
@@ -185,4 +185,27 @@
   --wxc-tooltip-background: #{$wxc-black-alpha-100};
   --wxc-tooltip-border: 1px solid #{$wxc-white-alpha-20};
   --wxc-tooltip-border-color: #{$wxc-white-alpha-20};
+
+  --wxc-input-on--normal--background: #{$wxc-blue-40};
+  --wxc-input-on--normal--foreground: #{$wxc-black-alpha-95};
+  
+  --wxc-input-on--hover--background: #{$wxc-blue-50};
```
cipak, line 192: please add all 3 variables for each state: border, background, foreground

src/components/inputs/RadioButton/RadioButton.jsx
cipak, line 43: please turn this into a <label> element so that clicking the title selects the radio

src/components/inputs/RadioButton/RadioButton.jsx
```diff
@@ -0,0 +1,88 @@
...
+export default function RadioButton({
+  className,
+  name,
+  value,
+  selected,
+  disabled,
+  required,
+  error,
+  onChange,
+  tabIndex,
```
cipak, line 30: please add a `title` prop and display it to the right of the radio button

src/components/inputs/RadioButton/RadioButton.jsx
```diff
@@ -0,0 +1,93 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import webexComponentClasses from '../../helpers';
+
+/**
+ * RadioButton component.
+ *
+ * @param {object} props  React props passed to the component
+ * @param {string} [props.ariaLabel] Radio button aria label
```
cipak, line 9: double space

src/components/inputs/RadioButton/RadioButton.jsx
```diff
@@ -0,0 +1,93 @@
...
+ * @param {string} [props.className]  Custom CSS class to apply
+ * @param {boolean} [props.selected=false]  Flag indicating whether radio button is selected
+ * @param {boolean} [props.disabled=false]  Flag indicating whether radio button is disabled
+ * @param {boolean} [props.required=false]  Flag indicating whether radio button is required
+ * @param {string} [props.error]  Error text
+ * @param {number} props.tabIndex  Value of the tabIndex
```
cipak, line 15: also optional

src/components/inputs/RadioButton/RadioButton.jsx
```diff
@@ -0,0 +1,95 @@
...
+      onChange(!selected);
+    }
+  };
+
+  return (
+    <div
+      className={cssClasses}
+      role="radio"
+      aria-checked={selected}
+      aria-label={`${ariaLabel}`}
```
cipak, line 54: just `{ariaLabel}`

src/components/inputs/RadioButton/RadioButton.jsx
```diff
@@ -0,0 +1,95 @@
...
+      onClick={handleClick}
+      onKeyDown={handleKeyDown}
+      required={required}
+      aria-disabled={disabled}
+      tabIndex={tabIndex}
+    >
+      <div
+        className={sc('background')}
+        aria-checked={selected}
```
cipak, line 63: remove aria-checked and tabIndex, they are not needed here

src/components/inputs/RadioButton/RadioButton.jsx
```diff
@@ -0,0 +1,93 @@
...
+      role="radio"
+      aria-checked={selected}
+      aria-label={ariaLabel}
+      onClick={handleClick}
+      onKeyDown={handleKeyDown}
+      required={required}
+      aria-disabled={disabled}
+      tabIndex={tabIndex}
+    >
+      <div
```
cipak, line 61: can be made 1 line

src/components/_components.scss
```diff
@@ -2,6 +2,7 @@
 @import 'icons/icons';
 @import 'inputs/inputs';
 @import 'adaptive-cards/adaptive-cards';
+@import 'inputs/inputs';
```
cipak, line 5: duplicate, see above

src/components/inputs/RadioButton/RadioButton.jsx
```diff
@@ -0,0 +1,91 @@
...
+      onClick={handleClick}
+      onKeyDown={handleKeyDown}
+      required={required}
+      aria-disabled={disabled}
+      tabIndex={tabIndex}
+    >
+      <div className={sc('background')}>
+        <div className={sc('foreground')} />
+      </div>
+      {title && <span className={sc('label')}>{title}</span>}
```
cipak, line 64: sc(title)

src/components/inputs/RadioButton/RadioButton.scss
```diff
@@ -0,0 +1,117 @@
...
+  }
+}
+
+.#{$C}[aria-disabled='false'] {
+  .#{$C}__background {
+    border: var(--wxc-input--off--normal--border);
+    background-color: var(--wxc-input--off--normal--background);
+  }
+
+  .#{$C}__foreground {
```
cipak, line 44: for consistency, put foreground inside background like the others. And this seems to belong under aria-checked='false'

src/components/inputs/RadioButton/RadioButton.scss
```diff
@@ -0,0 +1,117 @@
...
+    content: '';
+    border-radius: 50%;
+    width: 0.375rem;
+    height: 0.375rem;
+    top: 0.3rem;
+    left: 0.3rem;
+  }
+}
+
+.#{$C}[aria-disabled='false'] {
```
cipak, line 38: this can go under the previous block (the one that starts with .#{$C} and just add a & at the start of this selector:
```scss
  &[aria-disabled="false"] {

cipak (approved):

PR #629, opened 2021-12-21
feat(AdaptiveCardsComponent): add default props for adaptive cards

CoroDaniel:
https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-294171


src/components/adaptive-cards/Column/Column.jsx
```diff
@@ -43,4 +43,8 @@ Column.acPropTypes = {
   type: acPropTypes.type,
 };
 
+Column.acPropTypes = {
```
cipak, line 46: default

cipak (approved):

PR #631, opened 2021-12-22
feat(AdaptiveCardsInputText): implement Input.Text component

patricia0817:
Task link: https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-292447

### DIFF
--- a/src/components/adaptive-cards/InputText/InputText.jsx
+++ b/src/components/adaptive-cards/InputText/InputText.jsx
@@ -1,6 +1,8 @@
-import React from 'react';
+import React, {useState} from 'react';
 import PropTypes from 'prop-types';
-import {registerComponent} from '../Component/Component';
+import webexComponentClasses from '../../helpers';
+import {acPropTypes, registerComponent} from '../Component/Component';
+import InputField from '../../generic/InputField/InputField';
 
 /**
  * Adaptive Cards Input.Text component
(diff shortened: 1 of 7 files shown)


src/components/adaptive-cards/InputText/InputText.jsx
```diff
@@ -1,23 +1,67 @@
...
+InputText.acPropTypes = {
+  errorMessage: acPropTypes.errorMessage,
+  height: acPropTypes.height,
+  id: acPropTypes.id,
+  isRequired: acPropTypes.isRequired,
+  isVisible: acPropTypes.isVisible,
+  label: acPropTypes.label,
+  maxLength: acPropTypes.maxLength,
+  placeholder: acPropTypes.placeholder,
+  regex: acPropTypes.inputPattern,
```
cipak, line 59: why inputPattern and not regex?

src/components/generic/InputField/InputField.jsx
```diff
@@ -153,6 +161,8 @@ InputField.defaultProps = {
   value: '',
   name: '',
   placeholder: '',
+  maxLength: undefined,
+  pattern: '',
```
cipak, line 165: `undefined` would be slightly better (instead of asking the input to match an empty string regex (which matches anything, actually, so it would work), it's a bit better to don't ask it at all to match anything)

cipak (approved):

cipak (approved):

PR #673, opened 2022-02-01
feat(OptionsList): limit the height of dropdown options

patricia0817:
PR link: https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-302175

### DIFF
--- a/src/components/generic/OptionsList/OptionsList.jsx
+++ b/src/components/generic/OptionsList/OptionsList.jsx
@@ -1,7 +1,8 @@
-import React from 'react';
+import React, {useEffect, useState} from 'react';
 import PropTypes from 'prop-types';
 
 import webexComponentClasses from '../../helpers';
+import {useRef, useElementPosition} from '../../hooks';
 import Option from './Option';
 
 /**
@@ -27,6 +28,9 @@ export default function OptionsList({
   withKey,
 }) {
   const [cssClasses, sc] = webexComponentClasses('options-list', className);
+  const ref = useRef();
+  const position = useElementPosition(ref);
+  const [maxHeight, setMaxHeight] = useState(0);
 
   const onKeyDown = (event) => {
     if (event.key === 'Tab') {
@@ -34,9 +38,15 @@ export default function OptionsList({
     }
   };
 
+  useEffect(() => {
+    if (position) {
+      setMaxHeight(window.innerHeight - position.top - window.scrollY - 50);
+    }
+  }, [position]);
+
   return (
-    <div className={cssClasses}>
-      <ul role="menu" className={sc('list')} tabIndex={tabIndex} onKeyDown={onKeyDown}>
+    <div ref={ref} className={cssClasses}>
+      <ul style={{maxHeight}} role="menu" className={sc('list')} tabIndex={tabIndex} onKeyDown={onKeyDown}>
         {options.map((option, index) => (
           <Option
             key={option.value}
--- a/src/components/hooks/useElementPosition.js
+++ b/src/components/hooks/useElementPosition.js
@@ -1,0 +1,48 @@
+import {useEffect, useState} from 'react';
+
+/**
+ * Custom hook that receives an element reference and returns the position of the element
+ *
+ * @param {object} elementRef  Reference to the element
+ *
+ * @returns {DOMRect} Element position
+ */
+export default function useElementPosition(elementRef) {
+  const [position, setPosition] = useState(null);
+  const target = elementRef.current;
+
+  useEffect(() => {
+    if (target) {
+      const rect = target.getBoundingClientRect();
+
+      if (
+        !position ||
+        rect.top !== position.top ||
+        rect.right !== position.right ||
+        rect.bottom !== position.bottom ||
+        rect.left !== position.left ||
+        rect.x !== position.x ||
+        rect.y !== position.y ||
+        rect.height !== position.height ||
+        rect.width !== position.width
+      ) {
+        setPosition(rect);
+      }
+    }
+  }, [target, position, setPosition]);
+
+  useEffect(() => {
+    let cleanup;
+
+    if (target) {
+      const onWindowResize = () => setPosition(target.getBoundingClientRect());
+
+      window.addEventListener('resize', onWindowResize);
+      cleanup = () => window.removeEventListener('resize', onWindowResize);
+    }
+
+    return cleanup;
+  }, [target]);
+
+  return position;
+}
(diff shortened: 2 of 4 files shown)


src/components/hooks/useElementPosition.js
```diff
@@ -0,0 +1,15 @@
+/**
+ * Custom hook that returns an element reference and the position of the
+ * element it refers to.
+ *
```
cipak, line 4: add a jsdoc description for the parameter

it should be the ref (like useElementDimensions) not only the dom element

src/components/generic/OptionsList/OptionsList.jsx
```diff
@@ -27,16 +28,34 @@ export default function OptionsList({
...
   };
 
+  useEffect(() => {
+    const cleanup = () => window.removeEventListener('resize', handleSetSize);
+
+    handleSetSize();
+    window.addEventListener('resize', handleSetSize);
+
+    return cleanup;
+  });
```
cipak, line 54: if no deps, this effect will execute after each render.

The registering of the resize listener should be done only once, and I think it would be better to let useElementPosition handle it.

src/components/hooks/useElementPosition.js
```diff
@@ -0,0 +1,18 @@
+/**
+ * Custom hook that returns an element reference and the position of the
```
cipak, line 2: that receives an element reference and returns the position of...

src/components/hooks/useElementPosition.js
```diff
@@ -0,0 +1,18 @@
...
+ *
+ * @param {object} elementRef  Reference to the element
+ *
+ * @returns {DOMRect} Element position
+ */
+export default function useElementPosition(elementRef) {
+  const target = elementRef.current;
+  let elementPosition;
+
+  if (target) {
```
cipak, line 13: this should be inside an `useEffect` so that we know that getBoundingClientRect() is called after the dom has been updated https://reactjs.org/docs/hooks-effect.html#effects-without-cleanup

```js
useEffect(() => {
  if (target) {
    setRect(target.getBounding());
  }
});

useEffect(() => {
   window.addEventListener();
   return () => window.removeEventListener();
}, []);
```

the first effect executes after each render, the second one just once when the component mounts.

src/components/hooks/useElementPosition.js
```diff
@@ -0,0 +1,37 @@
...
+  }, [target]);
+
+  useEffect(() => {
+    handleSetSize();
+  }, [handleSetSize]);
+
+  useEffect(() => {
+    let cleanup;
+
+    if (target) {
```
cipak, line 28: no need for this test, handleSetSize knows how to handle a missing target

src/components/generic/OptionsList/OptionsList.jsx
```diff
@@ -27,6 +28,15 @@ export default function OptionsList({
   withKey,
 }) {
   const [cssClasses, sc] = webexComponentClasses('options-list', className);
+  const optionsListRef = useRef();
+  const elementPosition = useElementPosition(optionsListRef);
+  const [maxHeight, setMaxHeight] = useState(0);
+
+  useEffect(() => {
+    if (elementPosition.top) {
```
cipak, line 36: top=0 might be a valid value in some instances, maybe you can just return `null` from the hook if target not yet exists (and as the initial value) instead of an empty object

src/components/generic/OptionsList/OptionsList.jsx
```diff
@@ -27,6 +28,15 @@ export default function OptionsList({
   withKey,
 }) {
   const [cssClasses, sc] = webexComponentClasses('options-list', className);
+  const optionsListRef = useRef();
+  const elementPosition = useElementPosition(optionsListRef);
```
cipak, line 32: just `position` is fine

src/components/generic/OptionsList/OptionsList.jsx
```diff
@@ -27,6 +28,15 @@ export default function OptionsList({
   withKey,
 }) {
   const [cssClasses, sc] = webexComponentClasses('options-list', className);
+  const optionsListRef = useRef();
```
cipak, line 31: just `ref` is fine if it's the topmost element of a component

src/components/hooks/useElementPosition.js
```diff
@@ -0,0 +1,37 @@
...
+ *
+ * @param {object} elementRef  Reference to the element
+ *
+ * @returns {DOMRect} Element position
+ */
+export default function useElementPosition(elementRef) {
+  const [position, setPosition] = useState({});
+  const target = elementRef.current;
+
+  const handleSetSize = useCallback(() => {
```
cipak, line 15: maybe rename this to `updatePosition`

src/components/hooks/useElementPosition.js
```diff
@@ -0,0 +1,37 @@
...
+    if (target) {
+      setPosition(target.getBoundingClientRect());
+    }
+  }, [target]);
+
+  useEffect(() => {
+    handleSetSize();
```
cipak, line 22: because this executes after each render, we need to prevent running setPosition if the position has not changed:
```js
const handleSetSize = useCallback(() => {
  if (!target) {
    setPosition(null);
  } else {
    const rect = target.getBounding();
    if (!position || position.top !== rect.top || position.left !== rect.left || ...) {
      setPosition(rect);
    }
  }
});

File: src/components/hooks/useElementPosition.js [hunk 11]
```diff
@@ -0,0 +1,31 @@
...
+  }
+
+  useEffect(() => {
+    const onWindowResize = () => {
+      if (target) {
+        setPosition(target.getBoundingClientRect());
+      }
+    };
+
+    window.addEventListener('resize', onWindowResize);
```
cipak, line 25: there is no point in adding a resize handler if we have no target, we'd better wait for the target to show up:
```js
let cleanup;

if (target) {
  const onwindowresize = ... (without if);

  window.add;
  cleanup = window.remote;
}

return cleanup;

File: src/components/hooks/useElementPosition.js [hunk 12]
```diff
@@ -0,0 +1,31 @@
...
+ *
+ * @param {object} elementRef  Reference to the element
+ *
+ * @returns {DOMRect} Element position
+ */
+export default function useElementPosition(elementRef) {
+  const [position, setPosition] = useState(null);
+  const target = elementRef.current;
+
+  if (target && !position) {
```
cipak, line 14: we want to run this code after each render, so we need to put it inside a useEffect
we also don't want to call setPosition if the position has not changed from the prev render:
```js
useEffect(() => {
  if (target) {
    const rect = target.getRect();
    if (!position || rect.top !== position.top || rect.left !== position.left || ....) {    
      setPosition(rect);
    }
}, [target]);

cipak: useEffect() guarantees that the code inside is executed after the DOM has been updated, not part of the render function

> Sometimes, we want to run some additional code after React has updated the DOM.

https://reactjs.org/docs/hooks-effect.html#effects-without-cleanup

cipak (approved):

PR #674, opened 2022-02-01
fix(AdaptiveCard): transform getAllValues value to an arrow function

karinasigartau0798:


karinasigartau0798:
closed in favor of #670

PR #676, opened 2022-02-02
fix(InputField): refactor layout structure to match expected behaviour

patricia0817:
Task link: https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-302733

### DIFF
--- a/src/components/generic/InputField/InputField.jsx
+++ b/src/components/generic/InputField/InputField.jsx
@@ -80,26 +80,6 @@ export default function InputField({
       style={style}
     >
       <div className={sc('form-control')}>
-        <input
-          type={isPwdRevealed ? 'text' : type}
-          value={value}
-          name={name}
-          className={sc('input')}
-          placeholder={placeholder}
-          max={max}
-          min={min}
-          maxLength={maxLength}
-          pattern={pattern}
-          onChange={handleChange}
-          disabled={disabled}
-          aria-label={ariaLabel}
-          required={required}
-          // disabling no-autofocus because otherwise this element cannot be autofocused depending on the autoFocus prop
-          // eslint-disable-next-line jsx-a11y/no-autofocus
-          autoFocus={autoFocus}
-          tabIndex={tabIndex}
-          ref={inputRef}
-        />
         {type === 'password' && value && (
           <Button
             type="ghost"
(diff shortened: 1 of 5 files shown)


src/components/generic/InputField/InputField.scss
```diff
@@ -5,6 +5,12 @@ $C: #{$WEBEX_COMPONENTS_CLASS_PREFIX}-input-field;
   flex-direction: column;
   position: relative;
 
+  pointer-events: none;
+
+  & * {
```
cipak, line 10: use the direct child selector please, `& > *`, because the '*' applies to every element in the page, and the browser will only look at the parent to see if it's a .wxc-input-field instead of looking the whole dom up.

cipak: and actually you can use just `> *`, the `& ` prefix is always implied

src/components/generic/InputField/InputField.scss
```diff
@@ -99,8 +105,9 @@ $C: #{$WEBEX_COMPONENTS_CLASS_PREFIX}-input-field;
     right: 0;
     transform: translateY(-50%);
 
-    &:not(:last-child) {
-      right: 2rem;
+    right: 2rem;
```
cipak, line 108: i would prefer to keep the previous syntax, which means "put a margin if it's not the last child", instead of "put a margin, but if it's the last child, remove it"

of course, updated to "second-to-last"

cipak: and you can actually use the simpler `:nth-last-child(n+3)`, which means any child that matches any index between 3 and infinity (n goes from 0 to infinity)

src/components/generic/InputField/InputField.jsx
```diff
@@ -127,6 +107,25 @@ export default function InputField({
...
+          placeholder={placeholder}
+          max={max}
+          min={min}
+          maxLength={maxLength}
+          pattern={pattern}
+          onChange={handleChange}
+          disabled={disabled}
+          aria-label={ariaLabel}
+          required={required}
+          // eslint-disable-next-line jsx-a11y/no-autofocus
```
cipak, line 124: also add the newly added explanation on why we disable this

CoroDaniel:
closed in favor of #738

PR #679, opened 2022-02-03
fix(AdaptiveCard): fix return value of the validate() function

CernaianuMihai:
https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-302809

### DIFF
--- a/src/components/adaptive-cards/AdaptiveCard/AdaptiveCard.jsx
+++ b/src/components/adaptive-cards/AdaptiveCard/AdaptiveCard.jsx
@@ -135,6 +135,11 @@ export default function AdaptiveCard({
 
   const getError = (id) => inputs[id]?.error;
 
+  /**
+   * Validates input values, returns true if all are valid, false if any is invalid
+   *
+   * @returns {boolean} True if all values are valid
+   */
   const validate = () => {
     const newInputs = mapValues(inputs, (input) => {
       let error;
@@ -156,7 +161,7 @@ export default function AdaptiveCard({
 
     setInputs(newInputs);
 
-    return Object.values(newInputs).some((input) => input.error);
+    return Object.values(newInputs).every((input) => !input.error);
   };
 
   const submit = (values) => onSubmit(values);
--- a/src/components/adaptive-cards/ActionSubmit/ActionSubmit.jsx
+++ b/src/components/adaptive-cards/ActionSubmit/ActionSubmit.jsx
@@ -32,6 +32,8 @@ export default function ActionSubmit({className, data, style}) {
           values = {...values, ...data.data};
         }
         submit(values);
+      } else {
+        console.log('Submitted invalid values:', getAllValues());
       }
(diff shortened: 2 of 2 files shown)


src/components/adaptive-cards/AdaptiveCard/AdaptiveCard.jsx
```diff
@@ -155,8 +161,15 @@ export default function AdaptiveCard({
     });
 
     setInputs(newInputs);
+    const result = !Object.values(newInputs).some((input) => input.error);
 
-    return Object.values(newInputs).some((input) => input.error);
+    if (!result) {
```
cipak, line 166: move this in the Action.Submit.jsx file, in the submit logic there is a "if (validate()" check, add an "else" branch

src/components/adaptive-cards/AdaptiveCard/AdaptiveCard.jsx
```diff
@@ -155,8 +161,15 @@ export default function AdaptiveCard({
     });
 
     setInputs(newInputs);
+    const result = !Object.values(newInputs).some((input) => input.error);
```
cipak, line 164: this reads: if not (some inputs have error)
i would prefer: if all inputs don't have errors

hint: use array.every instead of array.some

src/components/adaptive-cards/AdaptiveCard/AdaptiveCard.jsx
```diff
@@ -135,6 +135,12 @@ export default function AdaptiveCard({
 
   const getError = (id) => inputs[id]?.error;
 
+  /**
+   * A function that verifies if the inputs values are valid and returns a boolean value that
```
cipak, line 139: Validates input values, returns true if all are valid, false if any is invalid

src/components/adaptive-cards/AdaptiveCard/AdaptiveCard.jsx
```diff
@@ -135,6 +135,12 @@ export default function AdaptiveCard({
...
+   * specifies whether the input is valid or not
+   *
+   * @returns {boolean} The boolean value
```
cipak, line 142: True if all values are valid

src/components/adaptive-cards/AdaptiveCard/AdaptiveCard.jsx
```diff
@@ -155,8 +161,15 @@ export default function AdaptiveCard({
     });
 
     setInputs(newInputs);
+    const result = !Object.values(newInputs).some((input) => input.error);
 
-    return Object.values(newInputs).some((input) => input.error);
+    if (!result) {
+      console.log('Submitted invalid values:', Object.values(newInputs).reduce(
+        (o, input) => ({...o, [input.id]: input.value}), {},
```
cipak, line 168: you can use getAllValues(), it does the same thing, it collects all values

PR #682, opened 2022-02-04
fix(OptionsList): prevent scroll on page when navigating with arrow keys between dropdown items

CoroDaniel:
https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-303306

### DIFF
--- a/src/components/generic/OptionsList/Option.jsx
+++ b/src/components/generic/OptionsList/Option.jsx
@@ -36,8 +36,10 @@ export default function Option({
     let toFocus;
 
     if (event.key === 'ArrowUp') {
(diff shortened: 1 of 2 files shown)


src/components/generic/OptionsList/OptionsList.jsx
```diff
@@ -29,6 +29,7 @@ export default function OptionsList({
   const [cssClasses, sc] = webexComponentClasses('options-list', className);
 
   const onKeyDown = (event) => {
+    event.preventDefault();
```
cipak, line 32: move this to Option and Select, just for ArrowUp and ArrowDown

cipak (approved):

PR #688, opened 2022-02-08
feat(AdaptiveCardsTextBlock): implement maxLines property

patricia0817:
Task link: https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-303817

### DIFF
--- a/src/components/adaptive-cards/TextBlock/TextBlock.jsx
+++ b/src/components/adaptive-cards/TextBlock/TextBlock.jsx
@@ -11,14 +11,17 @@ import {formatDateTime} from '../util';
  * @param {object} props  React props passed to the component
  * @param {object} props.data  Active cards definition
  * @param {string} [props.className]  Custom CSS class to apply
+ * @param {object} [props.style]  Custom style to apply
  * @returns {object} JSX of the component
  */
-export default function TextBlock({data, className}) {
-  const [cssClasses] = webexComponentClasses('adaptive-cards-text-block', className);
+export default function TextBlock({data, className, style}) {
+  const {maxLines} = data;
+  const [cssClasses] = webexComponentClasses('adaptive-cards-text-block', className, {'max-lines': maxLines});
 
   return (
     <div
       className={cssClasses}
+      style={{...style, WebkitLineClamp: data.maxLines}}
       role={data.style === 'heading' ? 'heading' : undefined}
     >
       {formatDateTime(data.text)}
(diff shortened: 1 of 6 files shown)


src/components/adaptive-cards/TextBlock/TextBlock.jsx
```diff
@@ -14,10 +14,18 @@ import {acPropTypes, registerComponent} from '../Component/Component';
  */
 export default function TextBlock({data, className}) {
   const [cssClasses] = webexComponentClasses('adaptive-cards-text-block', className);
+  const style = {
```
cipak, line 17: this should go in TextBlock.scss like for all other components

src/components/adaptive-cards/TextBlock/TextBlock.jsx
```diff
@@ -14,10 +14,18 @@ import {acPropTypes, registerComponent} from '../Component/Component';
...
+    display: '-webkit-box',
+    flexShrink: 0,
+    WebkitLineClamp: data.maxLines,
```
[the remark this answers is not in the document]
cipak: but I see that this component does not receive a style prop, this is a bug.

it should receive a style prop and it should be amended with this prop:
```jsx
<div
  className=...
  style={{...style, WebkitMaxLines: data.maxLines}}

File: src/components/adaptive-cards/TextBlock/TextBlock.scss [hunk 3]
```diff
@@ -0,0 +1,8 @@
+$C: #{$WEBEX_COMPONENTS_CLASS_PREFIX}-adaptive-cards-text-block;
+
+.#{$C} {
+  display: -webkit-box;
```
cipak, line 4: let's not add these properties unless we really need them (maxLines is defined). Otherwise all text blocks will have overflow:hidden and display:webkit-box, which might not be an issue but who knows.

so please add an optional class name .has-max-lines and put this definition under it.

src/components/adaptive-cards/TextBlock/TextBlock.jsx
```diff
@@ -11,14 +11,17 @@ import {formatDateTime} from '../util';
...
  * @param {object} props.data  Active cards definition
  * @param {string} [props.className]  Custom CSS class to apply
+ * @param {object} [props.style]  Custom style to apply
  * @returns {object} JSX of the component
  */
-export default function TextBlock({data, className}) {
-  const [cssClasses] = webexComponentClasses('adaptive-cards-text-block', className);
+export default function TextBlock({data, className, style}) {
+  const {maxLines} = data;
+  const [cssClasses] = webexComponentClasses('adaptive-cards-text-block', className, {maxLines});
```
cipak, line 19: nice and short, but this will generate a camel-case classname, and we're generally using dash-separated class names, so `{'max-lines': maxLines}`

cipak (approved):

PR #698, opened 2022-02-10
feat(AdaptiveCards): implement selectAction property

CoroDaniel:
https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-304091

### DIFF
--- a/src/components/adaptive-cards/hooks/useAction.js
+++ b/src/components/adaptive-cards/hooks/useAction.js
@@ -1,0 +1,17 @@
+import useActionOpenUrl from './useActionOpenUrl';
+import useActionSubmit from './useActionSubmit';
+import useActionToggleVisibility from './useActionToggleVisibility';
+
+/**
+ * Custom hook that receives an adaptive card action and returns html attributes for the target element
+ *
+ * @param {object} data  Action properties
+ * @returns {object} Action attributes
+ */
+export default function useAction(data) {
+  const handleActionOpenUrl = useActionOpenUrl(data);
+  const handleActionSubmit = useActionSubmit(data);
+  const handleActionToggleVisibility = useActionToggleVisibility(data);
+
+  return handleActionOpenUrl || handleActionSubmit || handleActionToggleVisibility;
+}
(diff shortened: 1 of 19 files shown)


src/components/adaptive-cards/hooks/useAction.js
```diff
@@ -0,0 +1,15 @@
+import useActionOpenUrl from './useActionOpenUrl';
+import useActionSubmit from './useActionSubmit';
+import useActionToggleVisibility from './useActionToggleVisibility';
+
+/**
```
cipak, line 5: Custom hook that receives an adaptive card action and returns html attributes for the target element

src/components/adaptive-cards/hooks/useActionOpenUrl.js
```diff
@@ -0,0 +1,24 @@
+import {isValidUrl} from '../../../util';
+
+/**
```
cipak, line 3: Custom hook that receives an adaptive card OpenURL action and returns html attributes for the target element

src/components/adaptive-cards/hooks/useActionOpenUrl.js
```diff
@@ -0,0 +1,24 @@
+import {isValidUrl} from '../../../util';
+
```
cipak, line 2:
```js
/**
 * HTML attributes for the action target element.
 * @typedef {object} ActionAttributes
 * @property {function} onClick ...
 */

File: src/components/adaptive-cards/hooks/useActionOpenUrl.js (hunk 2 above)
Comment (cipak) [reply src/components/adaptive-cards/hooks/useActionOpenUrl.js:3]: If no action is received, no attributes are returned

File: src/themes/light.scss [hunk 4]
```diff
@@ -308,6 +308,14 @@
 
   --wxc-date-input--calendar-picker-indicator--filter: invert(60%);
 
+  --wxc-adaptive-cards--action--hover--border: none;
```
cipak, line 311: this will force the border of that element to "none", even if that element has a different border.

I would suggest changing this to "inherit". Same for foreground.

But are these 2 variabiles used anywhere?

CoroDaniel: I now put them in Component.scss

cipak (approved):

PR #724, opened 2022-02-16
build(rollup): fix build

karinasigartau0798:
https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-306443


cipak:
We had some problems using the components build under widgets, because of some newly introduced 3rd party libraries, adaptive expressions and markdown-it. Adding "browser: true" to the rollup node resolver solves these problems.

cipak (approved):

PR #726, opened 2022-02-17
fix(AdaptiveCard): validate time around midnight

CernaianuMihai:


cipak (approved):

cipak:
https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/time#making_min_and_max_cross_midnight

PR #734, opened 2022-02-22
feat(Dropdown): improve accessibility specs

karinasigartau0798:


src/components/generic/OptionsList/OptionsList.jsx
```diff
@@ -20,6 +22,8 @@ import Option from './Option';
  */
 export default function OptionsList({
   className,
+  ID,
```
cipak, line 25: id and labelId

src/util.js
```diff
@@ -130,6 +132,17 @@ export function mapValues(object, mapper) {
   )));
 }
 
+/**
+ * Generate an unique id
+ *
+ * @returns {object} Returns an unique id
+ */
+export function lastUniqueID() {
```
cipak, line 140: just `uniqueId`

src/util.js
```diff
@@ -130,6 +132,17 @@ export function mapValues(object, mapper) {
   )));
 }
 
+/**
+ * Generate an unique id
```
cipak, line 136: Generates a unique id to be used in DOM by the components

src/util.js
```diff
@@ -130,6 +132,17 @@ export function mapValues(object, mapper) {
...
+ *
+ * @returns {object} Returns an unique id
```
cipak, line 138: {string}

src/components/inputs/Label/Label.jsx
```diff
@@ -28,7 +30,7 @@ export default function Label({
   return (
     // disabling label-has-associated-control as eslint does not see role attribute as a nested control
     // eslint-disable-next-line jsx-a11y/label-has-associated-control
-    <label className={cssClasses} style={style}>
+    <label className={cssClasses} style={style} id={ID}>
```
cipak, line 33: id={labelId}

the idea is that this id is intented just for the label, not for the whole component. Currently the label is the whole component, but we will change that in the future to:

```jsx
  <label className={sc('label-text'}} id={labelId} />
  <div className={sc('control')} />
  <div className={sc('error')} />
</div>

File: src/util.js [hunk 6]
```diff
@@ -1,5 +1,7 @@
 import {Observable} from 'rxjs';
 
+let lastID = 0;
```
cipak, line 3: can this be moved just above the uniqueId function? and use lastId

src/components/generic/OptionsList/OptionsList.jsx
```diff
@@ -10,6 +10,8 @@ import Option from './Option';
  *
  * @param {object} props  Data passed to the component
  * @param {string} props.className  Custom CSS class to apply
+ * @param {string} [props.id]  Options list id
+ * @param {string} [props.labelId]  Options list id
```
cipak, line 14: Label id

cipak (approved):

PR #737, opened 2022-02-25
refactor(InputField): improve input field UI states

CoroDaniel:
https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-307769

### DIFF
--- a/src/components/generic/InputField/InputField.jsx
+++ b/src/components/generic/InputField/InputField.jsx
@@ -1,5 +1,6 @@
 import React from 'react';
 import PropTypes from 'prop-types';
+import classNames from 'classnames';
 import webexComponentClasses from '../../helpers';
 import {useRef, useAutoFocus} from '../../hooks';
 import Label from '../../inputs/Label/Label';
@@ -77,7 +78,7 @@ export default function InputField({
           // disabling no-autofocus because otherwise this element cannot be autofocused depending on the autoFocus prop
           // eslint-disable-next-line jsx-a11y/no-autofocus
           autoFocus={autoFocus}
-          className={sc('input')}
+          className={classNames(sc('input'), {[sc('error')]: error})}
           disabled={disabled}
           max={max}
           maxLength={maxLength}
--- a/src/components/generic/InputField/InputField.scss
+++ b/src/components/generic/InputField/InputField.scss
@@ -39,55 +39,97 @@ $C: #{$WEBEX_COMPONENTS_CLASS_PREFIX}-input-field;
       }
     }
 
-    &:hover {
+    @mixin disabled {
+      background: var(--wxc-input-field--disabled--background);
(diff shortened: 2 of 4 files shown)


src/components/generic/InputField/InputField.scss
```diff
@@ -1,3 +1,52 @@
+@mixin input-disabled {
```
cipak, line 1: we should have the following states, which should each have their own set of css variables (no mixing of css variables from different states, because this will make it difficult to develop other themes):
- disabled
- static
- hover
- focus
- error-static
- error-hover
- error-focus

cipak: the set of css variables should be (for each state):
--wxc-input-field--<state>--background
- border
- color
- box-shadow
- outline

cipak: please use the same states for the mixin names as listed above (disabled, static, error-static etc.)

also, you can drop the 'input-' prefix if they are declared under `.#{$C} { ... }`, they will be local to this component.

src/components/generic/InputField/InputField.scss
```diff
@@ -39,55 +95,41 @@ $C: #{$WEBEX_COMPONENTS_CLASS_PREFIX}-input-field;
...
-      color: var(--wxc-text-color);
-
-      &:hover {
-        background: var(--wxc-input-field--hover--background);
-        border: var(--wxc-error-input-field--normal--border);
-        box-shadow: none;
-      }
+    &:not(:disabled) {
+      &:focus {
+        &:not(&.#{$C}__input--error) {
```
cipak, line 104: let's split by error first (instead of focus)

src/components/generic/InputField/InputField.jsx
```diff
@@ -77,7 +78,7 @@ export default function InputField({
           // disabling no-autofocus because otherwise this element cannot be autofocused depending on the autoFocus prop
           // eslint-disable-next-line jsx-a11y/no-autofocus
           autoFocus={autoFocus}
-          className={sc('input')}
+          className={classNames(sc('input'), {[sc('input--error')]: error})}
```
cipak, line 81: sc('error') should be enough

src/components/generic/InputField/InputField.scss
```diff
@@ -39,55 +95,41 @@ $C: #{$WEBEX_COMPONENTS_CLASS_PREFIX}-input-field;
...
+          }
+        }
+
+        &:not(:hover) {
+          &:not(&.#{$C}__input--error) {
+            @include input-normal-dormant;
+          }
+
+          &.#{$C}__input--error {
+            @include input-error-dormant
```
cipak, line 130: semicolon

cipak (approved):

PR #738, opened 2022-02-28
fix(components): link label and input control in a more accessible way

CoroDaniel:
https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-307765

### DIFF
--- a/src/components/adaptive-cards/InputChoiceSet/InputChoiceSet.jsx
+++ b/src/components/adaptive-cards/InputChoiceSet/InputChoiceSet.jsx
@@ -8,6 +8,7 @@ import Checkbox from '../../inputs/Checkbox/Checkbox';
 import Label from '../../inputs/Label/Label';
 import RadioButton from '../../inputs/RadioButton/RadioButton';
 import {formatDateTime} from '../util';
+import {uniqueId} from '../../../util';
 
 /**
  * Adaptive Cards InputChoiceSet component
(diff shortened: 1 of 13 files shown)


src/components/adaptive-cards/InputChoiceSet/InputChoiceSet.jsx
```diff
@@ -16,18 +17,20 @@ import {formatDateTime} from '../util';
  * @param {object} props  React props passed to the component
  * @param {object} props.data  Active cards definition
  * @param {string} [props.className]  Custom CSS class to apply
+ * @param {string} [props.id]  InputChoiceSet id
```
cipak, line 20: no need to list the component name here (we are not doing it for className or other props).

just "DOM id" should be fine

src/components/generic/InputField/InputField.jsx
```diff
@@ -70,15 +74,18 @@ export default function InputField({
       label={label}
       required={required}
       style={style}
+      uniqueId={id}
     >
-      <div className={sc('form-control')}>
+      <div className={sc('form-control')} id={id}>
```
cipak, line 79: this id should go to the topmost element (in this case, the Label above)

src/components/inputs/Label/Label.jsx
```diff
@@ -7,67 +7,67 @@ import Icon from '../../generic/Icon/Icon';
...
- * @param {object} [props.style]  Custom style to apply
  * @param {React.ReactNode[]}  props.children  List of children
- * @param {string} [props.label]  Label text
- * @param {string} [props.labelId]  Label id
+ * @param {string} [props.className]  Custom CSS class to apply
  * @param {string} [props.error]  Error text
+ * @param {string} [props.label]  Label text
  * @param {boolean} [props.required=false]  Flag indicating whether the control is required
+ * @param {object} [props.style]  Custom style to apply
+ * @param {string} [props.uniqueId]  Label id
```
cipak, line 16: [props.id] should be fine

cipak: it should be required

cipak (approved):

PR #741, opened 2022-03-01
fix(MeetingsJsonAdapter): prevent camera usage on iOS 15.1 due to iOS bug

patricia0817:
Task link: https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-309635

### DIFF
--- a/src/adapters/MeetingsJSONAdapter.js
+++ b/src/adapters/MeetingsJSONAdapter.js
@@ -61,6 +61,9 @@ const EMPTY_MEETING = {
   speakerID: null,
 };
 
+const ON_IOS_15_1 = typeof navigator !== 'undefined'
+  && navigator.userAgent.includes('iPhone OS 15_1');
+
 // Adapter Events
 const EVENT_MEETING_UPDATE = 'meeting:update';
 
@@ -394,6 +397,13 @@ export default class MeetingsJSONAdapter extends MeetingsAdapter {
     let deviceId;
     let perm;
 
+    if (constraints.video && ON_IOS_15_1) {
+      return {
+        stream: null,
+        perm: 'ERROR',
+      };
+    }
+
     try {
       const mediaStream = await navigator.mediaDevices.getUserMedia(constraints);
       const availableDevices = await this.getAvailableDevices();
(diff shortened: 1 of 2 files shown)


src/adapters/MeetingsJSONAdapter.js
```diff
@@ -403,16 +407,22 @@ export default class MeetingsJSONAdapter extends MeetingsAdapter {
 
       deviceId = availableDevices.find((device) => device.label === deviceLabel)?.deviceId;
 
-      // Filter out either video or audio from a given constraints and return a new media stream
-      if (constraints.video) {
-        stream = new MediaStream([mediaStream.getVideoTracks()[0]]);
-      }
+      if (ON_IOS_15_1) {
```
cipak, line 410: this test should only be made when requesting a video stream, not an audio stream.

src/adapters/MeetingsJSONAdapter.js
```diff
@@ -436,7 +446,12 @@ export default class MeetingsJSONAdapter extends MeetingsAdapter {
       }
     }
 
-    return {stream, perm, deviceId};
+    return {
+      stream,
+      perm,
+      deviceId,
+      err,
```
cipak, line 453: this error should be read by getMeeting() and put in meeting.localVideo.error so that it can be seen by the mute video control.

src/adapters/MeetingsJSONAdapter.js
cipak: and maybe it should be done before asking for permission for video

```js
const mediaStream = await navigator.mediaDevices.getUserMedia(constraints);

File: src/adapters/MeetingsJSONAdapter.js [hunk 3]
```diff
@@ -395,6 +398,7 @@ export default class MeetingsJSONAdapter extends MeetingsAdapter {
     let stream;
     let deviceId;
     let perm;
+    let err;
```
cipak, line 401: let's call this `error` (as it is returned) and rename the error in the catch block to `err`

cipak (approved):

PR #761, opened 2022-03-15
fix(inputs): add aria-label on clear button

CoroDaniel:
https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-313818


src/components/inputs/NumberInput/NumberInput.jsx
```diff
@@ -6,6 +6,10 @@ import {InputField} from '../../generic';
 import Button from '../../generic/Button/Button';
 import Icon from '../../generic/Icon/Icon';
 
+const HINTS = {
+  clearButton: 'Remove text from input',
```
cipak, line 10: being a number input, maybe we can find some other hint.

Eg: Clear input

cipak (approved):

PR #772, opened 2022-03-22
feat(WebexMember): split component into space members and meeting participant

thescripted:
Separate `WebexMember` into`WebexMember` for spaces and `WebexMeetingParticipants` for meetings

[SPARK-308942](https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-308492)


taymoork2 (approved):
LGTM

src/components/WebexMeetingParticipant/WebexMeetingParticipant.jsx
```diff
@@ -0,0 +1,90 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import {DestinationType} from '@webex/component-adapter-interfaces';
+import Spinner from '../generic/Spinner/Spinner';
+
```
lalli-flores, line 5: Is there a reason for this new line?

src/components/WebexMeetingParticipant/WebexMeetingParticipant.jsx
```diff
@@ -0,0 +1,90 @@
...
+ * @param {string} props.memberID  ID of the member for which to display avatar
+ * @param {object} props.style  Custom style to apply
+ * @returns {object} JSX of the component
+ *
+ */
+export default function WebexMeetingParticipant({
+  className,
+  meetingID,
+  displayStatus,
+  memberID,
```
lalli-flores, line 32: Let's call this `personID` because the expectation is that this is the Webex person ID. There is a different member ID, that is per person per meeting that is random and this component doesn't use that

src/components/WebexMember/WebexMember.jsx
```diff
@@ -28,33 +24,18 @@ import webexComponentClasses from '../helpers';
  */
 export default function WebexMember({
   className,
-  destinationID,
-  destinationType,
+  // eslint-disable-next-line no-unused-vars
+  roomID,
```
lalli-flores, line 28: If the room ID is not needed, something is off with this implementation...

lalli-flores: I think what is missing is a check that the given person actually has a membership to the room

src/components/WebexMeetingParticipant/WebexMeetingParticipant.jsx
thescripted: Nope, removing

cipak:
> Separate `WebexMember` into`WebexMember` for spaces and `WebexMeetingParticipant` for meetings Separate `WebexMemberRoster` into`WebexMemberRoster` for spaces and `WebexMeetingRoster` for meetings
> 
> [SPARK-308491](https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-308491) [SPARK-308492](https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-308492)

This seems to introduce some code duplication, can this be avoided?

For example, leaving the generic WebexMember component as-is, and introducing two additional components:
- WebexSpaceMember
- WebexMeetingMember

These components would just use WebexMember with a hardcoded destination type, avoiding of duplication of js code and css:
```jsx
function WebexSpaceMember(...) {
  return <WebexMember destinationType="ROOM" ... />;
}
```

Maybe the same could be done for the roster. I would suggest using the names WebexSpaceRoster and WebexMeetingRoster, so that the difference is clear: `SpaceRoster` vs `MeetingRoster`, instead of `MemberRoster` vs `MeetingRoster` which is a bit confusing.

PR #822, opened 2023-07-26
fix(WebexRemoteMedia): make remote video draggable

rajeshtezu:
Fixes the issue where user is not able to drag the remote video/participants while screen is being shared.

**Issue** - [SPARK-409279](https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-409279)

## Screenshot

https://github.com/webex/components/assets/16598065/2c5338a8-3745-4cf2-b3d5-0a72d4266ab1

### DIFF
--- a/src/components/WebexInMeeting/__snapshots__/WebexInMeeting.stories.storyshot
+++ b/src/components/WebexInMeeting/__snapshots__/WebexInMeeting.stories.storyshot
@@ -22,12 +22,16 @@ Array [
       <div
         className="wxc wxc-remote-media wxc-in-meeting__remote-media-in-meeting wxc-remote-media--remote-video-n-share"
       >
-        <video
-          autoPlay={true}
-          className="wxc-remote-media__video wxc-remote-media__remote-video"
-          muted={true}
-          playsInline={true}
-        />
+        <Draggable
+          bounds="parent"
+        >
+          <video
+            autoPlay={true}
+            className="wxc-remote-media__video wxc-remote-media__remote-video"
+            muted={true}
+            playsInline={true}
+          />
+        </Draggable>
         <video
           autoPlay={true}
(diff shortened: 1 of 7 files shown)


arun3528 (approved):
LGTB, looks really good to see it moving :D

Neeraj-swarnkar (approved):
Looks Good.

src/components/WebexInMeeting/__snapshots__/WebexInMeeting.stories.storyshot
```diff
@@ -22,12 +22,6 @@ Array [
       <div
         className="wxc wxc-remote-media wxc-in-meeting__remote-media-in-meeting wxc-remote-media--remote-video-n-share"
       >
-        <video
```
sreenara, line 25: Why has this been removed from the snapshots? Comment applies to other places where there are deletions as well.

sreenara (commented):
Just one comment.
Have we also tested this on a mobile browser?

rajeshtezu:
Yes, tested on browser. Working as expected

src/components/WebexInMeeting/__snapshots__/WebexInMeeting.stories.storyshot
rajeshtezu: This happened because we wrapped the remote video with `Draggable` component which is mocked in the test file.

sreenara: Can we look for the Draggable component at least in the snapshot? It should be rendered right even if it's a mock object?

rajeshtezu: Updated the mock, now it includes the Draggable and it's children nodes

sreenara (approved):

PR #840, opened 2024-07-19
fix(storybook): migrated storybook from 6.x to 8.x

Kesari3008:
Upgraded storybook to the latest version. Installed test-runner to use instead of storyshots.
https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-540929 

After upgrading to node 20, the storybook started to fail. Upgraded storybook to the latest version using the command 'npx sb upgrade'. After upgrading storybook, auto migration occurred suggesting several changes.

Migration Guide: https://storybook.js.org/docs/migration-guide/from-older-version 
https://github.com/storybookjs/storybook/blob/next/MIGRATION.md 
https://storybook.js.org/docs/writing-tests/snapshot-testing/storyshots-migration-guide#extend-your-test-coverage 
Test-Runner: https://storybook.js.org/docs/writing-tests/test-runner#setup


.storybook/main.js
```diff
@@ -1,12 +1,38 @@
...
   "addons": [
     "@storybook/addon-links",
     "@storybook/addon-essentials",
     "@storybook/preset-scss",
     "storybook-addon-themes",
+    "@storybook/addon-webpack5-compiler-babel",
+    "@chromatic-com/storybook"
   ],
+
+  framework: {
```
sreenara, line 15: Is there any documentation we can point to for adding all of this configuration? Why was it not needed previously?

sreenara (changes requested):
Please add more details to the description regarding how we solved the problem as well as root cause of the failures.
Have a few questions around the changes.

src/storyshots.test.js
```diff
@@ -1,78 +0,0 @@
-import initStoryshots, {Stories2SnapsConverter} from '@storybook/addon-storyshots';
```
sreenara, line 1: If we remove these tests, how are we replacing them with the new packages?

.storybook/preview.js
```diff
@@ -32,7 +32,7 @@ export const decorators = [
 ];
 
 export const parameters = {
-  actions: { argTypesRegex: "^on[A-Z].*" },
+  // actions: { argTypesRegex: "^on[A-Z].*" },
```
sreenara, line 35: Can we have a comment here for why this is removed? Also, please remove commented lines.

.storybook/preview.js
```diff
@@ -73,3 +73,4 @@ export const parameters = {
     ],
   },
 };
+export const tags = ['autodocs'];
```
sreenara, line 76: Can we know what this is supposed to help with?

sreenara (changes requested):
Remaining comments

.storybook/main.js
Kesari3008: Framework change was needed since we upgraded from webpack 4 to webpack 5 so we needed to mention that we are using webpack 5

.storybook/preview.js
Kesari3008: Generates autodocs for components that have been tagged 'autodocs'

.storybook/preview.js
Kesari3008: With Storybook 8.0, argTypesRegex is removed so a different config is recommended to make play functions work. But i cross-checked, we don't need it as there are no play functions being used

src/storyshots.test.js
Kesari3008: We are using test-runner now

Kesari3008:
> Please add more details to the description regarding how we solved the problem as well as root cause of the failures. Have a few questions around the changes.

Added the migration guide and test-runner documentation.

src/storyshots.test.js
rarajes2: There should be some equivalent tests if we are removing this, right?

PR #841, opened 2024-07-25
fix(tooling): upgraded webpack version to be used for storybook

Kesari3008:
https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-540929 

Storybook build failure happened due to incompatibility between Webpack 4 and node 20. Installed the builder-webpack5 version 6.5.16 and used it as a builder for the storybook.

### DIFF
--- a/.storybook/main.js
+++ b/.storybook/main.js
@@ -1,3 +1,5 @@
+const path = require('path');
+
 module.exports = {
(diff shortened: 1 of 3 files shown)


rarajes2 (approved):
LGTM

PR #843, opened 2024-10-05
fix: added WebexMeetingInfo to inmeeting component

rsarika:
Issue: in meeting does not show Meeting name.
Fix: added meeting info component in meeting component


Shreyas281299 (approved):

Kesari3008 (approved):
Test different options : Join meeting using Person ID, email and SIP URI and verify title works in all three

PR #845, opened 2024-11-08
fix(webexMember): updated member name handling

Parimala032:
## This pull request addresses
Issue:- The user is trying to enter the name according to their wish but whatever the name entered by the user unfortunately it not going to display on the sample app when you click to see in the participants list means it taking the system default name so for this issue the following changes are done
## by making following changes
This PR updates the logic for displaying the member's name in the `WebexMember` component. If the `member.name` is unavailable, it will fall back to the `displayName`. This ensures more consistent name handling within the component.

https://github.com/user-attachments/assets/1c4ad4c7-ea95-4dfc-b98d-74b4b9e48356

### **This below recording for the system default `displayName`** 
when you are not providing the name that means whenever the name field is empty and not provides anything then the display name will system default name

https://github.com/user-attachments/assets/0d3f84af-4170-4fbf-ab07-ab35e8ea0eb8

### DIFF
--- a/src/components/WebexMember/WebexMember.jsx
+++ b/src/components/WebexMember/WebexMember.jsx
@@ -39,8 +39,8 @@ export default function WebexMember({
   const members = useMembers(destinationID, destinationType);
   const member = members
     .find((itemMember) => itemMember.ID === personID);
+  const memberName = member?.name || displayName;
   const organization = useOrganization(orgID);
-
   const isMuted = member?.muted;
   const isSpeaking = member?.speaking;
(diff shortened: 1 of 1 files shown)


adhmenon (commented):
Let us get the other PR in first and then merge this.

Parimala032:
> Let us get the other PR in first and then merge this.

Sure @adhmenon first [this sdk-component-adpater PR will go first](https://github.com/webex/sdk-component-adapter/pull/344)

mkesavan13 (approved):

PR #846, opened 2024-11-26
feat(meetings-widgets): fixed-join-meeting-acc

adhmenon:
### Issue

[SPARK-564413](https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-564413)

In case of accessibility, when we were over the `Join Meeting` button, it incorrectly said the hint - 'mute, video on' and not the text.

### Fix

* Fixed code to include the text only and not the hint as per instructions from applause

### GIF
ScreenRecording2024-11-26at4 46 41PM-ezgif com-video-to-gif-converter

### Vidcast of working of buttons
https://app.vidcast.io/share/ae43d3c3-76ab-4c20-8dc9-47f6fb4e9ce3

### Manual Tests

* Tried mute/unmute and video on/video off and ensured it does not break any functionality in the `Join Meeting` button.
* [Working of all other buttons](https://app.vidcast.io/share/ae43d3c3-76ab-4c20-8dc9-47f6fb4e9ce3)
* All vo is exactly the same as before - only now we have fixed Join Meeting issue.

## Summary by CodeRabbit

- **Accessibility Improvements**
	- Updated the accessibility label for the button in the Webex Meeting Control component to enhance clarity for screen readers.

### DIFF
--- a/src/components/WebexMeetingControl/WebexMeetingControl.jsx
+++ b/src/components/WebexMeetingControl/WebexMeetingControl.jsx
@@ -64,7 +64,7 @@ function renderButton(sc, action, display, style, showText, asItem, autoFocus, t
         size={48}
         isDisabled={isDisabled}
         onClick={action}
-        ariaLabel={hint || text}
+        ariaLabel={text}
         pressed={isActive && type === 'TOGGLE'}
         tooltip={tooltip}
         autoFocus={autoFocus}


Shreyas281299 (approved):
Looks good. 
@pagour98 I guess this would cause a conflict with your changes

src/components/WebexMeetingControl/WebexMeetingControl.jsx
```diff
@@ -64,7 +64,7 @@ function renderButton(sc, action, display, style, showText, asItem, autoFocus, t
         size={48}
         isDisabled={isDisabled}
         onClick={action}
-        ariaLabel={hint || text}
+        ariaLabel={text}
```
sreenara, line 67: What were the hints doing earlier. Since we are removing hint, do we know what else gets affected?

sreenara (changes requested):
Please add the list of manual tests done in the description of the PR including existing workflows to ensure nothing else was broken

sreenara (approved):

src/components/WebexMeetingControl/WebexMeetingControl.jsx
sreenara: Let's resolve this after @adhmenon replies.

adhmenon: Thanks for the tip Parv. Yes, it works even with all buttons. I checked the older behaviour and it is the same for all buttons (except ofc the join meeting one). 
I think they were not setting hint for most buttons and for join meeting they were but this causes an issue in the voiceover. I think it has not broken anything else.
Here is a [Vidcast]() showing the proof.

pagour98 (approved):

PR #847, opened 2024-11-28
fix(fcc): Leave meeting button accessibility issue.

pagour98:
**Issue:** 

[SPARK-564414](https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-564414)

In case of accessibility, when we were over the `Leave meeting` button, the screen reader says "undefined, button"

**Fix**

* Fixed code to include the text and as well as tooltip

**Vidcast of testing:**
https://app.vidcast.io/share/5a101376-0ffd-4881-bcd8-2d41c1e746c3

**Manual Tests Performed**
 
* All voice over is exactly the same as before, the changes is only for "Leave meeting button"
* I have tested mute/unmute and video on/video off and ensured it does not break any functionality in the `Join Meeting` button
* After joining the meeting, I have also tested audio, video, share screen, settings & Leave meeting functionality.  

## Summary by CodeRabbit

- **New Features**
	- Improved accessibility of the Webex Meeting Control by updating button labels for better screen reader support. 

- **Bug Fixes**
	- Ensured that the button label falls back to a tooltip if no text is provided, enhancing user experience.

### DIFF
--- a/src/components/WebexMeetingControl/WebexMeetingControl.jsx
+++ b/src/components/WebexMeetingControl/WebexMeetingControl.jsx
@@ -64,7 +64,7 @@ function renderButton(sc, action, display, style, showText, asItem, autoFocus, t
         size={48}
         isDisabled={isDisabled}
         onClick={action}
-        ariaLabel={text}
+        ariaLabel={text || tooltip}
         pressed={isActive && type === 'TOGGLE'}
         tooltip={tooltip}
         autoFocus={autoFocus}


src/components/WebexMeetingControl/WebexMeetingControl.jsx
```diff
@@ -64,7 +64,7 @@ function renderButton(sc, action, display, style, showText, asItem, autoFocus, t
         size={48}
         isDisabled={isDisabled}
         onClick={action}
-        ariaLabel={text}
+        ariaLabel={text || tooltip}
```
adhmenon, line 67: While this does fix the issue, I am wondering - is there any other way we can get `text` to come as a string for the cross button? Only reason is - I am not sure if tooltip is best used here... although in this case it is highly unlikely that tooltip changes.
Just a thought.

pagour98: @adhmenon for Leave meeting button we are getting text as undefined and tooltip value we are getting as "Leave meeting", so when text is undefined it will read the value of tooltip, this will be a short fix.

adhmenon (approved):
Approving as of now, as it _does_ make sense to use tooltip as a workaround, as it won't change.
However, as discussed offline, please do create a ticket in tech debt to investigate why the text & hint are coming as `undefined`. The issue needs to be checked separately.

Shreyas281299 (approved):

Shreyas281299:
https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-587755 for further investigation

PR #848, opened 2024-11-29
fix(fcc): accessibility issue fix for disabled audio and video button.

pagour98:
**Issue:** 

[SPARK-564415](https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-564415)

In case of accessibility, when the audio and video is disabled (permission not given by the user) the screen reader is not announcing correct status of audio and video button, it should announce No Microphone, disabled and No Camera, disabled instead of this it is announcing No microphone, No camera

**Fix**

* Fixed the code to update the ariaLabel with correct status.

**Vidcast of testing:**
https://app.vidcast.io/share/6e99fbf8-b289-4782-b62a-30ed560feea3

**Manual Tests Performed**
 
* Checked meeting behaviour and ensured we can leave the meeting.
* All voice over is exactly the same as before, the changes is only for "audio and video button"
* I have tested mute/unmute and video on/video off and ensured it does not break any functionality on the `Join Meeting` screen
* I have tried to disable the audio and video button by changing the permission from the browser and verified that audio and video button state is called as No Microphone, disabled and No Camera, disabled as per the JIRA ticket requirements.
* After joining the meeting, I have also tested audio, video, share screen, settings & Leave meeting functionality.  

## Summary by CodeRabbit

- **New Features**
	- Enhanced accessibility of buttons in the Webex Meeting Control by providing more descriptive labels based on specific conditions, improving user experience for those using assistive technologies.
	- Added new ARIA label texts for microphone and camera status to better inform users about their current state.


src/components/WebexMeetingControl/WebexMeetingControl.jsx
```diff
@@ -57,14 +58,19 @@ function renderButton(sc, action, display, style, showText, asItem, autoFocus, t
       </Button>
     );
   } else {
+    if (display.ID === 'mute-audio' && display.state === 'disabled' && text === 'No microphone') {
+      ariaLabelText = 'No Microphone, disabled';
```
pagour98, line 62: The text ('No Microphone, disabled', 'No Camera, disabled') is suggested by the Applause team and mentioned inside the JIRA.

src/components/WebexMeetingControl/WebexMeetingControl.jsx
```diff
@@ -39,6 +39,7 @@ function renderButton(sc, action, display, style, showText, asItem, autoFocus, t
   } = display;
   const isDisabled = display.state === MeetingControlState.DISABLED;
   const isActive = display.state === MeetingControlState.ACTIVE;
+  let ariaLabelText = text;
```
pagour98, line 42: Added this variable because we are using variable text to show button-text and we don't want to show (No Microphone, disabled, No Camera, disabled) as button-text when mic and video is disabled, we have to show (No microphone, No camera) as button-text

src/components/WebexMeetingControl/WebexMeetingControl.jsx
Shreyas281299: Can we put the texts in variables and then use them here. The coderabbit suggestion is also similar.
We can have two variables NO_MIC_ARIA_LABEL. NO_CAMERA_ARIA_LABEL.

src/components/WebexMeetingControl/WebexMeetingControl.jsx
```diff
@@ -57,14 +58,19 @@ function renderButton(sc, action, display, style, showText, asItem, autoFocus, t
...
+    }
     output = (
       <Button
         className={sc('control-button')}
         type={controlTypeToButtonType[type] || 'default'}
         size={48}
         isDisabled={isDisabled}
         onClick={action}
-        ariaLabel={text}
+        ariaLabel={ariaLabelText || tooltip}
```
Shreyas281299, line 73: Why do we need the or condition here with tooltip?

src/components/WebexMeetingControl/WebexMeetingControl.jsx
pagour98: @Shreyas281299 , Code Updated!

adhmenon: Would it not automatically announce `disabled` - since the state is disabled right? Is there a need to add it here also?

pagour98: @adhmenon , Yes, we need to add it here because isDisabled property is not working even if we set the value to true

src/components/WebexMeetingControl/WebexMeetingControl.jsx
Shreyas281299: Okay.  From next time please keep PRs separate. I beleive these PRs are independent so we should not mix the PR's codes.

Shreyas281299 (approved):
Looks good.

adhmenon (approved):
Ok. In that case LGTM.

PR #850, opened 2025-03-31
Fix: Removed aria-label to label the select menu properly

khushi19099:
**Issue:** 
[SPARK-564420]: Screen Readers: Label includes incorrect instructions

**Fix**
Removed the aria-label and used aria-labelledby pointing to the id of the visible label so the select menu is properly labeled.

## Summary by CodeRabbit

- **Bug Fixes**
  - Removed hints for navigating microphone and speaker options, which may impact user accessibility and experience.
- **Refactor**
  - Revised the speaker options control to update the interactive guidance for keyboard navigation.


Shreyas281299 (changes requested):
Please add a testing recording with the VoiceOver enabled to show your changes.

PR #853, opened 2025-05-05
Fix/spark 564418 fixing download button for shared files and screenshot

khushi19099:
What's Changed
- Added a visible Download button below shared file thumbnails in the Webex Activity component.
- Updated the WebexActivity component to render a <button> with accessible attributes (aria-label, role, etc.).
- Introduced a downloadFile() utility inside the component to trigger native download behavior.
- Minor CSS tweaks to display the button properly and align within attachment UI.

Issue Addressed
Previously, the shared screenshots or file thumbnails were not accompanied by any visible or accessible download option.


Shreyas281299 (changes requested):
- Please add some testing recording showcasing the changes
- The JIRA doesnt mention adding a new download functionality it is just asking us to add the accessibility on focus.  Why is the method `downloadFile` required?
- Please follow the semantic way for the PR title - 
https://github.com/semantic-release/commit-analyzer?tab=readme-ov-file#rules-matching

src/components/WebexActivity/WebexActivity.jsx
```diff
@@ -17,21 +17,65 @@ import WebexAdaptiveCards from '../WebexAdaptiveCards/WebexAdaptiveCards';
...
-  const [cssClasses, sc] = webexComponentClasses('activity', className);
+  const [cssClasses, sc] = webexComponentClasses("activity", className);
+
+  function downloadFile(url, name) {
+    const link = document.createElement("a");
+    link.href = url;
+    link.download = name;
+    document.body.appendChild(link);
+    link.click();
+    document.body.removeChild(link);
```
Shreyas281299, line 33: We should avoid using document.body, you can directly add this to component being returned

PR #861, opened 2026-08-07
docs(sdd): add JiraToPr gate commands and rollup.config.js module mapping

mkesavan13:
## Summary

- add canonical `compile` and `unit-test` commands for JiraToPr Gate 1 and Gate 2
- force `NODE_ENV=test` for the unit-test command because the JiraToPr pod runs with `NODE_ENV=production`
- map root-level `rollup.config.js` to the existing styles module and canonical styles specification

## Root cause

The SPARK-843492 workflow stopped during Spec Sync because `rollup.config.js` is a JavaScript product source outside every module path declared in `.sdd/manifest.json`. After that mapping is corrected, Gate 1 also requires the canonical `commands.compile` role; `commands.build` is not a Gate 1 alias. The pod environment additionally exports `NODE_ENV=production`, which makes Jest load React's production build and fail Storyshots using `act()`.

## Validation

- JSON parsing: passed
- JiraToPr Spec Sync resolver with the representative SPARK-843492 change set: passed; components, adapters, and styles mapped with no warnings
- Gate 1 resolver: `compile` resolves to `npm run build`
- Gate 2 resolver: `unit-test` resolves to `NODE_ENV=test npm run test`
- `npm run build` on Node 20: passed
- `NODE_ENV=test npm run test` on Node 20: 14 suites passed, 230 tests passed
- `npm run linter`: passed with 0 errors and 20 existing warnings


.sdd/manifest.json
```diff
@@ -118,6 +126,10 @@
     },
     {
       "path": "src/styles/",
+      "source_paths": [
```
Kesari3008, line 129: Is this change also needed for the harness ?

Kesari3008 (approved):

